Issue #39, opened 2015-02-27
Implement operators for error handling (roadmap to 1.0)


alexandru:
This issue is related to issue #10. The following operators are not implemented and need to be in the following 1.0 milestones:

1. `onErrorResumeNext( )` — instructs an Observable to emit a sequence of items if it encounters an error
2. `onErrorReturn( )` — instructs an Observable to emit a particular item when it encounters an error
3. `retry( )` — if a source Observable emits an error, resubscribe to it in the hopes that it will complete without error

The first 2 are the equivalents of `recover` and `recoverWith` from Scala's `Future` API. All 3 of them are badly needed.

UPDATE ...

I don't necessarily agree with the naming, since for example `onErrorResumeNext` is something that comes from Visual Basic and in the original meaning implies resuming even if terminated normally, a behavior that wasn't picked up by RxJava and now it creates confusion. I also don't like parameter overloading too much. So we'll have the following signatures, which are pretty self explanatory and fairly familiar (made a combination with Future's API :)) ...

```scala
def onErrorRecoverWith[U >: T](pf: PartialFunction[Throwable, Observable[U]]): Observable[U]

def onErrorFallbackTo[U >: T](that: => Observable[U]): Observable[U]

def onErrorRetryUnlimited: Observable[T]

def onErrorRetry(maxRetries: Long): Observable[T]

def onErrorRetryIf(p: Throwable => Boolean): Observable[T]
```

PR #40, opened 2015-02-27
Trying to implement onErrorResumeNext

alcarvalho:
Implementing #39 onErrorResumeNext

Creating an initial PR to check what I am doing wrong. In the hope that it is not more work to help me then to actually implement it.

The tests are failing with a difference between expected and received emits. I have tried spliting the count with `divideCount` at the middle, at the end (with `sourceCount - 1`) and as it is now at the start. None of them seem to clarify as the error is by a larger amount then just `1`.

### DIFF
--- a/monifu/shared/src/main/scala/monifu/reactive/Observable.scala
+++ b/monifu/shared/src/main/scala/monifu/reactive/Observable.scala
@@ -531,6 +531,14 @@ trait Observable[+T] { self =>
     operators.scan(self, initial)(op)
 
   /**
+   * When stream is interrupted with an error, resumes it with given Observable.
+   *
+   * @param resumeSequence the Observable that will emit elements if the current has an error
+   */
+  def onErrorResumeNext[U >: T](resumeSequence: Observable[U]): Observable[U] =
+    operators.onError.onErrorResumeNext(self, resumeSequence)
+
+  /**
    * Executes the given callback when the stream has ended,
    * but before the complete event is emitted.
    *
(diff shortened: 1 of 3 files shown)


alexandru:
@alcarvalho that's actually how it's done :+1: 

- I have updated issue #39 with a description of the signatures needed, like for example since we are in Scala, we can use partial functions, which is pretty cool to have - I am considering changing the naming a little, not much for them to remain familiar, but enough to avoid confusion with the original `onErrorResumeNext` and to avoid overloading of parameters
- note that in your test, you're in control of how many events you return from the built observable and the `sourceCount` is just an indication meant for variety (i.e. a random number) ... in other words, if the source triggers an error, you can just resume with a fixed observable of size 10 and have the count be `source.count + 10` and the sum be `source.count + 9 * 5` or something along those lines ... the only rule is that you have to indicate what `sum` and what `count` should the test expect on execution
- the `observableInError` is meant for testing how the observable handles the piping of `onError` from upstream, but in this case there will be no error piped (unless you're using that partial function and thus avoiding to handle certain errors)
- I have weird preferences for style and try to avoid syntactic clutter, unless it is necessary, so drop the override and place those calls on different lines than the `def`s :-)

alexandru:
@alcarvalho I forged ahead and made commit [b30e0e0](https://github.com/monifu/monifu/commit/b30e0e0b5e7b917a3ae87a52119bab2661661c6b). Will close this for now.

PR #59, opened 2015-08-16
introduced the initial implementation for windowTimed

mihaisoloi:
This is an initial implementation for `windowTimed` it doesn't have tests yet. This pull request is just for a quick look-over to see if I'm heading in the right direction. If so, I will implement the other flavors of this operator and add the tests and submit it for review again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ TODO
 .*
 .tags*
 .idea*
+*.iml
 *.log
 .DS_Store
 target/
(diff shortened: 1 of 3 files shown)


alexandru:
Dude, that buffer should be a `ReplaySubject`.

alexandru:
See https://github.com/monifu/monifu/commit/38590ce86696ebeca24c38e7a068000a55a7f351

PR #73, opened 2015-09-03
added initialDelay to the debounce operator

mihaisoloi:


monifu/shared/src/main/scala/monifu/reactive/internals/operators/debounce.scala
```diff
@@ -45,7 +45,7 @@ private[reactive] object debounce {
         private[this] var hasValue = false
 
         locally {
-          scheduleNext(timeoutMillis)
+          scheduleNext(initialDelay.toMillis)
```
alexandru: Oh, this isn't so easy and won't work. If there is an `initialDelay`, it should be relative to the *last event emitted* by the source. Your `initialDelay` is relative to the start of your data source. And that is useless.

alexandru: Also note that my initial implementation of `debouce` was repeating elements. The current `debounce` on the other hand follows the behavior of Rx and will only emit a single element after silence. Therefore an `initialDelay` doesn't make sense on `debounce`. Unless we expose the `debounceRepeated` again.

mihaisoloi: Let me see if I got this. Basically an `initialDelay` doesn't make sense on the debounce operator, but if we don't want to wait the `timeout` to have the Observable emitting its items then we should basically have a value that is relative to the _last event emitted_. What I don't understand is if we don't have

alexandru: You're making a mistake in thinking that this operator after that timespan of silence it emits items (plural). After that delay it emits a single item. There's is no `initialDelay` because there is a single `delay` value we are dealing with.

mihaisoloi: ok, I now understand it, thanks

Issue #89, opened 2015-12-23
Integration with Cats


alexandru:
Cats is a new, modular and down to earth alternative for Scalaz. Monifu should integrate cats-core. This means providing implementation for relevant type-classes and possibly using some of its defined types.

Whether this will happen as a separate sub-project (maybe `monifu-cats`), or as a direct dependency, it remains to be seen. The only problem I worry about right now is to not have conflicts with the cats-core (e.g. Monifu's Task vs a future Cats Task, see #88), but hopefully `cats-core` will remain light.

Planning this for 2.0, because this will likely break the current API.

ngbinh:
:+1:

alexandru:
Integration with Cats is on `master` and fairly good.

PR #126, opened 2016-03-15
Issue #89 - cats integration, stage 1

alexandru:
Started to create required type-classes, based on the ones defined in Cats. 

This is not final, there's still a lot of work to do:

- define instances for standard Scala types
- maybe define Monix's own `Iterable` that can be canceled
- define laws, based on `cats-laws`
- define tests based on those laws


non:
How does `FFoldable` differ from something like `Functor`? It looks to me like I can use a functor to implement that type class:

```scala
class FFoldableFromFunctor[F[_]](functor: Functor[F]) extends FFoldable[F] {
  def foldLeftF[A, S](fa: F[A], seed: S)(f: (S, A) => S): F[S] =
    map(fa)(a => f(seed, a))
}
```

It seems like the types that are `FFoldable` are also `Streamable` (i.e. instances of `Functor`) so I wonder what additional work this type class is doing. Does it have different laws from `Functor`?

alexandru:
@non thanks for reviewing. `FFoldable` is not different from `Functor` only if function `f` is getting called only once. So for a container of 1 element, `FFoldable` would be the same as `Functor`. But for sequences of multiple elements, that wouldn't be true anymore. 

It's `foldLeftF` would behave like the normal `foldLeft` except that it wouldn't extract `A` from the `F[_]` context, which is precisely what makes `Functor` or `Monad` special as well.

```scala
foldLeftF(Seq(1,2,3,4,5), seed=0)(_ + _) == Seq(45)
```

This is important because for asynchronous streams is not implementable. Within the context of an asynchronous stream we cannot return that `S` synchronously. Not if we don't want to block threads. So with this in mind, we could solve it by returning some sort of `Future` or `Task`:

```scala
def foldLeft[A,S](fa: F[A], seed: S)(f: (S,A) => S): Future[S]
```

But then asynchronous execution isn't special. Think of an `Iterable`. That would be a synchronous lazy stream and the normal `foldLeft` would force its evaluation. But what if you want to fold an `Iterable` lazily? Lazy evaluation of an `Iterable` makes sense to me, as there are some `Iterables` that are quite expensive to work with. Think of JDBC result-sets.

It is true that `FFoldable` will only get implemented by streams. I thought it would be a good idea to separate it from `Streamable`. Not sure if that's worth it. Unfortunately I don't have experience designing type-classes.

non:
OK, sure, I see now what you are going for here.

It seems reasonable to me -- `cats.Foldable` itself doesn't really have very good laws which constrain its behavior, but we mostly know what it means anyway.

PR #183, opened 2016-06-24
InterleaveObservable and Pad2Observable

leakyabstraction:
Re: https://github.com/monixio/monix/issues/168

add the abilities that can interleave two streams as well as, zipping two streams with padding

(please DON'T merge by any means in its current shape, tests are failing. We will need to talk about and fix them)

### DIFF
--- a/rootdoc.txt
+++ b/rootdoc.txt
@@ -21,4 +21,10 @@ This is the API documentation for the [[https://monix.io Monix]] library.
      implementations
    - [[monix.reactive.subjects]] exposes [[monix.reactive.subjects.Subject Subject]]
      and [[monix.reactive.subjects.ConcurrentSubject ConcurrentSubject]],
-     which are implementations of hot observables that function both as input and output
+     which are implementations of hot observables that function both as input and output     
+ - [[monix.types]] implements type-class shims, to be translated to
+   type-classes provided by libraries such as
+   [[http://typelevel.org/cats/ Cats]] or [[http://scalaz.org/ Scalaz]]
+ - [[monix.cats]] is the optional integration with the
+   [[http://typelevel.org/cats/ Cats]] library, providing translations
+   for the types described in `monix.types`


leakyabstraction:
@alexandru 

I am really sorry for the delay for this ... been very busy this week. Unfortunately, this PR won't pass the tests. I was trying to fix the tests, but got not luck, so I figured might just create the PR first, and then we can talk about it. Regarding your recent comment, I think if I understand it correctly, that is exactly what I am doing in this PR. If one of the child streams completes early, the other child stream will keep emitting items until it completes as well. The problem is that for the composed observable in my local tests, the `onComplete` from time to time is called before the last `onNext`, and that is exactly the `hould work for synchronous observers` test reports, the sum of the emission of the observable misses the last `onNext`. I cannot figure out why that can happen, can you please provide some insights ? I also create the `Pad2Observable` (not sure this is a decent name, feel free to suggest a better one) which is basically a `zip2` type of operator, except when one of the streams finishes, the other stream will continue emit item which is paired with a default value provided in the constructor of the observable. I think this is a helpful pattern in that the InterleaveObservable can be implemented as follows,

```scala
Observable.pad2With(obA1.map(Some(_)), obA2.map(Some(_)), None, None)((a1, a2) => Observable.fromIterable(List(a1, a2).flatten)).flatten
```

Unfortunately, though this passes the synchronized property test, it from time to time fails the async and back pressure property tests. I am essentially reusing the `Zip2Suite` to implement the tests for both of these 2 new Observable operators, not sure if I am using the right one. Can you suggest ?

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Interleave2Observable.scala
```diff
@@ -0,0 +1,215 @@
...
+
+    def rawOnNext(a: A): Future[Ack] = {
+      try out.onNext(a)
+      finally {
+        if (completedCount == 0) select *= -1
+      }
+    }
+
+    def signalOnNextA1(a: A): Future[Ack] = {
+      lastAck = lastAck match {
```
alexandru: This should be `lastAck.syncTryFlatten` otherwise this `match` optimization isn't effective.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Interleave2Observable.scala
```diff
@@ -0,0 +1,215 @@
...
+    // MUST BE synchronized by `self`
+    var lastAck = Continue : Future[Ack]
+    // MUST BE synchronized by `self`
+    // select indicates which child observable that the `onNext` callback should listen to at the moment
+    // in this case, select alternates between 1 (listen to obsA1) and -1 (listen to obsA2)
+    var select = 0
+    // MUST BE synchronized by `self`
+    var hasElemA1 = false
+    // MUST BE synchronized by `self`
+    var hasElemA2 = false
```
alexandru: You copy-pasted `hasElemA1` and `hasElem2` from the zip operator, but you're not using that logic here.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Interleave2Observable.scala
```diff
@@ -0,0 +1,215 @@
...
+    import out.scheduler
+
+    // MUST BE synchronized by `self`
+    var isDone = false
+    // MUST BE synchronized by `self`
+    var lastAck = Continue : Future[Ack]
+    // MUST BE synchronized by `self`
+    // select indicates which child observable that the `onNext` callback should listen to at the moment
+    // in this case, select alternates between 1 (listen to obsA1) and -1 (listen to obsA2)
+    var select = 0
```
alexandru: In case we are going to also going to do `interleave3`, then alternating between `1` and `-1` no longer works. You could do `select = (select + 1) % 2` to alternative between `0` and `1`, or if you want a more efficient version doing the same thing `(select + 1) & 1`.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Interleave2Observable.scala
```diff
@@ -0,0 +1,215 @@
...
+          hasElemA1 = true
+          if (select == 0) select = 1
+
+          lastAck1 = if (select == 1)
+            signalOnNextA1(elem)
+          else
+          // here essentially means waiting until obsA2 is done with its work
+          // when that happens, before passing the `Continue` signal to the upstream observable,
+          // immediately signal `onNext` of the out Observable to emit the elem since it is already available
+            continueA2.future.flatMap {
```
alexandru: Here we can do `continueA2.future.syncTryFlatten.syncFlatMap` as an optimization. In Monix we basically try to collapse asynchronous pipelines when it makes sense and here it does. This is an optimization of course, not a correctness issue.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Interleave2Observable.scala
```diff
@@ -0,0 +1,215 @@
...
+              case Continue ⇒
+                self.synchronized { signalOnNextA1(elem) }
+              case Stop     ⇒
+                self.synchronized {
+                  if (select >= 0)
```
alexandru: You should be careful here. I don't think there should ever be a case where `select < 0` in this `if` and if it is, then it's a bug, because you just received an `onNext` after (I suppose) an `onComplete`. And if you have a bug in this code, you're not going to see it. IMHO, I would get rid of the `if (select >= 0)` and make sure that the opposite never happens by watching the code really hard :-))

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Interleave2Observable.scala
```diff
@@ -0,0 +1,215 @@
...
+    composite += obsA2.unsafeSubscribeFn(new Subscriber[A] {
+      implicit val scheduler = out.scheduler
+
+      def onNext(elem: A): Future[Ack] = self.synchronized {
+        if (isDone) Stop else {
+          hasElemA2 = true
+          if (select == 0) select = -1
+
+          lastAck2 = if (select == -1)
+            signalOnNextA2(elem)
```
alexandru: OK, I realized I don't like something in your logic :-)

You're using `select` to decide which should emit. This is not right, for one because it is more complicated and because the `continueA1` and `continueA2` promises are then superfluous.

IMHO, this `onNext` method should start like this:

```scala
def onNext(elem: A): Future[Ack] = self.synchronized {
  continueA1.future.flatMap {
    case Continue => ???
    case Stop => ???
  }
}
```

In other words, the permission to signal `A2` is given entirely by `continueA1` and nothing else. And `continueA2` should start as an already completed promise, such that `A1` can be signaled at first.

alexandru: BTW, what I'm proposing here is to simplify the state machine. This won't scale to `interleave3`, but you know what - `interleave3` would be really complicated to understand what it does (by users) as well and so lets not implement it until we've got a solid use-case for it.

alexandru:
@leakyabstraction don't worry - as you can see I've been busy as well. See my comments.

Also - as I mentioned in my last comment - we'll wait on `interleave3`, the priority being just normal `interleave` with 2 sources.

BTW, I'm actually working on a related problem - the ability to do load-balancing between multiple subscribers. Here's the work-in-progress if you care to take a look: https://github.com/monixio/monix/pull/184

leakyabstraction:
@alexandru I think what you said makes sense to me. I have made the changes, and indeed the code looks much cleaner now. However still it fails the tests, and the patterns seems to always be the last emission is missing from the counting. I am not quite certain how the test works, but it seems to be the one in my local tests...

Also could you take a look at the `Pad2Observable` which I think is a more general pattern to guide alternate emission of multiple child observables. I am kind of confident the implementation should work, but it doesn't pass the test either ... Is there any problem for the way I am utilizing the tests, or am I just plainly wrong ?

#184 seems to be interesting, but you know what will be also useful ? I would say something like `runLog` in `scalaz-stream`/`fs2`. Users would be able to get collection of items a particular stream emits along its life span (in blocking fashion if this is what they want). We only have `foreach` as an out of the box operator for any observable, but that is not quite enough.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -682,6 +682,23 @@ object Observable {
   def cons[A](head: A, tail: Observable[A]): Observable[A] =
     new builders.ConsObservable[A](head, tail)
 
+  /** Creates a new observable from this observable and another given
+    * observable by interleaving their items into a strictly alternating sequence.
+    *
+    * So the first item emitted by the new observable will be the item whichever
+    * child stream emits first, and the second item will be emitted by the other
```
alexandru: Wait - I thought the result is *deterministic*. This description says that whomever emits first, that's going to be the first stream. I don't think that's desirable in this case, I think the order should be strict, with the result deterministic (first stream emits first, second emits second).

The rationale is this: think of `zip` versus `combineLatest`. The `combineLatest` operator is like a `zip` that emits items whenever one of the streams emits and it doesn't pause any of them so you get a merge that's going to emit at the same pace as the fastest of the streams or even faster. Whereas with `zip`, in order to keep the result deterministic and get the first item pair joined, the second item pair, etc. we end up pausing the stream that's fastest, we make *a sacrifice*.

OK, so in other words - with non-deterministic operators, you need to gain something, like speed (e.g. `merge` versus `concat` / `flatMap`). But `interleave`  behaves on one hand very much like `zip`. So we are *rushing* to emit the first element, but for what? After the first element we are going to end up blocking and pausing so there's no gain. And that initial race condition buys us nothing IMHO.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -821,6 +838,23 @@ object Observable {
...
     Observable.fromIterable(sources).concatMapDelayError[A](identity)
 
+  /**
+    * like `zipWith2`, but when `oa1` completes before `oa2` does, the new observable will
+    * continue emitting items as `f(paddingA1, oa2.onNext())` until `oa2` completes;
+    * and vice versa, when `oa2` completes before `oa1` does, the new observable will
+    * continue emitting items as `f(oa1.onNext(), paddingA2)` until `oa1` completes.
+    */
+  def pad2With[A1,A2,R](oa1: Observable[A1], oa2: Observable[A2], paddingA1: ⇒ A1, paddingA2: ⇒ A2)(f: (A1,A2) ⇒ R): Observable[R] =
+    new builders.Pad2Observable(oa1, oa2, paddingA1, paddingA2)(f)
```
alexandru: Interesting. Two changes:

1. the number that indicates the arity, in this case "2", needs to be last, because we've used this as a convention. In other words I'd prefer `padWith2`. Because it is `zip2` and `zipWith2` and `combineLatest2` and `combineLatestWith2`. The reason is that, normally such functions are overloads and so other libraries simply make them multi-variadic, but I don't particularly like Java-style overloads and the number comes as a suffix to make it easier to change (e.g. you add a new param and you change 2 to 3), since the number being the suffix, you can move the keyboard cursor fast enough - and I've seen other libraries do this as well - stupid reason, but at least there's one :-)
2. would be better to make those by-name params `Coeval` - it's a superficial change, but by-name params IMHO are error-prone and by using [Coeval](https://monix.io/docs/2x/eval/coeval.html) we can control how evaluation happens

```scala
// Original:
pad2With(oa1, oa2, someA, someB)

// Modified:
padWith2(oa1, oa2, Coeval(someA), Coeval(someB))
```

Everything is subject for discussion btw and this type is new and so its usage is inconsistent for now.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -821,6 +838,23 @@ object Observable {
...
+  /**
+    * like `zip2`, but when `oa1` completes before `oa2` does, the new observable will
```
alexandru: This description needs a title.

Also `@param` descriptions in this case would be useful I think, even if a little redundant with the description.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Interleave2Observable.scala
```diff
@@ -0,0 +1,201 @@
...
+import monix.reactive.Observable
+import monix.reactive.observers.Subscriber
+
+import scala.concurrent.{Future, Promise}
+import scala.util.Success
+
+private[reactive] final class Interleave2Observable[+A]
+  (obsA1: Observable[A], obsA2: Observable[A]) extends Observable[A] { self ⇒
+
+  val idx = new java.util.concurrent.atomic.AtomicInteger(0)
```
alexandru: We are not using anything from `java.util.concurrent`, especially in the `shared` directory, because they won't work on Scala.js.

For atomic reference, we have [better atomic implementations](https://monix.io/docs/2x/execution/atomic.html) anyway:

```scala
import monix.execution.atomic.Atomic

val idx = Atomic(0)
```

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Interleave2Observable.scala
```diff
@@ -0,0 +1,201 @@
...
+
+  def unsafeSubscribeFn(out: Subscriber[A]): Cancelable = {
+    import out.scheduler
+
+    // MUST BE synchronized by `self`
+    var isDone = false
+    // MUST BE synchronized by `self`
+    var lastAck = Continue : Future[Ack]
```
alexandru: We have too many `lastAck` stuff. Rename this to `downstreamAck` (being the acknowledgement that we receive from the subscriber when we actually send something).

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Pad2Observable.scala
```diff
@@ -0,0 +1,191 @@
...
+        }
+      }
+
+    // MUST BE synchronized by `self`
+    def signalOnNext(a1: A1, a2: A2): Future[Ack] = {
+      lastAck = lastAck match {
+        case Continue => rawOnNext(a1,a2)
+        case Stop => Stop
+        case async =>
+          async.flatMap {
```
alexandru: I'm not sure if this back-pressuring here is necessary. We have to think about it, but here's my rationale:

1. on the first item only `A1` is allowed to emit, which means `continueA2` is on `Continue` from the start
2. we emit and then we say `continueA1.tryCompleteWith(lastAck)` 
3. and so `A2` will emit only when `continueA1` is complete and this implies that our `lastAck` here is already completed

Also when you store an `ack` and then do an `ack.flatMap`, many times it is better to do `ack.syncTryFlatten.syncFlatMap`, because otherwise a single asynchronous `ack` can introduce asynchronous boundaries on all the remaining events and that's very inefficient.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Pad2Observable.scala
```diff
@@ -0,0 +1,191 @@
...
+          /*!hasElem &&*/ {
+            completedCount += 1
+            completedCount == 2
+          })
+        if (shouldComplete) {
+          lastAck match {
+            case Continue => rawOnComplete()
+            case Stop => () // do nothing
+            case async =>
+              async.onComplete {
```
alexandru: Here we do need to back-pressure, but instead of doing this `match`, just do `lastAck.syncOnContinue`.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Pad2Observable.scala
alexandru: Oh no, ignore my comment, I thought I was still in the Interweave implementation :-))

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -682,6 +682,23 @@ object Observable {
...
+    * child stream which did not get a chance to emit in the first place, the stream
+    * that did get the chance to emit the first item will wait until the other
+    * stream emits; this will continue until one of the child streams finishes,
+    * after that, the other stream will just keep emitting its items until
+    * `onComplete` is called.
+    *
+    * See [[merge]] for a more relaxed alternative that doesn't
+    * emit items in strict alternating sequence.
+    */
+  def interleave[A](oa1: Observable[A], oa2: Observable[A]): Observable[A] =
```
alexandru: @leakyabstraction dude, I jsut realized, this is not the place for the definitions of these operators.

Move them to `ObservableLike.scala` (and note that the operators are more or less in alphabetical order).
This is because these operators are also needed and specialized for `Subject` and for `Pipe`.

alexandru: Again, ignore my comment - oh boy, today I'm distracted. You did the right thing here. This is the companion object :-))

alexandru:
@leakyabstraction OK, for `interleave` I basically took your code and I hit it with a hammer until it worked. I can't even tell you what was wrong in the first place :-)

But anyway, here's a version that now works: https://gist.github.com/alexandru/55b46a140391c5d40b4a2199f8d0bae2

The culprit was most probably a race condition. We might have set `continueA1` and `continueA2` too soon in `onComplete`. In `onComplete` I now forced back-pressure for that assignment as well.

Some notes:

1. the standard `onError` tests still fail so I disabled them for `InterleaveSuite`, because for this operator the behavior for `onError` is non-deterministic and that's totally fine
2. one of your tests had a small mistake, but otherwise the tests were fine

TODO:

1. Lets split this PR in two - I want us to deal with `Pad2` separately
2. Please add `interleave` in `ObservableLike` as a method in addition to that builder
3. See my comments on ScalaDoc

After this we can merge `interleave` and then deal with `Pad2`. Is that OK?

leakyabstraction:
@alexandru sure I will create a separate PR. I think the lessons I learned for implementing `interleave` is there should be one and only one mechanism to drive the concurrency of implementation of any builder/operator. It is simpler and easier to reason, and more importantly it avoids race conditions that are so hard to track and find out why.

PR #184, opened 2016-06-27
Introduce the Consumer type

alexandru:
There are 2 API-breaking changes:

- `SyncObserver` moves to `Observer.Sync` (trait becomes part of the `Observer` companion)
- `SyncSubscriber` moves to `Subscriber.Sync` (trait becomes part of the `Subscriber` companion)
- types with the deprecation warning will be provided

But the major addition is to introduce the `Consumer` type. The `Consumer` type is a factory of subscribers. The problems we are trying to solve:

1. `Observer` (and `Subscriber`) instances are stateful and error prone - for example, by contract, with an observer instance you can subscribe to a single data-source
2. there's no standard way to describe observers that will produce a final result - sure, with observables we could `foldLeft` and then `runAsyncGetFirst` or something along those lines, but those operators are usually meant for pure functions and there is no standard way to describe consumers that consume some streams and then finally produce a result - like for example a consumer that writes into a NIO async file channel and then when the stream is finished to return some statistics, like the number of bytes written to disk
3. and you can't compose these subscribers very well - for example if you have a subscriber instance you can't feed it into something that can build a load-balancer that processes things in parallel

Meet the `Consumer`. The `Consumer` is a factory of `Subscriber` instances with benefits:

```scala
// For processing sums in parallel, useless of course, but can become 
// really helpful for logic sprinkled with I/O bound stuff
val consumer = Consumer
  .loadBalance(parallelism=10, Consumer.foldLeft(Coeval(0L))(_ + _))
  .map(_.sum)

val observable: Observable[Int] = ???
// Our consumer turns our observable into a Task processing sums, w00t!
val task: Task[Long] = observable.runWith(consumer)

// Consume the whole stream and get the result
task.runAsync.foreach(println)
```


alexandru:
Guys, the star of this release is `Consumer.scala` - the other files changed are either tests or the API cosmetic issues and aren't that important. The review is about `Consumer.scala`.

aoprisan:
There is a failing test, not sure why: monix.tckTests.PublisherTest. See travis-ci/push.

alexandru:
@aoprisan it's probably just a timeout due to the system being overloaded with stuff being executed in parallel by SBT, the `tckTests` are a little fragile and I couldn't for the life of me disable parallel execution in SBT, not for a lack of trying. On my localhost if I want to run all the tests I have to do `sbt -mem 4096` otherwise it's going to crash with an out-of-memory-exception due to the Scala.js SBT runner being a memory hog.

aoprisan:
I had a look on the PR and it looks great. To be honest, I have to revisit it as I am not sure I fully got all that is going on there :).

Issue #168, opened 2016-06-17
implement `InterleaveObservable`


leakyabstraction:
This is essentially for the purpose of having a way to construct an observable by deterministically interleaving items emitted from two streams, with the similar behavior as observed in `scalaz-stream/fs2`. This is my first attempt to implement any `Observable` combinator in `monix`, so could you take a look at the below gist, and comment ? If it looks good, I will open a PR.

https://gist.github.com/leakyabstraction/145b57560f7acffb87ee7664d7713c71

alexandru:
Hey @leakyabstraction, interesting. 

Please do a proper Pull Request. I'll assign this ticket to you if you want it.

On a cursory look, is this equivalent with the following?

```scala
Observable
  .zipWith2(source1, source2) { (a,b) => Observable(a,b) }
  .flatten
```

Do you think there would be performance benefits to implementing it manually instead of expressing it with `zipWith2`? It's basically the same logic, except for the added `flatten`, so I guess so.

TODO:

- Please add comments. Monix is heavily commented for problematic stuff. Like for example that `select` variable needs a serious explanation. 
- Needs some tests. Btw - if you'll look at the current tests, I tried to come up with some basic tests that all operators should pass. It's not perfect, but works fine. You can also inspire yourself from `Zip2Suite` which I think will be similar. And add whatever tests that should go there. We also have ScalaCheck so you can use it if inclined.

A note on efficiency - as you probably saw, we strive to be efficient by collapsing pipelines synchronous when we can. There are also some private macros in [Ack.scala](https://github.com/monixio/monix/blob/v2.0-RC6/monix-execution/shared/src/main/scala/monix/execution/Ack.scala#L79), expressed as macros because of the [JVM inlining problem](http://www.azulsystems.com/blog/cliff/2011-04-04-fixing-the-inlining-problem). I'll expose them on the next release.

leakyabstraction:
@alexandru yes what you propose would work. However there are already some primitive operations in current code base that can also be implemented by others. Take `fromIterator`/`fromIterable` as example, they can be alternatively implemented as,

```scala
def fromIterator[A](it: Iterator[A]): Observable[A] = Observable.repeat(it).takeWhile(_.hasNext).map(_.next)
```

I guess with manual implementations, they can end up more efficient ? 

[...]

PR #186, opened 2016-06-28
Interleave2Observable (thanks to @alexandru :-)

leakyabstraction:
This is related to the issue #168, and a split of the original PR #183. I have changed the comments to the new operator that the order of interleaving two observables is fixed, i.e., the first item of the new observable will always come from the first child observable, and then second item will be coming from the other child observable, and so forth.


alexandru:
Cool. I'm merging it, then we can move on to Pad

Issue #166, opened 2016-06-15
Generalize `Task.sequence` for arbitrary collections


guersam:
`Task.sequence` takes `Seq[Task[A]]` and gives back `Task[List[A]]`, which might not be desirable in all cases. 

Generalizing it as `def sequence[A, M[X] <: TraversableOnce[X]: CanBuildFrom](xs: M[Task[A]]): Task[M[A]]` would be helpful, as well as adding `traverse: M[A] => (A => Task[B]) => Task[M[A]]` in the same way.

alexandru:
I don't dislike the idea. I'm generally fearful of `CanBuildFrom`, but it's indeed useful.

But if we do this, then `gather` and `gatherUnordered` and `zipList` also need it, for consistency.

guersam:
You're right. Then should be keep `zipList`, although it doesn't always produce a list? I'd suggest just dropping it IMHO, because it's just an alias and the word 'zip' is sometimes confusing.

alexandru:
Yes, we can still drop it.

alexandru:
@guersam also [Coeval.sequence](https://monix.io/docs/2x/eval/coeval.html#gather-results-from-a-seq-of-coevals) :-)

alexandru:
Note - I think zipList should not ne removed after all. And signature should stay the same. This is for consistency with Observable (I remembered the reason for why it's there :-))

alexandru:
I mean, on Observablr the zip operation is not like map2 or mapBoth. And zipList makes sense for Observable, whereas sequence and gather do not.

guersam:
1. Ok, then how about making `Task.zipList` receive a vararg as the same as `Observable.zipList`?
2. Rewriting `gatherUnordered` using mutable builder brings a lock which might not be desirable. It seems fine to keep it unchanged by itself, but it causes signature inconsistency between `gather` and `gatherUnordered`. We need to choose between more flexible `gather` and consistency between similar functions. What do you prefer?

alexandru:
1. I agree.
2. I did some benchmarks lately and using `Atomic` is cool, but combining `Atomic` with persistent data-structures (like `Vector`) destroys performance as `Vector` is terrible - it's non-blocking, sure, but the performance loss isn't worth it. Just work with `synchronize` and make the signatures similar.

guersam:
[...]

PR #188, opened 2016-06-29
#166 Generalize `Task.sequence` for arbitrary collections

guersam:
docs and test cases will follow.

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
@@ -338,7 +338,7 @@ object Coeval {
     * It's a simple version of [[traverse]].
     */
   def sequence[A, M[X] <: TraversableOnce[X]](sources: M[Coeval[A]])
-                                             (implicit cbf: CanBuildFrom[M[Coeval[A]], A, M[A]]): Coeval[M[A]] = {
+    (implicit cbf: CanBuildFrom[M[Coeval[A]], A, M[A]]): Coeval[M[A]] = {
     val init = evalAlways(cbf(sources))
     val r = sources.foldLeft(init)((acc,elem) => acc.zipWith(elem)(_ += _))
     r.map(_.result())
@@ -349,9 +349,8 @@ object Coeval {
     *
     * It's a generalized version of [[sequence]].
     */
-  def traverse[A, B, M[X] <: TraversableOnce[X]](sources: M[A])
-                                                (f: A => Coeval[B])
-                                                (implicit cbf: CanBuildFrom[M[A], B, M[B]]): Coeval[M[B]] = {
+  def traverse[A, B, M[X] <: TraversableOnce[X]](sources: M[A])(f: A => Coeval[B])
+    (implicit cbf: CanBuildFrom[M[A], B, M[B]]): Coeval[M[B]] = {
     val init = evalAlways(cbf(sources))
     val r = sources.foldLeft(init)((acc,elem) => acc.zipWith(f(elem))(_ += _))
     r.map(_.result())
--- a/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
@@ -21,6 +21,7 @@ import monix.eval.Coeval._
 import monix.types.{Bimonad, Evaluable}
 import scala.annotation.tailrec
 import scala.collection.mutable
+import scala.collection.generic.CanBuildFrom
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 import scala.language.implicitConversions
@@ -336,8 +337,20 @@ object Coeval {
     *
     * For [[Coeval]] this has the same behavior as [[zipList]].
     */
-  def sequence[A](sources: Seq[Coeval[A]]): Coeval[List[A]] =
-    zipList(sources)
+  def sequence[A, M[X] <: TraversableOnce[X]](sources: M[Coeval[A]])
+                                             (implicit cbf: CanBuildFrom[M[Coeval[A]], A, M[A]]): Coeval[M[A]] = {
+    val init = evalAlways(cbf(sources))
+    val r = sources.foldLeft(init)((acc,elem) => acc.zipWith(elem)(_ += _))
+    r.map(_.result())
+  }
+
+  def traverse[A, B, M[X] <: TraversableOnce[X]](sources: M[A])
+                                                (f: A => Coeval[B])
+                                                (implicit cbf: CanBuildFrom[M[A], B, M[B]]): Coeval[M[B]] = {
+    val init = evalAlways(cbf(sources))
+    val r = sources.foldLeft(init)((acc,elem) => acc.zipWith(f(elem))(_ += _))
+    r.map(_.result())
+  }
 
   /** Zips together multiple [[Coeval]] instances. */
   def zipList[A](sources: Seq[Coeval[A]]): Coeval[List[A]] = {
--- a/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
@@ -353,7 +353,7 @@ object Coeval {
   }
 
   /** Zips together multiple [[Coeval]] instances. */
-  def zipList[A](sources: Seq[Coeval[A]]): Coeval[List[A]] = {
+  def zipList[A](sources: Coeval[A]*): Coeval[List[A]] = {
     val init = evalAlways(mutable.ListBuffer.empty[A])
     val r = sources.foldLeft(init)((acc, elem) => acc.zipWith(elem)(_ += _))
     r.map(_.toList)
--- a/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
@@ -332,10 +332,10 @@ object Coeval {
   /** A `Coeval[Unit]` provided for convenience. */
   val unit: Coeval[Unit] = Now(())
 
-  /** Transforms a sequence of coevals into a coeval producing
-    * a list of gathered results.
+  /** Transforms a `TraversableOnce` of coevals into a coeval producing
+    * the same collection of gathered results.
     *
-    * For [[Coeval]] this has the same behavior as [[zipList]].
+    * It's a simple version of [[traverse]].
     */
   def sequence[A, M[X] <: TraversableOnce[X]](sources: M[Coeval[A]])
                                              (implicit cbf: CanBuildFrom[M[Coeval[A]], A, M[A]]): Coeval[M[A]] = {
@@ -344,6 +344,11 @@ object Coeval {
     r.map(_.result())
   }
 
+  /** Transforms a `TraversableOnce[A]` into a coeval of the same collection
+    * using the provided function `A => Coeval[B]`.
+    *
+    * It's a generalized version of [[sequence]].
+    */
   def traverse[A, B, M[X] <: TraversableOnce[X]](sources: M[A])
                                                 (f: A => Coeval[B])
                                                 (implicit cbf: CanBuildFrom[M[A], B, M[B]]): Coeval[M[B]] = {
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -969,7 +969,7 @@ object Task extends TaskInstances {
   /** Gathers the results from a sequence of tasks into a single list.
     * The effects are not ordered, but the results are.
     */
-  def zipList[A](sources: Seq[Task[A]]): Task[List[A]] = {
+  def zipList[A](sources: Task[A]*): Task[List[A]] = {
     val init = evalAlways(mutable.ListBuffer.empty[A])
     val r = sources.foldLeft(init)((acc,elem) => Task.mapBoth(acc,elem)(_ += _))
     r.map(_.toList)
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -821,7 +821,7 @@ object Task extends TaskInstances {
     Async { (scheduler, conn, finalCallback) =>
       var remaining = in.size
       val builder = cbf(in)
-      val lock = new Object()
+      val lock = new AnyRef
       val isActive = Atomic(true)
 
       val composite = CompositeCancelable()
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -28,6 +28,7 @@ import org.reactivestreams.Subscriber
 import monix.execution.atomic.{Atomic, AtomicAny}
 import scala.annotation.tailrec
 import scala.collection.mutable
+import scala.collection.generic.CanBuildFrom
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{Future, Promise, TimeoutException}
 import scala.util.control.NonFatal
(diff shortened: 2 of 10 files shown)


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -793,11 +803,13 @@ object Task extends TaskInstances {
     *
     * Although the effects are unordered, we ensure the order of results
     * matches the order of the input sequence. Also see [[gatherUnordered]].
-    *
-    * Alias for [[zipList]].
     */
-  def gather[A](in: Seq[Task[A]]): Task[List[A]] =
-    zipList(in)
+  def gather[A, M[X] <: TraversableOnce[X]](in: M[Task[A]])
+                                           (implicit cbf: CanBuildFrom[M[Task[A]], A, M[A]]): Task[M[A]] = {
```
guersam: Currently I've chosen diverging signatures between `gather` and `gatherUnordered` because forcing the result to be a list might be undesirable in more cases than having a vector. WDYT?

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -824,7 +837,7 @@ object Task extends TaskInstances {
               atom.get match {
                 case null => ()
                 case current =>
-                  val update = current :+ value
+                  val update = value :: current
                   if (!atom.compareAndSet(current, update))
                     onSuccess(value)
                   else if (update.length == expected) {
```
guersam: I want to avoid O(n) `List#length`. Currently `CountDownLatch` seems to be the simplest solution to me. Although we can't get rid of `val expected = in.length`, the input might not be a list so that the overall performance would not be that bad.

alexandru: We cannot use CountDownLatch because it's not implementable for Scala.js - we could have a State(list, length) in that atomic, though at this point we need to measure the performance compared to a synchronized approach.

guersam:
Sorry for such a long delay, just wrote a benchmark for both approaches. Will post the result of the currently running one tomorrow.

benchmarks/src/main/scala/monix/TaskGatherUnordered.scala
```diff
@@ -0,0 +1,186 @@
...
+        composite += stacked
+
+        Task.unsafeStartNow(task, scheduler, stacked,
+          new Callback[A] {
+            def onSuccess(value: A): Unit = {
+              lock.synchronized {
+                builder += value
+                remaining -= 1
+              }
+              if (remaining == 0) {
```
guersam: Maybe this statement should go into the `synchronized` block as well. Is it safe to fire the final callback inside it?

alexandru: This part needs to be in the `synchronized` block as well, because the `synchronize` block ensures not only atomicity, but also *visibility*.

What can basically happen is for instructions to be reordered. Even if the `remaining == 0` test passes, you might get an inconsistent / incomplete state in `builder.result()`.

benchmarks/src/main/scala/monix/TaskGatherUnordered.scala
```diff
@@ -0,0 +1,186 @@
...
+                conn.pop()
+                finalCallback.onSuccess(builder.result())
+              }
+            }
+
+            def onError(ex: Throwable): Unit = {
+              lock.synchronized {
+                if (!composite.isCanceled) {
```
alexandru: Yeah, but I've been using another pattern. I basically initialize an atomic boolean ...

```scala
val isActive = Atomic(true)
```

And then I'm doing:

```scala
if (isActive.getAndSet(false)) {
  finalCallback.onError(ex)
} else {
  scheduler.reportError(ex)
}
```

By doing so there's no need for a `synchronize` block and that `getAndSet` is very efficient on Java 8 (from what I understood it's no longer doing a CMPXCHG).

monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
```diff
@@ -331,16 +332,33 @@ object Coeval {
...
+  /** Transforms a `TraversableOnce` of coevals into a coeval producing
+    * the same collection of gathered results.
     *
-    * For [[Coeval]] this has the same behavior as [[zipList]].
+    * It's a simple version of [[traverse]].
     */
-  def sequence[A](sources: Seq[Coeval[A]]): Coeval[List[A]] =
-    zipList(sources)
+  def sequence[A, M[X] <: TraversableOnce[X]](sources: M[Coeval[A]])
+                                             (implicit cbf: CanBuildFrom[M[Coeval[A]], A, M[A]]): Coeval[M[A]] = {
```
alexandru: Minor style issue, I know that IntelliJ IDEA indents code like this, but I don't like code indented that far to the right, because you end up with something looking like a grapes hanging of a branch of a tree or something. It's visually annoying imho. Shift this line to the left.

monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
```diff
@@ -331,16 +332,33 @@ object Coeval {
...
+  }
+
+  /** Transforms a `TraversableOnce[A]` into a coeval of the same collection
+    * using the provided function `A => Coeval[B]`.
+    *
+    * It's a generalized version of [[sequence]].
+    */
+  def traverse[A, B, M[X] <: TraversableOnce[X]](sources: M[A])
+                                                (f: A => Coeval[B])
+                                                (implicit cbf: CanBuildFrom[M[A], B, M[B]]): Coeval[M[B]] = {
```
alexandru: Here too, I'd rather have these 2 lines not indented like this.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
alexandru: That's fine. They are different anyway - the point of not guaranteeing an order is performance. Though I'm fairly certain that I'll never use `Vector` again :-)

The question is - should we return `Seq` from `gatherUnordered`? That would leave us with the freedom to change the internal implementation. But then again `Seq` is not liked in the Scala community from what I've heard, plus `List` is fairly good if you don't care about ordering and describes precisely what it is. So for now this is fine.

benchmarks/src/main/scala/monix/TaskGatherUnordered.scala
```diff
@@ -0,0 +1,186 @@
...
+  ): Task[M[A]] = {
+    Task.unsafeCreate { (scheduler, conn, finalCallback) =>
+      val lock = new Object()
+      val builder = cbf(in)
+      var remaining = in.size
+
+      val composite = CompositeCancelable()
+      conn.push(composite)
+
+      for (task <- in; if !composite.isCanceled) {
```
guersam: In this implementation, using `map` (`yield`) instead of `foreach` makes the benchmark hang forever and the inner block never runs. Do you have any clue? Maybe it's throwing an exception that is just swallowed somewhere...

alexandru:
@guersam looks pretty good. Don't worry about the delay - it's summer, we're taking vacations.

The pace has been slow as you can see, I've personally been one week to the sea side and on Monday I'm leaving again. Planning to stay fairly disconnected, though that never happens :-)

So version 2.0 has been integrated into our projects at work, no big surprises, we are almost at the point where it should be declared stable. But being July, I'm inclined to postpone that moment until everybody is back from vacation. So I'm moving the deadline to September 1st. 

Eagerly awaiting the benchmarks results :) And if you finalize this by this Sunday night, I can merge it on master, otherwise it'll have to wait until I get back.

benchmarks/src/main/scala/monix/TaskGatherUnordered.scala
alexandru: Btw - an error is an error. And even if we cannot stream `onError` anymore, at the very least we can log it by means of `scheduler.reportError`.

benchmarks/src/main/scala/monix/TaskGatherUnordered.scala
alexandru: Not sure. Maybe the `map` has lazy behavior, probably depending on what you give as input. If it has lazy behavior, then it won't execute, so there will be nothing to call the final callback.

But a `foreach` is better in this case anyway.

guersam:
Unfortunately I found that the benchmark was broken this morning because of a stack overflow in `gather` given a list of 10,000 elements. I guess `Task.mapBoth` is not stack safe. Observing some initial results, `gatherUnordered` was significantly faster than `gather` regardless of the implementation approach.

Instead, I ran it again in much smaller size to continue discussion and here's the result: https://gist.github.com/guersam/556a17d739aae99b972c350181413ee1

Non-blocking version performs better than blocking version by 2~8% in 4 threads, and the gap might be larger in more cores and threads.

I'm still not sure that the performance gap is large enough to choose non-blocking version over blocking one, because it restricts the result always to be a `List` and the user might not be happy with its performance characteristics in both time and space.

guersam:
Interestingly, moving the final callback part into the `synchronized` block in the blocking version increased its performace and now it's even slightly faster than the blocking version in 4 cores. I wonder how they will behave in larger machines.

guersam:
Reactive Streams Spec 313 [occasionally fails](https://travis-ci.org/monixio/monix/jobs/146550559#L6043-L6055), maybe there's a race condition or GC issue.

alexandru:
@guersam `mapBoth` should be stack safe. All the operations should be stack safe! I just verified and you are right, `mapBoth` is indeed not stack-safe. I'm fixing it, stand by.

On the "Reactive Streams Spec 313", the problem with the provided TCK that we need to pass is that it is resource intensive and the tests rely on JVM-controlled timeouts. So if you do other things in parallel, like compiling or testing other subprojects, you can trigger errors in that TCK because things get delayed when doing things in parallel. Monix's tests do not rely in general on such timeouts (because of `TestScheduler`) and in the very few exceptions where we used timeouts, we've been very generous with the specified timespan. Unfortunately I tried getting SBT to not do things in parallel anymore and I failed. I'd like SBT to not compile and test things in parallel, but it still does. I suspect there's some kind of problem with Scala.js, because if you'll take a look, the Scala.js tests are running in parallel. So don't worry about this one.

Thanks a lot for this btw :-)

guersam:
Thanks for the clarification. Regarding the TCK issue, I tried increasing the timeout from 300ms to 500ms via an environment variable and it seems working so far. Let's see if it still occurs in other builds after this commit merged.

alexandru:
@guersam I fixed it, see issue #194 along with PR #196.

What you need to do:

1. merge with master - you'll have a conflict in `gatherUnordered`, but ignore that code, just add your own thing
2. delete the benchmark, we no longer need it
3. for `gatherUnordered` settle on the version that does `synchronize` - it's easier to reason about it, doesn't seem to have a penalty compared with the other version and it's definitely better for Javascript
4. if we are going to use `synchronize`, at this point we might as well use `CanBuildFrom` here as well, so make the signature of `gatherUnordered` match that of `gather`, it's totally fine

guersam:
Done, maybe it was an overkill removing the benchmark entirely from the history though.

monix-eval/shared/src/test/scala/monix/eval/TaskTraverseSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+ * limitations under the License.
+ */
+
+package monix.eval
+
+import concurrent.duration._
+import scala.util.{Success, Failure}
+
+object TaskTraverseSuite extends BaseTestSuite {
+  test("Task.traverse should execute in parallel") { implicit s =>
```
alexandru: One small thing - traverse does NOT execute things in parallel. The test code says so as well - if the tasks would execute in parallel, they'd execute in 3 seconds and not 6.

You just need to rename this test - "should" -> "should not"

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -783,54 +801,57 @@ object Task extends TaskInstances {
...
     */
-  def gatherUnordered[A](in: Seq[Task[A]]): Task[Seq[A]] =
+  def gatherUnordered[A, M[X] <: TraversableOnce[X]](in: M[Task[A]])
+    (implicit cbf: CanBuildFrom[M[Task[A]], A, M[A]]): Task[M[A]] =
     Async { (scheduler, conn, finalCallback) =>
-      val atom = Atomic(Vector.empty[A])
-      val expected = in.length
+      var remaining = in.size
+      val builder = cbf(in)
+      val lock = new Object()
```
alexandru: In Scala this should be `new AnyRef`, `Object` being a Javaism :-P

guersam: Indeed! Maybe I've been too functional so that I forgot about `AnyRef` completely :)

monix-eval/shared/src/test/scala/monix/eval/TaskTraverseSuite.scala
guersam: Nice catch :)

guersam:
There's [another Travis failure](https://travis-ci.org/monixio/monix/jobs/146590830#L13498) which seems to be an SBT issue.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -783,54 +801,57 @@ object Task extends TaskInstances {
...
     * results will be ordered. Useful when you don't need ordering because it
     * can be more efficient than `gather`.
     */
-  def gatherUnordered[A](in: Seq[Task[A]]): Task[Seq[A]] =
+  def gatherUnordered[A, M[X] <: TraversableOnce[X]](in: M[Task[A]])
+    (implicit cbf: CanBuildFrom[M[Task[A]], A, M[A]]): Task[M[A]] =
     Async { (scheduler, conn, finalCallback) =>
-      val atom = Atomic(Vector.empty[A])
-      val expected = in.length
+      var remaining = in.size
```
alexandru: Dude, I think here we might have a problem. If we have a true `TraversableOnce`, it means we can traverse it only once, being the equivalent of an `Iterator`. And this signature allows it.

One proposal would be to just convert it into a `Seq` before using it, since a `Seq` can be traversed multiple times and the normal usecase will be for people to pass us Seqs anyway.

Another possible but more complicated solution that does traverse that only once would be:

1. we initialize `var remaining = 1` (one extra unit, because we don't want to trigger the final callback too early)
2. we place the while loop inside a `lock.synchronize` as well
3. we do `remaining += 1` on each cycle, before running the task
4. at the end of the loop, but before we leave the `synchronize`, we do `remaining -= 1`, but at this point the result might be zero already, so we need to trigger the `finalCallback` doing the `if (remaing == 0)` check just like in `onSuccess`

guersam: You're right, how did I miss it... given the while loop has obtained the lock, the inner `synchronize` block won't run until the whole loop is finished. Then just adding 1~3 with just `var remaining = 0` would be sufficient. What do you think?

alexandru: Don't worry, we don't usually work with TraversableOnce, that's why :-)

Doing `var remaining = 0` is not sufficient because `lock.synchronize` is reentrant and some tasks can execute immediately on the same thread (e.g. `evalAlways`, `evalOnce`, `now`). And we are starting these tasks with `unsafeStartNow`, so serial execution is possible. If we would have started them with `unsafeStartAsync`, then your point would be correct.

guersam: Understood, thanks for the kind explanation :smiley:

alexandru:
Looks good, it can probably be merged. Anything else you'd like to do?

guersam:
Nope, I'm fine to go.

Additionally, as the implementations of `gather` and `gatherUnordered` changed since the benchmark, I'll run full benchmark again over the weekend for the record. I'm also curious of the performance benefit of `gatherUnordered` over `gather`.

guersam:
The build failure is quite annoying... Could you try flushing travis cache?
I'm not sure that I have the permission.

alexandru:
@guersam I deleted all caches. Lets see what it does.

guersam:
Thanks! The SBT issues is fixed, but sadly the TCK spec still fails on master :( Maybe we need much longer timeout such as 3 seconds.

alexandru:
@guersam BTW, if you want, contribute a benchmark comparing gather and gatherUnordered. I'm also curious how we perform against Scalaz - for normal operations (map, flatMap) we are on par for synchronous tasks and for async tasks we are kicking their ass :-) so we need to be at least similar in throughput, otherwise we are doing something wrong.

PR #192, opened 2016-06-29
Pad2Observable

leakyabstraction:
@alexandru  Please by no means merge this PR, as test still fails from time to time. The other thing I kind of dislike is the boxing and unboxing of items emitted by child observables using `Coeval` in the implementation. Should we do something about it ?


alexandru:
@leakyabstraction what's the status on `PadObservable`? Does it work, does it need anything else?

alexandru:
@leakyabstraction closing this for now, as I'm cleaning up the issue tracker. If you want to keep working on this, then you can re-open another PR.

PR #200, opened 2016-08-02
Execute an Action (() => Unit)

aoprisan:
Added an execute overload which takes an Action (() => Unit), wraps it in RunnableAction and forwards it to the default execute method.


alexandru:
- changed name, because this particular override is problematic as it can conflict with the main method, because of that by-name parameter - in all other cases the arity of the methods are different and so you can't end up with a conflict - this is related to by-name params, I would have been fine with a `Function1`, but then that wouldn't have been consistent
- added a simple test

PR #208, opened 2016-08-15
Uniquely name threads generated from ThreadFactory

### LINKED ISSUE #204: Include number in name of threads
When using [Scheduler.fixedPool](https://github.com/monixio/monix/blob/13021b735320ecd76581158716557df0d7d1dcf4/monix-execution/jvm/src/main/scala/monix/execution/schedulers/SchedulerCompanionImpl.scala#L305) and similar methods that create a ThreadPool, all threads in the pool have the same name.

Ideally the threads should have different names to assist with investigating issues when looking at the logs.

Suggest a name contention of `{poolName}-number` where number is incremented each time a thread is created.

If there are no objections, I'll submit a PR for this change.

lukestephenson:
Fixes #204

### DIFF
--- a/monix-execution/jvm/src/main/scala/monix/execution/schedulers/SchedulerCompanionImpl.scala
+++ b/monix-execution/jvm/src/main/scala/monix/execution/schedulers/SchedulerCompanionImpl.scala
@@ -18,9 +18,8 @@
 package monix.execution.schedulers
 
 import java.lang.Thread.UncaughtExceptionHandler
-import java.util.concurrent.{Executors, ScheduledExecutorService, ThreadFactory}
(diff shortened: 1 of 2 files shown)


lukestephenson:
Not sure why the PR build is failing. `sbt test` locally runs fine.

alexandru:
Thanks @lukestephenson, looks good. Don't worry about the tests. The tests suite contain some things that depend on timeouts and it's quite resource intensive, so they can trigger errors in Travis. Quite annoying, will probably have to tune those timeouts.

lukestephenson:
Thanks @alexandru - when do you plan to merge this?

PR #240, opened 2016-10-09
Change monix-types encoding, provide tailRecM, upgrade Cats

### LINKED ISSUE #232: Task should use TrampolinedRunnables everywhere it can
The optimization trick we've been doing with `LocalRunnable` / `LocalBatchingExecutor` works well and should be taken to the next level. But this means the trampoline needs to do special handling for `nextFrameIndex`.

As a reminder, the `FrameIndex` is a simple `Int` that gets incremented on each loop iteration of Task's run-loop implementation. And depending on the `ExecutionModel` injected by the `Scheduler` implementation, once every X iterations (currently set to 1024 iterations) we force an asynchronous boundary, with the purpose of not keeping the current thread occupied forever.

However, in version 2.0.0 we had to reset this `FrameIndex` on every `Async` Task. But now that we have thread-local asynchronous execution capabilities provided by the `Scheduler`,  we have a problem, since thread-local execution is resetting our `FrameIndex` counter without having real asynchronous boundaries.

The plan is:

- renamings for clarity:
  - `LocalRunnable` to `TrampolinedRunnable` for clarity
  - `LocalBatchingExecutor` to `BatchingTrampolineExecutor` 
  - the extension `scheduler.executeLocal` to `scheduler.executeTrampolined`
- introduce `scheduler.executeAsyncBatch` as another macros-enabled extension method, which forces a real asynchronous boundary (much like `scheduler.executeAsync`), but then starts trampolined execution, as an optimization, since the very first `executeTrampolined` incurs the penalty of invoking the logic from Scala's `BlockContext`
- the `frameIndex` in the Task trampoline needs special handling, therefore we need to introduce `ThreadLocal` values (see [issue 237](https://github.com/monixio/monix/issues/237))
- changed the signature of the `OnFinish` in `Task.unsafeCreate`; unfortunately this doesn't admit a backwards-compatible solution

So basically the `OnFinish` callback when creating asynchronous `Task` instances, changes like this:

```scala
type FrameIndex = Int

/** Type alias representing callbacks for [[unsafeCreate asynchronous]] tasks. */
type OnFinish[+A] = (Scheduler, StackedCancelable, ThreadLocal[FrameIndex], Callback[A]) => Unit

private[eval] def startFrameRef(): ThreadLocal[FrameIndex] =
  ThreadLocal(1)
```

So when evaluating a `Task` the starting `FrameIndex` is `1`. This was also true in version `2.0.0`. But now when creating a `Task` with `unsafeCreate`, we receive this `FrameIndex` and in case of an actual asynchronous boundary we have to `reset()` it:

So in this case we don't have to do anything, since no real asynchronous boundary happens:

```scala
Task.unsafeCreate { (scheduler, subscription, frameIndex, callback) =>
  callback.asyncOnSuccess("hello")(scheduler)
}
```

But in this case we have to reset the `FrameIndex` because we have a real asynchronous boundary:

```scala
Task.unsafeCreate { (scheduler, subscription, frameIndex, callback) =>
  scheduler.executeAsync {
    frameIndex.reset()
    callback.asyncOnSuccess("hello")(scheduler)
  }
}
```

The result of all this effort is to get rid of unwanted asynchronous boundaries.

alexandru:
- Issue #250: Change the `monix-types` encoding for the exposed type-classes, add laws and tests
- Issue #254: Add `Observable.tailRecM`, `Task.tailRecM` and `Coeval.tailRecM`
- Issue #258: Add executeWithFork, executeWithModel on Observable
- Issue #260: Upgrade Cats to 0.8.0


alexandru:
CC: @aoprisan please take a look. Thanks!

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -656,49 +752,41 @@ object Task extends TaskInstances {
...
-        catch {
-          case NonFatal(ex) =>
-            // We cannot stream the error, because the callback might have
-            // been called already and we'd be violating its contract,
-            // hence the only thing possible is to log the error.
-            scheduler.reportFailure(ex)
-        })
-    }
-  }
+  def create[A](register: (Scheduler, Callback[A]) => Cancelable): Task[A] =
```
alexandru: Comment is now incorrect as callback is trampolined

monix-eval/shared/src/main/scala/monix/eval/internal/TaskFromFuture.scala
```diff
@@ -0,0 +1,60 @@
...
+      case c: Cancelable =>
+        // Cancelable future, needs canceling
+        Task.unsafeCreate { (context, cb) =>
+          import context.implicitScheduler
+          // Already completed future?
+          if (f.isCompleted) cb.asyncApply(f.value.get) else {
+            val conn = context.connection
+            conn.push(c)
+            f.onComplete { result =>
+              conn.pop()
```
alexandru: Needs frame reset

monix-eval/shared/src/main/scala/monix/eval/internal/TaskFromFuture.scala
```diff
@@ -0,0 +1,60 @@
...
+          }
+        }
+      case _ =>
+        // Simple future, convert directly
+        Task.unsafeCreate { (context, cb) =>
+          import context.implicitScheduler
+          if (f.isCompleted)
+            cb.asyncApply(f.value.get)
+          else
+            f.onComplete(cb)
```
alexandru: Needs frame reset.

Issue #247, opened 2016-10-18
Avoid runtime reflection


Taig:
In version 2.0.3 runtime reflection was [introduced](https://github.com/monixio/monix/blob/v2.0.3/monix-execution/jvm/src/main/scala/monix/execution/schedulers/LocalBatchingExecutor.scala#L126) which caused my ProGuard optimized Android app to crash without further manual configuration. If possible, it would be desirable to avoid runtime reflection entirely.

alexandru:
@Taig This is released in latest version 2.0.5

PR #248, opened 2016-10-18
Reset overflow counter on None onOverflow result

ddworak:
Hey! 
I know that according to the contribution guidelines I should've opened an issue, but this is just a short bugfix. I will be happy to do it in case you decide it's required here.
I believe that current behaviour of Drop*AndSignal wasn't intended and seems to have been introduced just recently. The docs state:
> The function can return `None` in which case no message is sent and thus you can use it just to log a warning.

yet I've noticed that after an overflow the callback is called for each new message.
I've modified the tests to check for this behaviour and proposed a solution.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,3 +15,6 @@ https://github.com/aoprisan
 
 Jisoo Park
 https://github.com/guersam
+
+Dawid Dworak
+https://github.com/ddworak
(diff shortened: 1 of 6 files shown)


alexandru:
Thanks a lot @ddworak, stupid bug, stupid test.

Thanks for catching and fixing it. Scala.js also has released a new version, I guess it is time for another minor release. After also fixing #247.

alexandru:
@ddworak this is released in latest version 2.0.5

Issue #249, opened 2016-10-20
Task.traverse does not have a nondeterministic analogue


mutantmell:
The Future api in the base library provides a parallel traverse, and Task should have a parallel version as well.

Task already has a parallel and a sequential implementation of sequence: 

parallel: https://github.com/monixio/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/Task.scala#L886
sequential: https://github.com/monixio/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/Task.scala#L852

monix should be able to use the same basic scheme to provide a parallel traverse.

alexandru:
@mutantmell to tell you the truth, I don't really see the point in `traverse` at all, at least in its current incarnation.

I mean, it has value in its current incarnation when the given `M[_]` collection has lazy behavior (e.g. Stream or Iterable) and thus you avoid evaluating the current `f(a): Task` until needed. But then a version that does things in parallel needs to evaluate the whole thing at once, so that argument is gone.

Could be useful, but now I'm not seeing it. Could you provide a use-case for a parallel traverse?

alexandru:
@mutantmell any further thoughts on this matter? Do you have a use-case for a parallel traverse?

mutantmell:
We are currently using the stdlib Future library for most of our code, and many of the frameworks we use (e.g. akka http, slick, etc.) return/require Futures.  We are very i/o heavy in our service, with lots of blocking (waiting for data to return from a database, waiting for the result of a web call, etc.).  We use Future.traverse to mitigate this -- while the first request in a list is blocked, we start evaluating the second value, etc, allowing us to overlap the blocking calls as much as possible.

Adding a parallel traverse to monix would help assist a transition from just stdlib Futures to using monix Tasks - we could wrap Futures in Tasks where libraries provide them, and still have the ability to overlap blocking calls.

alexandru:
@mutantmell I understand, but assuming you have an `M[A]` list that you want to convert into a `Task`, I guess you're doing something like this:

```scala
def f(a: A): Future[B] = ???
val list: List[A] = ???

val future: Future[List[B]] =
  Future.traverse(list)(f)
```

So my question is, why not do this instead:

```scala
def f(a: A): Task[B] = ???
val list: List[A] = ???

val task: Task[List[B]] =
  Task.gather(list.map(f))
```

Isn't it the same thing?

[...]

PR #267, opened 2016-11-08
Added wander/wanderUnordered to Task.

mutantmell:
This should satisfy Issue #249.

I ended up modifying the backing implementation of `TaskGather` to requiring multiple CanBuildFroms for gatherUnordered.

I added a first pass at the Scaladoc, based heavily on the Scaladoc for gather(unordered).

Let me know if there's any questions/comments/concerns.

-Matt

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -896,7 +896,7 @@ object Task extends TaskInstances {
    */
   def wander[A, B, M[X] <: TraversableOnce[X]](in: M[A])(f: A => Task[B])
     (implicit cbf: CanBuildFrom[M[A], B, M[B]]): Task[M[B]] = {
-    TaskGather(in.toArray.map(f), () => cbf(in))
+    TaskGather(in.map(f).toArray, () => cbf(in))
   }
 
   /** Nondeterministically gather results from the given collection of tasks,
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -875,7 +875,7 @@ object Task extends TaskInstances {
     */
   def gather[A, M[X] <: TraversableOnce[X]](in: M[Task[A]])
     (implicit cbf: CanBuildFrom[M[Task[A]], A, M[A]]): Task[M[A]] =
-    TaskGather(in.toArray, () => cbf(in))
+    TaskGather(in, () => cbf(in))
 
    /** Given a `TraversableOnce[A]` and a function `A => Task[B]`,
    * nondeterministically apply the function to each element of the collection
@@ -895,9 +895,8 @@ object Task extends TaskInstances {
    * It's a generalized version of [[gather]].
    */
   def wander[A, B, M[X] <: TraversableOnce[X]](in: M[A])(f: A => Task[B])
-    (implicit cbf: CanBuildFrom[M[A], B, M[B]]): Task[M[B]] = {
-    TaskGather(in.map(f).toArray, () => cbf(in))
-  }
+    (implicit cbf: CanBuildFrom[M[A], B, M[B]]): Task[M[B]] =
+    Task.eval(in.map(f)).flatMap(col => TaskGather(col, () => cbf(in)))
 
   /** Nondeterministically gather results from the given collection of tasks,
     * without keeping the original ordering of results.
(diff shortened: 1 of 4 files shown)


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -875,7 +875,29 @@ object Task extends TaskInstances {
     */
   def gather[A, M[X] <: TraversableOnce[X]](in: M[Task[A]])
     (implicit cbf: CanBuildFrom[M[Task[A]], A, M[A]]): Task[M[A]] =
-    TaskGather(in)(cbf)
+    TaskGather(in.toArray, () => cbf(in))
```
alexandru: Problem with this `.toArray` is that depending on the implementation of `M[A]`:

1. this code can throw exceptions, leading to a different result than the `TaskGather` that took `in` directly (see that the original `.toArray` was inside that `try/catch`)
2. a `.toArray` on a large data-structure or on some Java Iterable that blocks threads can be an expensive operation (I know it sounds awful, but there are some truly awful Java implementations out there) and by not processing this in the context of the task, the user can get no opportunity to process it on a different thread

It's also why the second parameter is a `Function1`. My suggestion is to make `TaskGather` take a `TraversableOnce[Task[A]]` parameter and move the `.toArray` inside `TaskGather` again. Or maybe make this parameter lazy as well?

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -875,7 +875,29 @@ object Task extends TaskInstances {
...
+   *
+   * Although the effects are unordered, we ensure the order of results
+   * matches the order of the input sequence. Also see [[wanderUnordered]]
+   * for the more efficient alternative.
+   *
+   * It's a generalized version of [[gather]].
+   */
+  def wander[A, B, M[X] <: TraversableOnce[X]](in: M[A])(f: A => Task[B])
+    (implicit cbf: CanBuildFrom[M[A], B, M[B]]): Task[M[B]] = {
+    TaskGather(in.map(f).toArray, () => cbf(in))
```
alexandru: I would have done it like ... 

```scala
def wander[A, B, M[X] <: TraversableOnce[X]](in: M[A])(f: A => Task[B])
  (implicit cbf: CanBuildFrom[M[A], A, M[A]]): Task[M[B]] = {

  Task.eval(in.map(f)).flatMap(col => TaskGather(col, () => cbf(in)))
}
```

This way `in.map(f)` is protected from thrown exceptions and you don't have to turn it into an array.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -895,6 +917,30 @@ object Task extends TaskInstances {
...
+    * results will be ordered. Useful when you don't need ordering because:
+    *
+    *  - it has non-blocking behavior (but not wait-free)
+    *  - it can be more efficient (compared with [[wander]]), but not
+    *    necessarily (if you care about performance, then test)
+    *
+    * It's a generalized version of [[gatherUnordered]].
+    */
+  def wanderUnordered[A, B, M[X] <: TraversableOnce[X]](in: M[A])(f: A => Task[B])
+    (implicit cbf: CanBuildFrom[M[A], B, M[B]]): Task[M[B]] = {
```
alexandru: I would not go with a version with `CanBuildFrom` and simply follow the normal `gatherUnordered` signature.

This is new in master and the version 2.1 I'm going to release (probably today), but the reason is that we are no longer using `CanBuildFrom` in `gatherUnordered` and returning a `Task[List[A]]` instead is because it better reflects what the function does, because the result really is a `List`, being kept in an `Atomic` and updated in a non-blocking fashion and if the user wants something else, he can convert that `List` himself, with whatever cost that incurs and that cost will be visible, whereas if you hide it in this function, that cost will not be visible.

So basically traversing that `List` to build the collection you want is not worth it imho. It's also the defining characteristic versus normal `gather`. I mean originally `gather` was built on top of `gatherUnordered`, following Scalaz in this regard, but now `gather` has a pretty efficient version too, so the defining characteristics of `gatherUnordered` are:

1. non-blocking
2. more memory efficient, since we traverse the original collection only once and build the final `List` as we go along, only once

Let me know your thoughts on this matter.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
mutantmell: I changed `TaskGather` to take a `TraversableOnce[Task[A]]` parameter - this helped address the next issue as well.

(the `.toArray` being there is an artifact of a previous draft, when the code still took in `CanBuildFrom` instead of a builder maker)

monix-eval/shared/src/main/scala/monix/eval/Task.scala
mutantmell: Done

monix-eval/shared/src/main/scala/monix/eval/Task.scala
mutantmell: No objections here

mutantmell: Im adding tests to verify these properties, then I'll update the pr

alexandru:
Thanks @mutantmell - version 2.1 is already released on Maven Central, your PR will make it in the next release.

PR #277, opened 2016-12-11
be friendly to the Scala 2.13 community build

SethTisue:
Scala 2.13 is only just getting underway, but nonetheless, it would
be nice to have this change in


alexandru:
Thanks @SethTisue

PR #280, opened 2016-12-21
WIP: pull-based streaming based on Task / Coeval

alexandru:
We are introducing a new `Iterant[F[_], A]` type in a new sub-project called `monix-tail`, that's going to be a pull-based alternative to `Observable`, with a customizable `F[_]` evaluation context and thus powered by `Task`, `Coeval` or other monadic types.

Rx.NET has `IAsyncEnumerable<T>` and `IAsyncEnumerator<T>`. The `Enumerable` (`Iterable`) type is considered the [dual](https://en.wikipedia.org/wiki/Duality_(mathematics)) of the `Observable` implementation and `IAsyncEnumerator` is basically an `Enumerator` that returns a `Future` (or `Task`) when calling `next()`.

We need a similar pull-based dual because:

1. a pull-based communication model is many times easier to reason about for certain use-cases
2. it's easier to provide support for certain FP-ish operations, like `foldRight`, which isn't possible to implement for `Observable` (without converting it into a pull-based type first, hence this new type)
3. `Observable` will still be much better for performance and for time-based / reactive operations; actually this new type will not implement operations that are non-deterministic in nature, e.g. it will have `zip`, but not `combineLatest` ;-)
4. for people wanting a replacement for Scala's own `scala.collection.immutable.Stream` type, the `Observable` type might seem to be a little too much, since it is oriented towards reactive stuff; but *conversions* back and forth will be seamless

However the implementation will be different compared to an `Iterable` / `Enumerator`: our implementation will be an example of functional programming in Scala, thus following principles of functional programming, building on the knowledge we acquired and versatility of `Task`, so our communication protocol will not be based on Java's `Iterator`.

## Design

Basically our `Iterant[F[_], A]` is modeled as a bunch of states, similar to `Task` and `Free`, inspired by the `List` implementation (in Scala, Haskell and others) and by the `Streaming` type in Dogs.

The minimal/elegant encoding, with which we can achieve every operation that we want is this:

```scala
sealed abstract class Iterant[F[_], A]

case class Next[F[_], A](
    item: A,
    rest: F[Iterant[F, A]],
    stop: F[Unit])
    extends Iterant[F ,A]

case class Suspend[F[_], A](
    rest: F[Iterant[F,A]],
    stop: F[Unit])
    extends Iterant[F,A]

case class Halt[F[_], A](ex: Option[Throwable])
    extends Iterant[F, A]
```

So this is very similar with Scala's `List`, except that:

- the `rest` in `Next` (the equivalent of the `tail` in Scala's `List`) is an `F[A]`, which can be a `Task` or a `Coeval`, meaning that its evaluation can also be lazy and/or asynchronous
- the stream can end in an error (`Halt(Some(exception))`)
- we need to handle file handlers and network sockets, hence we need a `stop: F[Unit]` routine that is supposed to be called whenever we want to prematurely stop the stream processing

So basically with a Scala `List` at any point the only choice is to keep processing the `rest` tail reference. But our stream gives us a choice: either process the `tail` or `stop`.

## Optimisations

The above encoding is minimal, but not efficient. We also need:

```scala
case class NextCursor[F[_], A](
  cursor: Cursor[A],
  rest: F[Iterant[F, A]],
  stop: F[Unit])
  extends Iterant[F, A]
```

The `Cursor` type is a light alternative to Scala's and Java's `Iterator`, which
can wrap either a normal `Iterator`, an `Array` or whatever can be iterated
synchronously and efficiently. This is because:

1. performance of heap-allocated linked-lists, especially linked lists evaluated by means of a trampolined call-stack is absolutely terrible
2. IMO any streaming abstraction sucks badly if it cannot iterate over arrays efficiently (so without trashing cache locality or the garbage collector)
3. the `Iterator` pattern is very, very efficient and best of all it supports our needs for efficient head / tail decomposition, even if it is destructive when compared with Scala's `LinearSeq` (e.g. `List`, `Stack`), because linked-lists suck for performance

The `Cursor` type is basically an `Iterator`, however I decided to introduce
a custom type because I want to control its implementation and because `Iterator`
does not have the right semantics for the provided transformations. For instance
when working with arrays, I want `Cursor.map` and `Cursor.filter` and other
such operations to have strict behavior, not lazy. I also want `Cursor` to have specialised implementations for arrays of primitives.

So YES, I'm optimising this FP abstraction by shoving an `Iterator` in one of these states. It's dirty, but it is well encapsulated and it works.

But that's not all, we also need:

```scala
case class Last[F[_], A](item: A)
  extends Iterant[F, A]
```

This is an optimisation over signalling `Next(item, F.pure(Halt(None)), F.unit)`,
which in benchmarking shows that it greatly improves the performance of `flatMap`.
Basically without this state, the `flatMap` operation is slower than the one
implemented on `Observable` and that would be quite bad, given that
`Observable.flatMap` is forced to do concurrency handling due to the
back-pressure protocol handling of `onComplete` (currently by means of 2
`getAndSet` operation per concatenated child).

## Work in progress

A lot of stuff is not implemented yet. I've been cautious about writing untested code. In order to release this in 2.3.0, we need good test coverage, so tests are being written in parallel with each added function.


edmundnoble:
I think the namespace collision with the Predef is enough reason to keep off calling it `Stream`, but it's not like I have a better name :)

One question: do all stop actions get chained when the stream is prematurely ended? I'm looking at `foldLeftL` and it looks like it only executes the latest stream step's stop action on failure; that seems to mean you cannot open resources and keep them open for several stream steps.

alexandru:
@edmundnoble no, stop actions don't get chained; usually the same reference is supposed to be given repeatedly, which on evaluation is supposed to call `handler.close()` or `resource.shutdown()` or something. These `Cons` events are generated in a loop, so you're supposed to give the `stop` that's relevant for the currently opened resources and only the current one is relevant. And in case repeated evaluation is dangerous, the given `Task` (or whatever) can be made to be idempotent.

There can be complications, but these are solved in the operators. For example in the `++` operator (logic that's also used by `flatMap`) you might need to stop both the current stream and the next one, since the next one might have already been started, so we make sure to build a `composite`:

```scala
private final def concatF[B >: A](rhs: F[Stream[F,B]])(implicit F: Monad[F]): Stream[F,B] = {
  import F.{applicative, functor}

  this match {
    case Cons(a, lt, stop) =>
      val composite = stop.flatMap(_ => F.flatMap(rhs)(_.onStop))
      Cons[F,B](a, lt.map(_.concatF(rhs)), composite) 
  ...
```

A composite would be needed for a `zip` operator as well. I think it works fine, I tested it for a couple of use-cases, but I'm not entirely 100% sure that I'm not missing anything.

I'm not sure if that answers your question.

On the naming, yes, I think the conflict is too great, plus now we've got `java.util.stream.Stream` as well, so that's too much.

edmundnoble:
That answers my question, thank you :)

alexandru:
@edmundnoble I saw you just worked on a `Cofree` implementation for Cats.

I didn't know about what that is, apparently I need some catching up to do, but by the looks of it, it's sort of similar. Any thoughts on that?

This type is inspired by the `Streaming` type in the Dogs project and its "innovation" will be in resource handling. If you see the `Cofree` type evolving towards that, then my own implementation will serve no purpose. On the other hand I see that `Cofree` is a pretty simple implementation and that may be its virtue; adding stuff to it might not be desirable by all.

So what do you think? Does this type look interesting in Monix, should we proceed?

edmundnoble:
@alexandru `Cofree` does not have a notion of resource handling, as you've noted. It's also likely to be the wrong abstraction, because given a `Cofree[F, A]` you can extract a value of type `A` and one of type `F[Cofree[F, A]]`, similar to the `Cons` case for `Stream` but unable to represent `Suspend` or `Halt`. Many non-empty recursive data structures are isomorphic to `Cofree[F, A]` for some `F`, but if at any point the structure does not have an `A` to yield `Cofree` becomes the wrong structure.

The dual to `Cofree` which may be empty is `Free`, so what you are looking for will end up representable with `Free[F, A]` (for a sufficiently complicated `F[_]`, and `Option[Throwable]` chosen for `A` to represent `Halt`), but the `Free` machinery we get won't provide tools to usefully interact with that type, especially not safely given resources, so you'd still have to code up some awesome API. For example:

```scala
case class StreamAlgebra[F[_], A, B](next: F[A], newValue: Option[B], stop: F[Unit])
type Stream[F[_], A] = Free[Algebra[F, ?, A], Option[Throwable]]
def stop[F[_], A](ex: Option[Throwable]): Stream[F, A] = 
  Free.pure(ex)
def suspend[F[_], A](rest: F[Stream[F, A]], stop: F[Unit]) = 
  Free.roll(Algebra(rest, None, stop))
def cons[F[_], A](rest: F[Stream[F, A]], value: A, stop: F[Unit]) = 
  Free.roll(Algebra(rest, Some(value), stop))
```

I do not think this encoding is particularly practical, however, if all it will give us is unsafe tools.

alexandru:
Changes:

1. renamed `Stream` to `Iterant`, so we now have `Iterant[F[_], +A]`
2. `AsyncStream` is now just a type alias for `Iterant[Task, +A]`, along with a companion object for useful builders, since the old approach adds too much overhead and doesn't play nice with type-class based design 
3. ditto for `LazyStream` now being just a type alias for `Iterant[Coeval, +A]`, along with a companion object of builders

I've also made the builders nicer. So you can now do stuff like this (without overhead):
```scala
Iterant[Task].range(0,100)

// ... versus ...

Iterant.range[Task,Int](0, 100)
```

And by using this new builder style, we can also expose builders specific for `Iterant[Task,A]` (with the type returned being the `AsyncStream` alias), which is nice. Using Scala's type-system for the win.

monix-iterant/shared/src/main/scala/monix/iterant/Iterant.scala
```diff
@@ -0,0 +1,623 @@
...
+  * controlled by an `F[_]` monadic type that you have to supply
+  * (like [[monix.eval.Task Task]] or [[monix.eval.Coeval Coeval]])
+  * which will control the evaluation. In other words, this `Iterant`
+  * type is capable of strict or lazy, synchronous or asynchronous
+  * evaluation.
+  *
+  * Consumption of a `Iterant` happens typically in a loop where
+  * the current step represents either a signal that the stream
+  * is over, or a (head, rest) pair, very similar in spirit to
+  * Scala's standard `List` or `Iterant`.
```
igstan: "standard `List` or `Iterant`." I presume you meant `Iterator` here, not `Iterant`.

alexandru: Yes, a search & replace gone wrong :)

README.md
```diff
@@ -18,7 +18,7 @@ composing asynchronous and event-based programs using observable sequences
 that are exposed as asynchronous streams, expanding on the
 [observer pattern](https://en.wikipedia.org/wiki/Observer_pattern),
 strongly inspired by [ReactiveX](http://reactivex.io/),
-but designed from the ground up  for back-pressure and made to cleanly interact
+but designed from the ground up  for back-pressure and made to cleanly iterant
```
edmundnoble: Runaway find/replace? Should this still be interact?

alexandru: :-)) yeah, freaking IntelliJ IDEA, it's a bad idea to "search in comments and strings"

alexandru:
Resuming work on this feature 😃 I've merged with `master`, which contains the upgrade to Cats `1.0.0-MF`.

alexandru:
I've merged this PR in master, even though the `Iterant` type isn't finished, but it provides a good base to build on and this PR was getting too large already.

PR #282, opened 2016-12-25
Parallel consumer by partition

Fristi:
As I asked here: https://gitter.im/monix/monix?at=58341ba79a5479d048483b04.

I would like to consume a stream of values in parallel by partitioning the value. The values are buffered internally in buckets and each bucket is consumed by `PartitionedConsumer` which includes the partition and the elements which were buffered.

I've used this in my Kafka driver. See: https://github.com/vectos/flumina/blob/feature/monix/monix/src/main/scala/flumina/monix/package.scala#L15 (slighty different API, but will merge to this API when it gets merged).

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
@@ -19,14 +19,10 @@ package monix.reactive
 
 import cats.effect.Effect
 import monix.eval.{Callback, Task}
-import monix.execution.Ack.Continue
-import monix.execution.atomic.AtomicBoolean
-import monix.execution.cancelables.{AssignableCancelable, RefCountCancelable}
+import monix.execution.cancelables.AssignableCancelable
 import monix.execution.{Cancelable, Scheduler}
-import monix.reactive.exceptions.CompositeException
 import monix.reactive.internal.consumers._
 import monix.reactive.observers.Subscriber
-import monix.reactive.subjects.PublishSubject
 
 /** The `Consumer` is a specification of how to consume an observable.
   *
@@ -414,86 +410,5 @@ object Consumer {
     */
   trait Sync[-In, +R] extends Consumer[In, R] {
     override def createSubscriber(cb: Callback[R], s: Scheduler): (Subscriber.Sync[In], AssignableCancelable)
-  }
-
-  /** Implementation for [[Consumer.partitionParallel]] */
-  private final class PartitionParallelConsumer[A](nrPartitions: Int, overflowStrategy: OverflowStrategy.Synchronous[A], partition: A => Int, consumer: (Int, Seq[A]) => Task[Unit]) extends Consumer[A, Unit] {
-    override def createSubscriber(cb: Callback[Unit], s: Scheduler): (Subscriber[A], AssignableCancelable) = {
-
-      val errors = scala.collection.mutable.ArrayBuffer.empty[Throwable]
-      val sac = SingleAssignmentCancelable()
-
-      val publishSubject = PublishSubject[A]()
-      val stream = publishSubject.whileBusyBuffer(overflowStrategy)
-      val isUpstreamDone = AtomicBoolean(false)
-
-      val refCount = RefCountCancelable { () =>
-        sac.cancel()
-        if (errors.nonEmpty) {
-          cb.onError(CompositeException(errors))
-        } else {
-          cb.onSuccess(())
-        }
-      }
-
-      def subscriber(part: Int): Cancelable = {
-
-        def consumerSubscriber(ref: Cancelable) = new Subscriber[Seq[A]] {
-          implicit def scheduler: Scheduler = s
-
-          def onNext(elements: Seq[A]) =
-            try consumer(part, elements).coeval.value match {
-              case Left(future) =>
-                future.map(_ => Continue)
-              case Right(()) =>
-                Continue
-            } catch {
-              case NonFatal(ex) =>
-                errors += ex
-                //cancel the partition cancellation token
-                ref.cancel()
-                //only cancel the "main" cancellation token once
-                if (!isUpstreamDone.getAndSet(true)) {
-                  refCount.cancel()
-                }
-                Stop
-            }
-
-          def onError(ex: Throwable) = ()
-
-          def onComplete() = {
-            //cancel the partition cancellation token
-            ref.cancel()
-            //only cancel the "main" cancellation token once
-            if (!isUpstreamDone.getAndSet(true)) {
-              refCount.cancel()
-            }
-          }
-        }
-
-        stream
-          .filter(x => partition(x) % nrPartitions == part)
-          .bufferIntrospective(s.executionModel.recommendedBatchSize)
-          .subscribe(consumerSubscriber(refCount.acquire())) // per partition we create a cancellation token
-      }
-
-      val out = new Subscriber[A] {
-        implicit def scheduler: Scheduler = s
-
-        def onNext(elem: A): Future[Ack] = publishSubject.onNext(elem)
-
-        def onError(ex: Throwable): Unit = {
-          //upon receiving an error from upstream, cancel immediately and signal this error once
-          sac.cancel()
-          cb.onError(ex)
-        }
-
-        def onComplete(): Unit = publishSubject.onComplete()
-      }
-
-      (0 until nrPartitions).foreach(subscriber)
-
-      out -> sac
-    }
   }
 }
--- a/monix-execution/shared/src/main/scala/monix/execution/exceptions/CompositeException.scala
+++ b/monix-execution/shared/src/main/scala/monix/execution/exceptions/CompositeException.scala
@@ -34,6 +34,15 @@ class CompositeException(val errors: Seq[Throwable])
         "(" + reasons + ")"
       })
   }
+
+  def canEqual(other: Any): Boolean = other.isInstanceOf[CompositeException]
+
+  override def equals(other: Any): Boolean = other match {
+    case that: CompositeException =>
+      (that canEqual this) &&
+        errors == that.errors
+    case _ => false
+  }
 }
 
 object CompositeException extends AbstractFunction1[Seq[Throwable], CompositeException] {
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/consumers/PartitionParallelConsumer.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/consumers/PartitionParallelConsumer.scala
@@ -1,0 +1,96 @@
+package monix.reactive.internal.consumers
+
+import monix.eval.{Callback, Task}
+import monix.execution.Ack.{Continue, Stop}
+import monix.execution.{Ack, Cancelable, Scheduler}
+import monix.execution.atomic.AtomicBoolean
+import monix.execution.cancelables.{AssignableCancelable, RefCountCancelable, SingleAssignmentCancelable}
+import monix.execution.exceptions.CompositeException
+import monix.reactive.{Consumer, OverflowStrategy}
+import monix.reactive.observers.Subscriber
+import monix.reactive.subjects.PublishSubject
+
+import scala.concurrent.Future
+import scala.util.control.NonFatal
+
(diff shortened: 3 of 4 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
```diff
@@ -125,6 +128,18 @@ trait Consumer[-In, +R] extends ((Observable[In]) => Task[R])
     new MapAsyncConsumer[In,R,R2](self, f)
 }
 
+trait PartitionedConsumer[A] {
```
alexandru: In Monix I've had this organization philosophy going against a flat namespace, to keep the root with the essentials for beginners to get started; not necessarily a good idea, but it's also in tune with what RxJava (and other projects that I like) have been doing.

If you think about `PartitionedConsumer`, thing is people are going to discover this type only by bumping into the parameter of `partitionParallel`. So when introducing this particular type, we have to think about where to place it and we could have 3 choices:

1. in the companion object of `Consumer`: in Monix plenty of types are prefixed by companion objects, especially ADTs
2. we could also start a new namespace called `monix.reactive.consumers` and move consumer implementations there - there's already precedent (e.g. `monix.reactive.observers` and `monix.reactive.observables`)
3. this is basically a `Function2` so we shouldn't introduce a new type

Thing is, it's nice to have `PartitionedConsumer.ignore`, however it's not going to be used in practice, because you want to partition stuff in order to do some sort of I/O where that partitioning is useful. Also, this type is prefixed with `Consumer` however it does not inherit from `Consumer`. So personally I would get rid of this type and just work with `Function2`.

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
```diff
@@ -369,6 +394,74 @@ object Consumer {
...
+                Continue
+            }
+
+          def onError(ex: Throwable) = errors.synchronized {
+            errors += ex
+            onComplete()
+          }
+
+          def onComplete() = {
+            if (!isUpstreamDone.getAndSet(true)) {
```
alexandru: This is fishy. So you need a `refID.cancel()` for each partition and that `refCount.cancel()` to happen in order for the stop logic to proceed. However, you don't need to do this synchronization here, because you can just move the `refCount.cancel()` in your `Subscriber.onComplete` implementation below.

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
```diff
@@ -369,6 +394,74 @@ object Consumer {
...
+    override def createSubscriber(cb: Callback[Unit], s: Scheduler): (Subscriber[A], AssignableCancelable) = {
+
+      val errors = scala.collection.mutable.ArrayBuffer.empty[Throwable]
+
+      val obs = PublishSubject[A]()
+      val pub = obs.whileBusyBuffer(overflowStrategy)
+      val isUpstreamDone = AtomicBoolean(false)
+
+      val refCount = RefCountCancelable { () =>
+        errors.synchronized {
```
alexandru: When this happens, it means that all subscribers have already received their `onComplete` or `onError`, hence you don't need to synchronize on `errors`, because no other `onError` events can happen after this is triggered.

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
```diff
@@ -369,6 +394,74 @@ object Consumer {
...
+        }
+      }
+
+      def subscriber(part: Int): Cancelable = {
+
+        def consumerSubscriber(refID: Cancelable) = new Subscriber[Seq[A]] {
+          implicit def scheduler: Scheduler = s
+
+          def onNext(elements: Seq[A]) =
+            consumer.consume(part, elements).coeval.value match {
```
alexandru: The `consumer.consume` invocation, along with `coeval.value` can throw an exception - for this particular kind of optimization, also wrap this into a `try/catch`, something like:

```scala
try consumer.consume(...).coeval.value match {
  ???
} catch {
  case NonFatal(ex) =>
    onError(ex)
    Stop
}
```
It doesn't happen usually, but this is library code and library code has to defend against users making mistakes.

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
```diff
@@ -369,6 +394,74 @@ object Consumer {
...
+        implicit def scheduler: Scheduler = s
+
+        def onNext(elem: A) = obs.onNext(elem)
+        def onError(ex: Throwable) = obs.onError(ex)
+        def onComplete() = obs.onComplete()
+      }
+
+      (0 until nrPartitions).foreach(subscriber)
+
+      out -> AssignableCancelable.dummy
```
alexandru: We've got a problem: in case all subscribers stop due to errors or something, they'll get unsubscribed for your `PublishSubject`, however the source does not stop and continues to stream events. It would be cool to stop the source as well.

This `AssignableCancelable` reference here can be used to cancel the source. You could return an actual `SingleAssignmentCancelable()` that you could then cancel in your logic for `RefCountCancelable` above.

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
```diff
@@ -369,6 +394,74 @@ object Consumer {
...
+          }
+        }
+
+        pub
+          .filter(x => partition(x) % nrPartitions == part)
+          .bufferIntrospective(s.executionModel.recommendedBatchSize)
+          .subscribe(consumerSubscriber(refCount.acquire()))
+      }
+
+      val out = new Subscriber[A] {
```
alexandru: Instead of building this `Subscriber` manually, there's a builder on `Subscriber` for wrapping an `Observer` instance and a `Subject` is an `Observer`.

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
Fristi: I agree on the matter that you won't use `PartitionedConsumer.ignore` much. I also agree that the interface is to top-level. Though by seeing a `Function2[Int, Seq[A], Task[Unit]]` I don't know what the `Int` is, you have to consult the docs for that matter. Maybe add the trait at the Consumer object? In that case it's not at the top level. For now I'll stick with `Function2`

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
Fristi: Good point. Another question arises; right now I have this:

```scala
try consumer(part, elements).coeval.value match {
  case Left(future) =>
    future.map(_ => Continue)
  case Right(()) =>
    Continue
} catch {
  case NonFatal(ex) =>
    onError(ex)
    Stop
}
```

The part `future.map(_ => Continue)` only deals with the happy path. Should we also deal with errors?

Adding a `.recover` case like this:

```scala
.recover {
  case NonFatal(ex) =>
    onError(ex)
    Stop
}
```

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
alexandru: It's true that `Int` is obscure, but you could add a type alias for it. Something like `type PartitionIndex = Int`. That should be enough for documentation purposes.

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
Fristi: I see the point. I tried to move the `refCount.cancel()` to the `val out = new Subscribe[A] { ... }`, but it didn't propagate the errors. Guess I need some synchronization there? I am abit puzzled about this. As you mentioned above we should propagate errors from calling `consumer(part, elements).coeval.value` to `onError` (the method which resides in impl in the `def consumerSubscriber`). Should that only write errors to the `val errors = ArrayBuffer.empty[Throwable]`? By doing that we are scattering things a bit around

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
alexandru: No, don't deal with errors exposed by that future. That subscriber is already wrapped in a `SafeSubscriber`, which is the one dealing with it.

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
alexandru: Try doing it in both `out.onError` and `out.onComplete` and yes, I guess you will duplicate the `errors += ex` logic by doing this.

It may distribute that logic around, however you then no longer need `isUpstreamDone.getAndSet(true)`, because the `RefCountCancelable` is the one protecting you from sync issues, expecting `n+1` cancel operations (where `n` is the number of your partitions for which you did an `acquire()`), using its own `Atomic`  internally. And getting rid of extra synchronization logic is more desirable IMO than not duplicating logic.

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
alexandru: I guess this comment of mine will no longer be useful if you add that error handling to `onComplete` and `onError`.

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
Fristi: Oke cool. Almost done with all your comments. What if a consumer throws a exception? It seems the errors are currently not propagated.

```scala
try consumer(part, elements).coeval.value match {
              case Left(future) =>
                future.map(_ => Continue)
              case Right(()) =>
                Continue
            } catch {
              case NonFatal(ex) =>
                errors += ex
                Stop
            }
```

alexandru: Hi @Fristi, sorry, forgot to return to this.

What do you mean errors don't get propagated? Describe what you expect to have happened.

So do describe how general error handling works:

1. errors being thrown are usually against the contract; we do not have an official way of dealing with errors, because we cannot catch all errors in all cases, or we cannot stream them etc... when exceptions get thrown, the behavior is considered undefined
2. that said we do have to make a reasonable effort to protect against user code, hence all the try/catching
3. in `PublishSubject` this means that when an error gets caught from downstream, the subscriber that generates the error gets unsubscribed and it won't receive any further notifications
4. in `subscribe()` we are also wrapping subscribers (given by users) with `SafeSubscriber` - this has `try/catch` logic in it that also catches errors happening in `onNext` and in case it happens it tries streaming it to `onError` while returning a `Stop` to the source (take a look if interested)

In other words, as long as you use normal `subscribe()` (and not `unsafeSubscribeFn`), your logic should be completely fine. And you can use `unsafeSubscribeFn` in your `def consumer`, but if you do, then you can do the error handling yourself. And in this case I think it is better to do it explicitly, then to rely on the difference between `subscribe()` and `unsafeSubscribeFn`, because the former is for users and has a behavior that is subject to change.

Fristi:
I think I fixed your comments. However now a question arises; what to should we do when a downstream error comes up (from the callback: `(Int, Seq[A]) => Task[Unit]`. Should we cancel upon receiving the next element per partition?

A simple test case shows that we'll miss 10% of the results if partition 0 raises an error.

```scala

  test("should deal with consumer errors") { implicit s =>

    val ex = DummyException("consumer error")
    val counter = AtomicLong(0l)
    val count = 20l

    val f = Observable.fromIterable(0l to count)
      .consumeWith(Consumer.partitionParallel[Long](10, OverflowStrategy.Unbounded, _.toInt, (part, seq) => if(part == 0) Task.raiseError(ex) else Task(counter.increment(seq.size)))).runAsync

    s.tick()

    assertEquals(counter.get, count - 2l)
    assertEquals(f.value, Some(Failure(CompositeException(Seq(ex)))))
  }
```

Fristi:
Would this still be relevant (or can we implement this with existing components)? If it's relevant I would like to have another look on it next week.

alexandru:
@Fristi It's still relevant, yes. Sorry for not mentoring this through.

I'm unsure about the correctness of the implementation. Please review it, merge with `master` and we can talk afterwards.

PR #298, opened 2017-01-09
Don't publish empty jars in root projects

Taig:
Disables the generation of a *.jar file for the empty monix aggregate/root project. Only for the JVM build though, because the JS version adds some additional files to the *.jar which are probably crucial.

This resolves some minor build issues on the Android platform.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,6 +16,13 @@ lazy val doNotPublishArtifact = Seq(
   publishArtifact in (Compile, packageBin) := false
 )
 
+lazy val noSources = Seq(
+  // disable automatic dependency on the Scala library
+  autoScalaLibrary := false,
+  publishArtifact in Compile := false,
+  sources in Compile := Seq.empty
+)
+
 lazy val warnUnusedImport = Seq(
   scalacOptions ++= {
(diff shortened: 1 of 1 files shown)


alexandru:
@Taig we need to be able to import this meta-package (and have it work cross-platform):

```scala
libraryDependencies += "io.monix" %% "monix" % "2.1.2"
```

This meta dependency that brings in all dependencies in core is enabled by `coreJVM` for the JVM. 

Will this still work?

alexandru:
Or to add details, I'm not 100% sure how Maven repos are supposed to work, but I'm sure at the very least we need this POM published and I'm not sure that we can do it without an accompanying JAR: https://repo1.maven.org/maven2/io/monix/monix_2.12/2.1.2/monix_2.12-2.1.2.pom

alexandru:
@Taig ok, let's try it.

alexandru:
Unfortunately we have issues with Coursier in https://github.com/alexarchambault/coursier/issues/429 :-(

Technically I don't think we are doing anything wrong because we are specifying `<packaging>pom</packaging>` instead of `<packaging>jar</packaging>`.

PR #300, opened 2017-01-17
Update to cats 0.9.0

guersam:
just a bump.


rstradling:
Looks like this updates the issue #302, that I just submitted. This may be nice for tracking purposes.  But do what you will and apologies for any issue.

PR #318, opened 2017-02-16
Adds bracket to Coeval and Task

reactormonk:
Not entirely sure how to unit-test this.

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
@@ -394,6 +394,20 @@ object Coeval {
   def fromTry[A](a: Try[A]): Coeval[A] =
     Attempt.fromTry(a)
 
+  /** Aquires a `Closeable` and closes it after the `Coeval` has been run.
+    */
+  def bracket[R <: java.io.Closeable](aquire: => R): Coeval[R] = defer {
+    val resource: R = aquire
+    now(resource).doOnFinish(_ => delay(resource.close))
+  }
+
+  /** Aquires a resource and closes it after the `Coeval` has been run.
+    */
+  def bracket[R](aquire: => R, close: R => Unit): Coeval[R] = defer {
+    val resource: R = aquire
+    now(resource).doOnFinish(_ => delay(close(resource)))
+  }
+
   /** Keeps calling `f` until it returns a `Right` result.
     *
     * Based on Phil Freeman's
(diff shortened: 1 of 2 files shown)


monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
```diff
@@ -394,6 +394,20 @@ object Coeval {
   def fromTry[A](a: Try[A]): Coeval[A] =
     Attempt.fromTry(a)
 
+  /** Aquires a `Closeable` and closes it after the `Coeval` has been run.
+    */
+  def bracket[R <: java.io.Closeable](aquire: => R): Coeval[R] = defer {
+    val resource: R = aquire
+    now(resource).doOnFinish(_ => delay(resource.close))
```
alexandru: Well, this isn't going to work, as `Coeval` is a simpler abstraction than the `Process` in FS2. `doOnFinish` is equivalent to this:

```scala
def doOnFinish[A](source: Coeval[A])(f: Option[Throwable] => Coeval[Unit]): Coeval[A] =
  source.materialize.flatMap {
    case Now(value) => f(None)
    case Error(ex) => f(Some(ex))
  }
```

In other words, this isn't end-of-the-world evaluation. For `Task` and `Coeval` you can't specify that in advance. What you can do instead is something like this:

```scala
def using[R <: java.io.Closable, A](resource: => R)(f: R => A): Coeval[A] =
  Coeval.defer {
    val ref = resource
    val close = Task.eval(ref.close())
    Coeval.eval(f(ref)).doOnFinish(_ => close)
  }
```

scottcarey: Then perhaps the following would be more generally useful:

````
    def bracket[R, A](resource: => R)(cleanup: R => ())(f: R => A): Coeval[A] =
  Coeval.defer {
    val ref = resource
    val close = Task.eval(cleanup(ref)))
    Coeval.eval(f(ref)).doOnFinish(_ => close)
  }
````

another flavor with a `cleanup: R => Task[Unit]`  would be useful too.

Lastly, in a signature that takes something from java.lang, `AutoClosable` should be preferred over `Closeable` as it is a supertype.

````
def using[R <: java.lang.AutoClosable, A](resource: => R)(f: R => A): Coeval[A] = bracket(resource)(r => r.close())(f)
````

Another question would be what to name variants that take a `R => Coeval[A]` rather than a `R => A` ?    Both are probably equally useful.

PR #331, opened 2017-03-13
WIP: Simplified Iterant Type

alexandru:
This new `Iterant` data type describes pull-based streams powered by `Task`.

This is an `Iterant` design offered as an alternative to PR #280, providing no plugable `F[_]` context, as a simplification.

Working with higher-kinded polymorphism powered by type classes in Scala is not pleasant or even feasible:

- interactions with variance and the compiler is pushed to its limits - for example I discovered a case where a simple change meant the compilation times increasing from 5 seconds to over one minute on my machine, not to mention dozens of cases where the compiler couldn't infer types, with really ugly fixes when pattern matching
- moving this implementation forward with useful stuff means designing new type-classes, because not all needed type-classes are available in Cats or Scalaz, and this is actually a hard exercise, see for example: https://github.com/typelevel/cats/pull/1552
- a higher-kinded design makes it harder to understand by users, especially due to lack of education and most of the time they'll want `Task`; I just want a simple pull-based solution, for "bringing your own booze" scenarios, maybe other libraries would be better, like FS2 or [iteratee.io](http://iteratee.io)

The upside is that I might actually ship it in this lifetime. Waiting for feedback...


alexandru:
I'm dropping this PR because it's not as cool as the other one :-)

notxcain:
> I'm dropping this PR because it's not as cool as the other one :-)

@alexandru which one?

alexandru:
@notxcain I moved the development in my personal repository and that branch is a little stagnant, waiting on the progress in `cats-effects` ... https://github.com/alexandru/monix/tree/wip-streams

I basically wanted to get rid of `monix-types` from Monix first. Which will happen in version 3.0.0.

PR #337, opened 2017-03-25
Refactor weighted random generators

dabd:
Remove Arbitrary instance for 'Either' (already provided by scalacheck).


alexandru:
Thanks @dabd

PR #351, opened 2017-04-27
Specify that Monix is now a Typelevel project

peterneyens:
After https://github.com/typelevel/general/issues/61


alexandru:
Oh, thanks, forgot to update :-)

PR #368, opened 2017-06-01
Add intersperse operator

### LINKED ISSUE #367: New operator `Observable.intersperse`
It could be useful an operator that emit a specific separator between every pair of elements, with an optional start and end element.

The signature could be like:
```scala
def intersperse(sep: A): Observable[A] =
def intersperse(start: A, sep: A, end: A): Observable[A]
```

Here is an example of usage:
```scala
val fooObservable: Observable[Foo] = ...
fooObservable.map(_.asJson.spaces2).intersperse("[", ",", "]")
```

And the output will be:
```json
[{"id": 1, "name": "Foo"}, {"id": 2, "name": "Bar"}, {"id": 3, "name": "Baz"}]
```

omainegra:
Proposal for the operator `Observable.intersperse`. It also includes unit tests for it. This PR resolves #367

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
@@ -35,7 +35,6 @@ private[reactive] final class IntersperseObservable[+A](
       implicit val scheduler = out.scheduler
 
       private[this] var firstTime = true
-      private[this] var isDone = false
       private[this] var downstreamAck = Continue : Future[Ack]
 
       override def onNext(elem: A): Future[Ack] = {
@@ -56,21 +55,12 @@ private[reactive] final class IntersperseObservable[+A](
         downstreamAck
       }
 
-      def onError(ex: Throwable) = {
-        if (!isDone){
-          isDone = true
-          out.onError(ex)
-        }
-      }
+      def onError(ex: Throwable) = out.onError(ex)
+
       def onComplete() = {
-        downstreamAck.syncFlatMap {
-          case Continue if end.isDefined => out.onNext(end.get)
-          case ack => ack
-        }.syncOnContinue {
-          if (!isDone){
-            isDone = true
-            out.onComplete()
-          }
+        downstreamAck.syncOnContinue {
+          if (end.nonEmpty) out.onNext(end.get)
+          out.onComplete()
         }
       }
     })
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
@@ -1,0 +1,56 @@
+package monix.reactive.internal.operators
+
+import monix.execution.Ack.Continue
+import monix.execution.atomic.AtomicBoolean
+import monix.execution.{Ack, Cancelable}
+import monix.reactive.Observable
+import monix.reactive.observers.Subscriber
+
+import scala.concurrent.Future
+
+/**
+  * Created by omainegra on 5/29/17.
+  */
+private[reactive] final class IntersperseObservable[+A](source: Observable[A],
+                                                        separator: A) extends Observable[A]{ self =>
+
+  override def unsafeSubscribeFn(out: Subscriber[A]): Cancelable = {
+    val firstTime = AtomicBoolean(true)
+    val isDone = AtomicBoolean(false)
+    var downstreamAck = Continue : Future[Ack]
+
+    val upstream = source.unsafeSubscribeFn(new Subscriber[A] {
+      implicit val scheduler = out.scheduler
+
+      override def onNext(elem: A): Future[Ack] = {
+        downstreamAck = if (firstTime.getAndSet(false)) {
+          out.onNext(elem)
+        }
+        else {
+          out.onNext(separator).flatMap {
+            case Continue => out.onNext(elem)
+            case ack => ack
+          }
+        }
+        downstreamAck
+      }
+
+      def onError(ex: Throwable) = {
+        if (!isDone.getAndSet(true)){
+          out.onError(ex)
+        }
+      }
+      def onComplete() = {
+        downstreamAck.syncOnContinue {
+          if (!isDone.getAndSet(true)){
+            out.onComplete()
+          }
+        }
+      }
+    })
+
+    Cancelable { () =>
+      upstream.cancel()
+    }
+  }
+}
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
@@ -18,45 +18,57 @@
 package monix.reactive.internal.operators
 
 import monix.execution.Ack.Continue
-import monix.execution.atomic.AtomicBoolean
 import monix.execution.{Ack, Cancelable}
 import monix.reactive.Observable
 import monix.reactive.observers.Subscriber
 
 import scala.concurrent.Future
 
-private[reactive] final class IntersperseObservable[+A](source: Observable[A],
-                                                        separator: A) extends Observable[A]{ self =>
+private[reactive] final class IntersperseObservable[+A](
+  source: Observable[A],
+  start: Option[A],
+  separator: A,
+  end: Option[A] ) extends Observable[A] { self =>
 
   override def unsafeSubscribeFn(out: Subscriber[A]): Cancelable = {
-    val firstTime = AtomicBoolean(true)
-    val isDone = AtomicBoolean(false)
-    var downstreamAck = Continue : Future[Ack]
-
     val upstream = source.unsafeSubscribeFn(new Subscriber[A] {
       implicit val scheduler = out.scheduler
 
+      private[this] var firstTime = true
+      private[this] var isDone = false
+      private[this] var downstreamAck = Continue : Future[Ack]
+
       override def onNext(elem: A): Future[Ack] = {
-        downstreamAck = if (firstTime.getAndSet(false)) {
-          out.onNext(elem)
-        }
-        else {
-          out.onNext(separator).flatMap {
+        downstreamAck = if (firstTime) {
+          firstTime = false
+          start.map(out.onNext).getOrElse(Continue).syncFlatMap {
             case Continue => out.onNext(elem)
             case ack => ack
           }
         }
+        else {
+          out.onNext(separator).syncFlatMap {
+            case Continue => out.onNext(elem)
+            case ack => ack
+          }
+        }
+
         downstreamAck
       }
 
       def onError(ex: Throwable) = {
-        if (!isDone.getAndSet(true)){
+        if (!isDone){
+          isDone = true
           out.onError(ex)
         }
       }
       def onComplete() = {
-        downstreamAck.syncOnContinue {
-          if (!isDone.getAndSet(true)){
+        downstreamAck.syncFlatMap {
+          case Continue if end.isDefined => out.onNext(end.get)
+          case ack => ack
+        }.syncOnContinue {
+          if (!isDone){
+            isDone = true
             out.onComplete()
           }
         }
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
@@ -1,3 +1,20 @@
+/*
+ * Copyright (c) 2014-2016 by its authors. Some rights reserved.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
```diff
@@ -0,0 +1,56 @@
...
+import monix.execution.Ack.Continue
+import monix.execution.atomic.AtomicBoolean
+import monix.execution.{Ack, Cancelable}
+import monix.reactive.Observable
+import monix.reactive.observers.Subscriber
+
+import scala.concurrent.Future
+
+/**
+  * Created by omainegra on 5/29/17.
```
alexandru: As a general rule, I abstained from adding names in source files - you can add yourself to the AUTHORS file.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
```diff
@@ -0,0 +1,56 @@
...
+  */
+private[reactive] final class IntersperseObservable[+A](source: Observable[A],
+                                                        separator: A) extends Observable[A]{ self =>
```
alexandru: I'm obsessed with indentation and I'd prefer for this to be consistent with the other files. Personally I don't like things dangling so far on the right, although I can understand the preference.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
```diff
@@ -0,0 +1,56 @@
...
+
+    val upstream = source.unsafeSubscribeFn(new Subscriber[A] {
+      implicit val scheduler = out.scheduler
+
+      override def onNext(elem: A): Future[Ack] = {
+        downstreamAck = if (firstTime.getAndSet(false)) {
+          out.onNext(elem)
+        }
+        else {
+          out.onNext(separator).flatMap {
```
alexandru: Doing a `flatMap` on a `Future` can be pretty inefficient since it forces an asynchronous boundary.

This is why Monix introduces a couple of special Future extension methods for avoiding async boundaries if the `Future` returned by `onNext` is already complete. Monix is efficient because it tries to collapse these async boundaries if the downstream subscriber returns a plain `Continue` or `Stop` that's immediately available.

Don't get me wrong, this code is correct, but it can be more efficient if you use `syncFlatMap` instead:

```scala
out.onNext(separator).syncFlatMap {
  ???
}
```

As a note, for user code this is actually dangerous because it is stack-unsafe (in loops), but we know what we are doing here and it's fine.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
```diff
@@ -0,0 +1,56 @@
...
+            case Continue => out.onNext(elem)
+            case ack => ack
+          }
+        }
+        downstreamAck
+      }
+
+      def onError(ex: Throwable) = {
+        if (!isDone.getAndSet(true)){
```
alexandru: Just as with `firstTime`, you're only modifying `isDone` in `onComplete` and `onError`. The calls to `onNext`, `onComplete` and `onError` are ordered (synchronized) already so this can be a `private[this] var`.

The only concurrency problem you can have is that the producer doesn't need to back-pressure for that final `Future[Ack]` returned by the last `onNext`, so depending what you do in those `flatMaps`, you can end up with a concurrency problem, but you don't need anything special here.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
```diff
@@ -0,0 +1,56 @@
...
+          out.onError(ex)
+        }
+      }
+      def onComplete() = {
+        downstreamAck.syncOnContinue {
```
alexandru: In this case you don't need to back-pressure on `downstreamAck`, because this is legal (as an optimization of the protocol:

```scala
out.onNext(e)
out.onComplete()
```

But this is NOT legal of course:

```scala
out.onNext(e)
out.onNext(e)
out.onComplete()
```

There are operators that keep that last `ack` around and use it in `onComplete`, but that's only because they have to do more stuff.

I suspect that if you have to stream a final `end`, then you do need to have that `ack` to back-pressure on, but you don't need that here.

monix-reactive/shared/src/main/scala/monix/reactive/observables/ObservableLike.scala
```diff
@@ -2183,6 +2183,14 @@ trait ObservableLike[+A, Self[+T] <: ObservableLike[T, Self]]
...
   def zipWithIndex: Self[(A, Long)] =
     self.liftByOperator(new ZipWithIndexOperator[A])
+
+  /** Creates a new observable from this observable that will emit a specific `separator`
+    * between every pair of elements.
+    *
+    * @param separator is the separator
+    */
+  def intersperse[B >: A](separator: B): Self[B] =
+    self.transform(self => new IntersperseObservable(self, separator))
```
alexandru: If we introduce this, we also need the version with the `start` and the `end` tokens and I think it can be handled by the same implementation - in that `Observable` source you can make them optional or something.

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/IntersperseSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+
+package monix.reactive.internal.operators
+
+import monix.execution.Ack.Continue
+import monix.reactive.subjects.PublishSubject
+import monix.reactive.{Observable, Observer}
+
+import scala.concurrent.duration._
+
+object IntersperseSuite extends BaseOperatorSuite {
```
alexandru: 👍 it's good that you introduced the test as well

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
```diff
@@ -0,0 +1,56 @@
...
+import scala.concurrent.Future
+
+/**
+  * Created by omainegra on 5/29/17.
+  */
+private[reactive] final class IntersperseObservable[+A](source: Observable[A],
+                                                        separator: A) extends Observable[A]{ self =>
+
+  override def unsafeSubscribeFn(out: Subscriber[A]): Cancelable = {
+    val firstTime = AtomicBoolean(true)
```
alexandru: You are modifying `firstTime` in `onNext` only, so you don't need to synchronize it, because the `onNext` calls are guaranteed to be ordered (the same guarantee you get with actors processing messages).

Atomics come with overhead and this can be a simple `private[this] var` in that `Subscriber` class.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
omainegra: Sorry about this, just IntelliJ template, fixing it

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
omainegra: I didn't know about this, good to know

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
```diff
@@ -1,48 +1,74 @@
...
           out.onError(ex)
         }
       }
       def onComplete() = {
-        downstreamAck.syncOnContinue {
-          if (!isDone.getAndSet(true)){
+        downstreamAck.syncFlatMap {
+          case Continue if end.isDefined => out.onNext(end.get)
+          case ack => ack
+        }.syncOnContinue {
```
alexandru: Looking good dude. One final touch ... there's no need for back-pressure with this `syncOnContinue` here, because you can do ...

```scala
out.onNext(end.get)
out.onComplete()
```

This optimisation (not waiting on the final `onNext` for sending `onComplete`) is good from a networking POV, but it's also why `Observable.tailRecM` can work well.

I would also prefer if you wouldn't wait on `downstreamAck` in case `end.isDefined == false`.
It's totally fine if we have multiple if branches or match statements.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
```diff
@@ -1,48 +1,74 @@
...
+
       override def onNext(elem: A): Future[Ack] = {
-        downstreamAck = if (firstTime.getAndSet(false)) {
-          out.onNext(elem)
+        downstreamAck = if (firstTime) {
+          firstTime = false
+          start.map(out.onNext).getOrElse(Continue).syncFlatMap {
+            case Continue => out.onNext(elem)
+            case ack => ack
+          }
```
alexandru: 👍 looking good

monix-reactive/shared/src/main/scala/monix/reactive/observables/ObservableLike.scala
```diff
@@ -2190,7 +2190,17 @@ trait ObservableLike[+A, Self[+T] <: ObservableLike[T, Self]]
...
+    self.transform(self => new IntersperseObservable(self, None, separator, None))
+
+  /** Creates a new observable from this observable that will emit the `start` element
+    * followed by the upstream elements paired with the `separator`, and lastly the `end` element.
+    *
+    * @param start is the first element emitted
+    * @param separator is the separator
+    * @param end the last element emitted
+    */
+  def intersperse[B >: A](start: B, separator: B, end: B): Self[B] =
```
alexandru: 👍

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/IntersperseObservable.scala
omainegra: Ok then :-). I'm still learning about all these things. You have been a great help

alexandru: It's totally fine @omainegra, no rush, I breathe this protocol and know the optimisation opportunities because I came up with it.

And your first try was correct and pretty good.

omainegra: I need assistance here please. If I don't do back-pressure for the onComplete, one unit test fail. 

```
[info] monix.reactive.internal.operators.IntersperseSuite
[info] - should emit exactly the requested elements
[info] - should work for synchronous observers
[info] - should work for asynchronous observers *** FAILED ***
[info]   received 24310 != expected 24530 (BaseOperatorSuite.scala:157)
[info]     minitest.api.Asserts$class.assertEquals(Asserts.scala:67)
[info]     monix.reactive.internal.operators.BaseOperatorSuite.assertEquals(BaseOperatorSuite.scala:30)
[info]     monix.reactive.internal.operators.BaseOperatorSuite$$anonfun$3.apply(BaseOperatorSuite.scala:157)
[info]     monix.reactive.internal.operators.BaseOperatorSuite$$anonfun$3.apply(BaseOperatorSuite.scala:125)
[info]     minitest.api.TestSpec$$anonfun$from$1.apply(TestSpec.scala:54)
[info]     minitest.api.TestSpec$$anonfun$from$1.apply(TestSpec.scala:52)
[info]     minitest.api.TestSpec.apply(TestSpec.scala:28)
```

alexandru: Sure, but please copy/paste your newest `onComplete`

alexandru: It does look like a contract problem, I don't dismiss the possibility of me being wrong :-)

omainegra: This is the new `onComplete`. In the else branch is commented the code with back-pressure. As it is, it's not working.

```scala
def onComplete() = {
  if (end.isDefined){
    downstreamAck.syncOnComplete {
      case Success(_) =>
        out.onNext(end.get)
        signalOnComplete()
      case _ =>
    }
  } else {
    signalOnComplete()
//          downstreamAck.syncOnComplete {
//            case Success(_) => signalOnComplete()
//            case _ =>
//          }
  }
}
```

alexandru: OK, I think I know what's going on. In `onNext` you can end up pushing 2 events instead of one, so if the subscriber is delaying on `sep`, you still have the `elem` queued up for sending it, so you can end up triggering `onNext(end)` and `onComplete` before that happens. See, I was wrong :-)

Btw, you need to check that you receive a `Continue` and don't need to do anything in case of `Stop` or `Failure(ex)`, so doing just one `syncOnContinue` there is totally fine.

```scala
ack.syncOnContinue {
  if (end.nonEmpty) out.onNext(end.get)
  out.onComplete
}
```
Or smth.

omainegra: Yes, It's as you say. One last question. Is it required the `isDone ` logic`?, because I just removed it and all test passed OK

alexandru: Yeah, you probably don't need it - `isDone` in `onComplete` is usually useful when you're doing stuff in `onNext` that triggers a completion and then you want to ensure that you're not going to be completing that stream a second time.

It might be OK to remove it.

PR #373, opened 2017-06-06
Fix: In `IntersperseObservable` do not emit the end element if the upstream completes without emitting

omainegra:


alexandru:
Ah, good catch.

PR #384, opened 2017-07-03
Add contramap method to Callback

Wogan:
Allows you to perform transformations on callback based APIs. Using contramap on a callback means you don't need to provide an ExecutionContext to map on a Future.

Example: 
```scala
def example: Future[Result] = {
    val promise: Promise[Result] = Promise[Result]()
    val callback: Callback[ApiResult] =
         Callback.fromPromise(promise).contramap { apiResult: ApiResult =>
             Result.fromApiResult(apiResult)
         }
    api.method("request data", callback)
    promise.future
}
```


alexandru:
Thanks @Wogan, will merge after I fix the build, which is currently failing for some reason.

PR #390, opened 2017-07-31
`flatMap` cancels all involved futures

### LINKED ISSUE #381: flatMap on CancelableFuture
Currently, the `flatMap` operation takes a function `A => Future[B]`. But there's no check that the resulting `Future[B]` might be cancelable itself. Hence, cancelling the returned `CancelableFuture[B]` might not cancel correctly. Here's my workaround:

```scala
  final implicit class CancelableFutureOps[A](future: CancelableFuture[A]) {
    def flatMapC[B](cont: A => CancelableFuture[B])(implicit ec: ExecutionContext): CancelableFuture[B] = {
      val conn = MultiAssignmentCancelable()
      val c1 = future.flatMap { a =>
        val c2 = cont(a)
        conn.orderedUpdate(c2, order = 2)
        c2
      }

      conn.orderedUpdate(c1, order = 1)
      CancelableFuture(c1, conn)
    }
  }
```

In principle, this could go into `CancelableFuture` itself, where there could be an `asInstanceOf` check.

larsrh:
Fixes #381

### DIFF
--- a/monix-execution/shared/src/main/scala/monix/execution/FutureUtils.scala
+++ b/monix-execution/shared/src/main/scala/monix/execution/FutureUtils.scala
@@ -19,6 +19,7 @@ package monix.execution
 
 import java.util.concurrent.TimeoutException
 
+import monix.execution.cancelables.MultiAssignmentCancelable
 import monix.execution.misc.NonFatal
 
 import scala.concurrent.duration._
@@ -115,10 +116,20 @@ object FutureUtils {
     *
     * Similar to `Future.transformWith` from Scala 2.12.
     */
-  def transformWith[A,B](source: Future[A], f: Try[A] => Future[B])(implicit ec: ExecutionContext): Future[B] = {
+  def transformWith[A,B](source: Future[A], f: Try[A] => Future[B])(implicit ec: ExecutionContext): CancelableFuture[B] = {
     val p = Promise[B]()
+
+    val cancelable = source match {
+      case c: Cancelable => MultiAssignmentCancelable(c)
+      case _ => MultiAssignmentCancelable()
+    }
+
     source.onComplete { result =>
       val fb = try f(result) catch { case NonFatal(t) => Future.failed(t) }
+      fb match {
+        case c: Cancelable => cancelable := c
+        case _ =>
+      }
       if (fb eq source)
         p.complete(result.asInstanceOf[Try[B]])
       else fb.value match {
@@ -126,7 +137,7 @@ object FutureUtils {
         case None => p.completeWith(fb)
       }
     }
-    p.future
+    CancelableFuture(p.future, cancelable)
   }
 
   /** Utility that transforms a `Future[Try[A]]` into a `Future[A]`,
--- a/monix-execution/shared/src/main/scala/monix/execution/CancelableFuture.scala
+++ b/monix-execution/shared/src/main/scala/monix/execution/CancelableFuture.scala
@@ -34,8 +34,6 @@ trait CancelableFuture[+A] extends Future[A] with Cancelable {
     CancelableFuture(super.transform(s, f), this)
   override def map[S](f: (A) => S)(implicit executor: ExecutionContext): CancelableFuture[S] =
     CancelableFuture(super.map(f), this)
-  override def flatMap[S](f: (A) => Future[S])(implicit executor: ExecutionContext): CancelableFuture[S] =
-    CancelableFuture(super.flatMap(f), this)
   override def filter(p: (A) => Boolean)(implicit executor: ExecutionContext): CancelableFuture[A] =
     CancelableFuture(super.filter(p), this)
   override def collect[S](pf: PartialFunction[A, S])(implicit executor: ExecutionContext): CancelableFuture[S] =
@@ -59,7 +57,14 @@ trait CancelableFuture[+A] extends Future[A] with Cancelable {
 
   // Needed for Scala 2.12 compatibility
   def transformWith[S](f: Try[A] => Future[S])(implicit executor: ExecutionContext): CancelableFuture[S] =
-    CancelableFuture(FutureUtils.transformWith(this, f), this)
+    FutureUtils.transformWith(this, f)
+
+  override def flatMap[S](f: (A) => Future[S])(implicit executor: ExecutionContext): CancelableFuture[S] =
+    // trick borrowed from Scala 2.12
+    transformWith {
+      case Success(s) => f(s)
+      case Failure(_) => this.asInstanceOf[CancelableFuture[S]]
+    }
 }
 
 object CancelableFuture {
--- a/monix-execution/shared/src/main/scala/monix/execution/CancelableFuture.scala
+++ b/monix-execution/shared/src/main/scala/monix/execution/CancelableFuture.scala
@@ -156,41 +156,5 @@ object CancelableFuture {
 
     override def cancel(): Unit =
       cancelable.cancel()
-
-    // Overriding methods for getting CancelableFuture in return
-    override def failed: CancelableFuture[Throwable] =
-      new Implementation(underlying.failed, cancelable)
-    override def transform[S](s: (A) => S, f: (Throwable) => Throwable)(implicit executor: ExecutionContext): CancelableFuture[S] =
-      new Implementation(underlying.transform(s, f), cancelable)
-    override def map[S](f: (A) => S)(implicit executor: ExecutionContext): CancelableFuture[S] =
-      new Implementation(underlying.map(f), cancelable)
-    override def flatMap[S](f: (A) => Future[S])(implicit executor: ExecutionContext): CancelableFuture[S] =
-      new Implementation(underlying.flatMap(f), cancelable)
-    override def filter(p: (A) => Boolean)(implicit executor: ExecutionContext): CancelableFuture[A] =
-      new Implementation(underlying.filter(p), cancelable)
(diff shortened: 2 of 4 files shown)


larsrh:
Error in Travis is because of binary incompatibility. I wasn't aware that returning a more specific type could cause this.

monix-execution/shared/src/main/scala/monix/execution/FutureUtils.scala
```diff
@@ -115,18 +116,28 @@ object FutureUtils {
     *
     * Similar to `Future.transformWith` from Scala 2.12.
     */
-  def transformWith[A,B](source: Future[A], f: Try[A] => Future[B])(implicit ec: ExecutionContext): Future[B] = {
+  def transformWith[A,B](source: Future[A], f: Try[A] => Future[B])(implicit ec: ExecutionContext): CancelableFuture[B] = {
```
ktoso: Chiming in since asked on twitter https://twitter.com/ktosopl/status/892221038242570240 

This is binary incompatible, MiMa is right.
Changing the type like this will cause `MethodNotFound`.

Simple way to test those things if you think MiMa may be mistaken:

```scala
$ ls *scala
it.scala    lol.scala   main.scala
ktoso @ 三日月/tmp
$ cat *scala
object f { def f(): A = null }

trait A
trait B extends A

object Main extends App {

 val a =   f.f()
}
ktoso @ 三日月/tmp
$ scalac *.scala
ktoso @ 三日月/tmp
$ java -cp /Users/ktoso/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.12.2.jar:. Main
ktoso @ 三日月/tmp
$ vim it.scala
ktoso @ 三日月/tmp
$ scalac it.scala
ktoso @ 三日月/tmp
$ cat it.scala
object f { def f(): B = null }

ktoso @ 三日月/tmp
$ java -cp /Users/ktoso/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.12.2.jar:. Main
Exception in thread "main" java.lang.NoSuchMethodError: f$.f()LA;
```

Hope this helps

larsrh: Yeah, it helps. I'll have to rework this change to make it binary compatible.

@alexandru WDYT?

alexandru:
@larsrh this change is for master and master has upgraded to depending on cats-core directly.

This means that binary compatibility is already broken heavily, because `monix-cats`, `monix-scalaz` and `monix-types` sub-projects are now gone, along with the instances defined for `Task`, `Coeval` and `Observable`.  And because we are doing this, this is also a good time to get rid of deprecated methods, which lingered in the codebase for preserving compatibility.

This is why we have a big list of incompatibilities [documented for 3.0.0](https://github.com/monix/monix/blob/master/project/MimaFilters.scala#L138).

So don't worry about it. If there's a time to break binary compatibility, it is now.
But Mima still serves a purpose here, because all breaking changes need to be documented.

monix-execution/shared/src/main/scala/monix/execution/FutureUtils.scala
```diff
@@ -115,18 +116,28 @@ object FutureUtils {
...
         p.complete(result.asInstanceOf[Try[B]])
       else fb.value match {
         case Some(value) => p.complete(value)
         case None => p.completeWith(fb)
       }
     }
-    p.future
+
+    source match {
+      case c: Cancelable => cancelable.orderedUpdate(c, order = 1)
```
alexandru: This is correct, but you already have `source` when initializing your `val cancelable` so it might be a better idea to initialize that as `MultiAssignmentCancelable(source)` (thus `source` being an initial underlying reference), because we avoid doing an `Atomic.compareAndSet` that way.

And so instead of the second `orderedUpdate(order = 2)`, you can just use `:=`, because `orderedUpdate` also implies some extra logic.

monix-execution/shared/src/main/scala/monix/execution/FutureUtils.scala
```diff
@@ -161,6 +172,10 @@ object FutureUtils {
     p.future
   }
 
+  /** Creates a future that never completes. */
+  def never[A]: Future[A] =
+    Promise[A].future
```
alexandru: 👍

ktoso: Since I got notified since I chimed in about the MiMa thing thought I'll share some insight here
 too, feel free to tell me to go away if you'd prefer that :-)

I would say 👎, since this can create memory leaks in the form of callbacks registered on this future, which will never trigger, however still will be registered and kept in memory. 

Explanations: http://viktorklang.com/blog/Futures-in-Scala-2.12-part-6.html 
Scala 2.12 solution: https://github.com/scala/scala/blob/2.12.x/src/library/scala/concurrent/Future.scala#L561

Sneaky little edge case, isn't it

monix-execution/shared/src/main/scala/monix/execution/FutureUtils.scala
alexandru: Thanks @ktoso for your input. @larsrh I've written a comment below.

ktoso: Sounds like good timing if you can break now 👍

monix-execution/shared/src/main/scala/monix/execution/FutureUtils.scala
larsrh: I only use this for testing. I could move this to the test suite, such that nobody will accidentally use it.

larsrh: The other option would be copy-pasting this from Scala into Monix, but I'm not sure whether that's a good idea.

ktoso: Yeah if it's just in tests then ok I guess. Your call :)

alexandru: @ktoso damn,

We also have something like this:

```scala
  private object Never extends CancelableFuture[Nothing] {
    def onComplete[U](f: (Try[Nothing]) => U)
      (implicit executor: ExecutionContext): Unit = ()

    val isCompleted = false
    val value = None

    @scala.throws[Exception](classOf[Exception])
    def result(atMost: Duration)(implicit permit: CanAwait): Nothing =
      throw new TimeoutException("This CancelableFuture will never finish!")

    @scala.throws[InterruptedException](classOf[InterruptedException])
    @scala.throws[TimeoutException](classOf[TimeoutException])
    def ready(atMost: Duration)(implicit permit: CanAwait): this.type =
      throw new TimeoutException("This CancelableFuture will never finish!")

    def cancel(): Unit = ()
  }
```

What do you think?

alexandru: OH, I now see that it is basically `never` from Scala 2.12. Nice.

larsrh: @alexandru I'll use that, then.

monix-execution/shared/src/main/scala/monix/execution/FutureUtils.scala
larsrh: Will update accordingly.

monix-execution/shared/src/main/scala/monix/execution/FutureUtils.scala
alexandru: But then `never` from Scala 2.12 overrides everything else, e.g. `flatMap`.

@ktoso If we aren't overriding the other methods like`flatMap`, I think that represents a potential memory leak as well.

@larsrh we might need to do a Scala 2.11 specific object (in a `scala_2.11` directory) and just rely on Scala's `never` for 2.12.

alexandru: @ktoso sorry for the noise, just wanted to mention that your input is greatly appreciated. Thanks.

larsrh:
I've addressed all feedback & Travis is green.

monix-execution/shared/src/main/scala/monix/execution/FutureUtils.scala
ktoso: No problem, cheers!

larsrh:
@alexandru Are you planning to release milestone builds of 3.0.0?

alexandru:
@larsrh yes, snapshots get published automatically whenever I merge master into `snapshot`.

I started a new build, when finished you should have a new version published if you want to play: https://travis-ci.org/monix/monix/builds/259733983

larsrh:
Your publish script might be failing:

    java.lang.IllegalStateException: Ivy file not found in cache for io.monix#monix-eval_2.11;3.0.0-27c1508!

(see [build log](https://travis-ci.org/monix/monix/jobs/259733986))

alexandru:
@larsrh unfortunately I've been having issues due to tooling, like Coursier, not recognizing empty projects (i.e. projects where the POM itself is the artifact,with no JAR file) and the root monix project is one.

Cannot reproduce it locally, but I've just pushed an attempted fix for it, a build is in progress: https://travis-ci.org/monix/monix/builds/259961033

alexandru:
@larsrh the latest build succeeded after those fixes, you should see `3.0.0-22bf9c6` on Maven Central.

larsrh:
Indeed, thanks!

PR #395, opened 2017-08-05
Monad[CancelableFuture]

### LINKED ISSUE #394: Monad[CancelableFuture]
Should this go into `object CancelableFuture`?

larsrh:
Fixes #394

### DIFF
--- a/monix-execution/shared/src/main/scala/monix/execution/CancelableFuture.scala
+++ b/monix-execution/shared/src/main/scala/monix/execution/CancelableFuture.scala
@@ -17,10 +17,14 @@
 
 package monix.execution
 
+import cats.{CoflatMap, Eval, Monad, MonadError, StackSafeMonad}
+
 import scala.concurrent._
 import scala.concurrent.duration.Duration
 import scala.reflect.ClassTag
 import scala.util.{Failure, Success, Try}
+import scala.util.control.NonFatal
+
 
 /** Represents an asynchronous computation that can be canceled
   * as long as it isn't complete.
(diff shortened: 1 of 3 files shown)


monix-execution/shared/src/main/scala/monix/execution/CancelableFuture.scala
```diff
@@ -162,4 +164,10 @@ object CancelableFuture {
     override def cancel(): Unit =
       cancelable.cancel()
   }
+
+  implicit def cancelableFutureMonad(implicit ec: ExecutionContext): Monad[CancelableFuture] = new Monad[CancelableFuture] with StackSafeMonad[CancelableFuture] {
```
alexandru: I'm trying to have consistent naming and just made some changes, so according to PR #397 we'll now have:

- `Task.catsInstances` (implementing `Async[Task]`)
- `Task.catsEffectInstances` (implementing `Effect[Task]`, different from above because this one requires a `Scheduler`)
- `Coeval.catsInstances`
- `Observable.catsInstances`
- `Iterant.catsInstances`

It would be good if this where named `catsInstances` too.

Also, we should probably implement `MonadError[CancelableFuture, Throwable]` as well.

larsrh: On it.

larsrh: Done.

alexandru:
@larsrh have you committed anything new? I'm still seeing your original commit only.

larsrh:
Sorry, forgot the `-a` in `git commit --amend`. Should be fixed now.

PR #414, opened 2017-08-14
Add flip convenience method to AtomicBoolean.

Wogan:
I find `ab.flip(true)` reads better than `ab.compareAndSet(expect = false, update = true)`.


alexandru:
Thanks @Wogan, but please add a test.

Wogan:
Hopefully these tests are in line with what you expect.

Issue #246, opened 2016-10-15
Support for dynamically scoped variables


missingfaktor:
This is a feature request for support for dynamically scoped variables in Monix.

The standard library implementation of dynamically scoped variables – [`DynamicVariable`](http://www.scala-lang.org/api/2.11.8/#scala.util.DynamicVariable) – uses `ThreadLocal` variables. This is not safe to use in code using `Task`s, since the computations along the chain could be dispatched on different threads. It would be useful to have a `DynamicVariable` implementation that works with Monix `Task`s. 

Here is how a typical use may look like:

``` scala
// Assume all of the following three live in three different classes.

val currentRequestContext = new MonixDynamicVariable[AwesomeAppRequestContext]

...

def foo(request: Request) = {
  val requestId = generateRequestIdForEndToEndTracking(request)
  val requestContext = buildRequestContext(request, requestId)
  currentRequestContext.withValue(requestContext) {
    bar(request.someParam)
  }
}

...

def bar(param: Param) = {
  doStuff()
  someLogger.debug("Beginning stuff", requestId = currentRequestContext.get.requestId)
  returnSomeOtherStuff()
}

```

In absence of such a construct, users are burdened with having to pass this parameter explicitly down the stack. The situation could be improved slightly by passing it as an `implicit` argument or including a `Reader` in your effect stack, but both approaches bring in their own tediousness. `DynamicVariable` is potentially a cleaner and more lightweight approach than the two.

Here is an implementation for Twitter/Finagle futures – [Local](https://github.com/twitter/util/blob/develop/util-core/src/main/scala/com/twitter/util/Local.scala) – for a reference.

fwbrasil:
Locals would make monix very appealing, especially because it already supports cancellation if I understood it correctly. To be honest, it's hard to imagine how I'd implement more complex systems without these features.

Note that the local works with Twitter Futures because `Local.Context` is propagated by the future implementation ([example](https://github.com/twitter/util/blob/develop/util-core/src/main/scala/com/twitter/util/Promise.scala)).

alexandru:
Hi guys, 

`Local` sounds cool and I'd like us to have it in Monix. But off the top of my head, here are some concerns:

[...]

PR #420, opened 2017-09-01
[WIP] Tracing Scheduler

leandrob13:
An attempt to solve #246 by adding a implementations of `LocalContext`, based on finagle's implementation, `TracingScheduler` and `TaskTracedRunLoop` for propagation of locals. Using the CorrelationId Tracing context we can do the following:

Example in [tests](https://github.com/leandrob13/monix/blob/feature/tracingScheduler/monix-eval/jvm/src/test/scala/monix/execution/misc/CorrelationIdSuite.scala)
```scala
    def taskFlatMap(implicit sc: Scheduler): Task[(Int, Option[CorrelationId])] = {
      for {
        a <- Task.fromFuture(Future(1))
        b <- Task.fromFuture(Future(1))
        c <- Task.fromFuture(Future(1))
        d <- Task.pure(CorrelationId.current)
      } yield (a + b + c, d)
    }

    val t1 = CorrelationId("1111").asCurrent {
      // Works with normal AsyncScheduler and runAsyncTraced
      taskFlatMap.runAsyncTraced
    }
    val (_, cid1) = Await.result(t1, 10.seconds)
    assert(cid1.contains(CorrelationId("1111")))

    val t2 = CorrelationId("1111").asCurrent {
      import monix.execution.schedulers.TracingScheduler.Implicits.traced
      // Works with TracingScheduler and runAsync
      taskFlatMap.runAsync
    }
    val (_, cid2) = Await.result(t2, 10.seconds)
    assert(cid2.contains(CorrelationId("1111")))
```

Currently a WIP.


alexandru:
@leandrob13 I don't think this will work well.

This is indeed a problem of transporting those values over asynchronous boundaries, but you don't have a guarantee that your tracing scheduler will be charge.

```scala
// Stupid sample to illustrate a point
val tick: Task[Unit] = Task.unsafeCreate { (ctx, cb) => 
  import java.util.concurrent.Executors
  val ec = Executors.newSingleThreadExecutor() 

  ec.execute(new Runnable () {
    def run() = {
      try cb(Success(())) finally ec.shutdown()
    }
  })
}
```

In this example `doStuff` basically does its own thing and you have no way of knowing that you have an async boundary in it.

This is why, if you're going to implement something that works with `Task`, then the restore of those values has to happen when that `cb(Success(()))` call is made above, which in the implementation ends up executing the continuation of the run-loop. Such that this can work ...

```scala
tick.flatMap(_ => Task.pure(TracingContext.value))
```

In case the problem isn't clear, in this `tick.flatMap(...)` call, the `TracingScheduler` that you described is not involved at all. Yes, we receive its reference in `Task.unsafeCreate` in that `ctx` that's injected, but we are not using it.

Also pay attention to laziness. A construct like this can't work because with `Task` nothing gets executed yet:

```scala
TracingContext.withValue(Map("xID" -> "1234")) {
  Task(println(s"Log this map ${TracingContext.value}"))
}
```

alexandru:
@leandrob13 btw, nice to see you working on this 🙂

leandrob13:
@alexandru so defining a new scheduler to catch a context is not the way to go for `Task`? Or is is the way I am doing it with `DynamicVariable`? I understand your point and it seems I will have to focus my efforts on `TaskRunLoop` where I am guessing it might get messy. 

I am thinking of keeping the scheduler to see how it plays along with scala Futures and do some changes just to experiment. I am thinking of two other approaches: 
-Try it with java's `ThreadLocal`.
-Implementing a `Local` like in finagle.

Thanks for the pointers! And expect more questions!

leandrob13:
@alexandru I made some progress with the LocalContext implementation. I used the tick example you gave me and it seems to be propagating it using only the TracingScheduler. I couldn't make it work with the new `TaskTracedRunLoop`, considering I was passing the local until the las part of the `onFinish` execution.

The tests are [here](https://github.com/leandrob13/monix/blob/feature/tracingScheduler/monix-eval/shared/src/test/scala/monix/execution/misc/CorrelationIdSuite.scala)

and the implementation of [TaskTracedRunLoop](https://github.com/leandrob13/monix/blob/feature/tracingScheduler/monix-eval/shared/src/main/scala/monix/eval/internal/TaskTracedRunLoop.scala)

leandrob13:
@alexandru I have some working [tests](https://github.com/leandrob13/monix/blob/feature/tracingScheduler/monix-eval/shared/src/test/scala/monix/execution/misc/CorrelationIdSuite.scala) that you can review now. It works with your tick example. I had a bug in the `LocalContext` implementation, it is fixed now.

I think I am getting closer, just need to check more scenarios to test.

leandrob13:
Closing it, I opened a cleaner PR in #429.

PR #429, opened 2017-09-11
Add LocalContext support for propagation of locals

leandrob13:
An attempt to solve #246 by adding an implementations of `Local` (based on finagle's implementation), `TracingScheduler` (for propagation through async boundaries) and `TaskTracedRunLoop` (for propagation of locals in trampolined execution). Using the CorrelationId tracing context we can do the following:

Example in [tests](https://github.com/leandrob13/monix/blob/feature/LocalContext/monix-eval/jvm/src/test/scala/monix/execution/tracing/CorrelationIdSuite.scala)
```scala
    def taskFlatMap(implicit sc: Scheduler): Task[(Int, Option[CorrelationId])] = {
      for {
        a <- Task.fromFuture(Future(1))
        b <- Task.fromFuture(Future(1))
        c <- Task.fromFuture(Future(1))
        d <- Task.pure(CorrelationId.current)
      } yield (a + b + c, d)
    }

    val t1 = CorrelationId("1111").asCurrent {
      // Works with normal AsyncScheduler and runAsyncTraced
      taskFlatMap.runAsyncTraced
    }
    val (_, cid1) = Await.result(t1, 10.seconds)
    assert(cid1.contains(CorrelationId("1111")))

    val t2 = CorrelationId("1111").asCurrent {
      import monix.execution.schedulers.TracingScheduler.Implicits.traced
      // Works with TracingScheduler and runAsync
      taskFlatMap.runAsync
    }
    val (_, cid2) = Await.result(t2, 10.seconds)
    assert(cid2.contains(CorrelationId("1111")))
```


alexandru:
@leandrob13 just an FYI, I'm very busy this week at work and with the presentation for Scala World, I'll review this next week and if I'm silent, please ping me :-)

Thanks,

leandrob13:
@alexandru No problem and thanks!

monix-eval/shared/src/main/scala/monix/eval/internal/TaskTracedRunLoop.scala
```diff
@@ -0,0 +1,475 @@
...
+      private[this] val runLoopIndex = context.frameRef
+
+      def prepare(bindCurrent: Bind, bindRest: CallStack): Unit = {
+        canCall = true
+        this.bFirst = bindCurrent
+        this.bRest = bindRest
+      }
+
+      def onSuccess(value: Any): Unit =
+        Local.withContext(local) {
```
alexandru: Why use `Local.withContext` here instead of `Local.setContext`? 

Can you give me an example of why restoring that context at the end of the block matters?

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -0,0 +1,104 @@
...
+    * Clear the Local state.
+    */
+  def clearContext(): Unit =
+    localContext.set(null)
+
+  /**
+    * Execute a block of code using the specified state of Local
+    * Context and restore the current state when complete
+    */
+  def withContext[T](ctx: Context)(f: => T): T = {
```
alexandru: I care about performance a lot and having this be a function call is probably going to mess up the numbers - just a feeling, I haven't tested anything.

One possibility is for us to not work with it in the implementation and just leave it in the API for the users.
Another possibility would be for this to be implemented by a macro, like we have for `Atomic.transform*` functions.

Well, validation that transforming it into a macro improves performance would be nice. As an example, here is:

1. [syncOnContinue macro](https://github.com/monix/monix/blob/master/monix-execution/shared/src/main/scala/monix/execution/Macros.scala#L50)
2. [signature](https://github.com/monix/monix/blob/master/monix-execution/shared/src/main/scala/monix/execution/Ack.scala#L93)

That one is being defined as an extension method, which I think it is cool for macros.

In other words, I am not sure if turning this `withContext` into a macro helps performance, but if it does, then we definitely need to do it.

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -0,0 +1,104 @@
...
+    }
+    localContext.set(newCtx)
+  }
+
+  private def get(id: String): Option[_ <: LocalContext] = {
+    Option(localContext.get).flatMap(_.get(id).flatten)
+  }
+
+  private def clear(id: String): Unit =
+    Option(localContext.get).foreach { m =>
```
alexandru: I know you like FP code and doing stuff like this, but it adds overhead and in the context of a library the code does not have to be nice. So don't wrap `localContext.get` in `Option` anymore, just pattern match on it, e.g. 

```scala
localContext.get match {
  case null => None
  case m => localContext.set(m - id)
}
```

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -0,0 +1,104 @@
...
+    scala.collection.immutable.Map(k -> v)
+
+  private[this] val localContext = ThreadLocal[Context]()
+
+  private def register(): String = synchronized {
+    UUID.randomUUID().toString
+  }
+
+  /**
+    * Return the state of the current Local state.
```
alexandru: So `Local` needs more ScalaDoc. I suggest also coming up with some dumb samples for these functions.

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private def clear(id: String): Unit =
+    Option(localContext.get).foreach { m =>
+      localContext.set(m - id)
+    }
+
+}
+
+final class Local[T <: Local.LocalContext] {
```
alexandru: Definitely needs ScalaDoc, with samples.

monix-execution/jvm/src/main/scala/monix/execution/schedulers/TracingScheduler.scala
```diff
@@ -0,0 +1,89 @@
...
+import java.util.concurrent.{ScheduledExecutorService, TimeUnit}
+import monix.execution.misc.Local
+import monix.execution.{Cancelable, Scheduler, UncaughtExceptionReporter, ExecutionModel => ExecModel}
+import scala.concurrent.ExecutionContext
+
+final class TracingScheduler private (
+                                       scheduler: ScheduledExecutorService,
+                                       ec: ExecutionContext,
+                                       r: UncaughtExceptionReporter,
+                                       val executionModel: ExecModel)
```
alexandru: Please fix formatting. You can configure IntelliJ IDEA to not align parameters like this.

monix-execution/jvm/src/main/scala/monix/execution/schedulers/TracingScheduler.scala
```diff
@@ -0,0 +1,89 @@
+package monix.execution.schedulers
+
+import java.util.concurrent.{ScheduledExecutorService, TimeUnit}
+import monix.execution.misc.Local
+import monix.execution.{Cancelable, Scheduler, UncaughtExceptionReporter, ExecutionModel => ExecModel}
+import scala.concurrent.ExecutionContext
+
+final class TracingScheduler private (
+                                       scheduler: ScheduledExecutorService,
```
alexandru: I think it would be a better idea to wrap our own `SchedulerService`, otherwise you're restricting instances to just wrap `ScheduledExecutorService` and for example that excludes `ForkJoinPool` (AFAIK, not absolutely sure).

monix-execution/jvm/src/main/scala/monix/execution/schedulers/TracingScheduler.scala
```diff
@@ -0,0 +1,89 @@
+package monix.execution.schedulers
+
+import java.util.concurrent.{ScheduledExecutorService, TimeUnit}
+import monix.execution.misc.Local
+import monix.execution.{Cancelable, Scheduler, UncaughtExceptionReporter, ExecutionModel => ExecModel}
+import scala.concurrent.ExecutionContext
+
+final class TracingScheduler private (
```
alexandru: This needs ScalaDoc, with samples :-)

monix-execution/jvm/src/main/scala/monix/execution/schedulers/TracingScheduler.scala
```diff
@@ -0,0 +1,89 @@
...
+    r.reportFailure(t)
+
+  override def withExecutionModel(em: ExecModel): TracingScheduler =
+    new TracingScheduler(scheduler, ec, r, em)
+}
+
+object TracingScheduler {
+
+  object Implicits {
+    implicit lazy val traced: Scheduler =
```
alexandru: Don't put implicits here. They should be in `Scheduler.Implicits`. So besides `global`, we could have this `traced` as well.

monix-execution/jvm/src/main/scala/monix/execution/tracing/CorrelationId.scala
```diff
@@ -0,0 +1,16 @@
+package monix.execution.tracing
+
+import monix.execution.misc.{TracingContext, TracingContextCompanion}
+
+/**
+  * A tracing context that allows to propagate a correlation id
+  * through an execution.
+  * @param id the correlation id.
+  */
+case class CorrelationId(id: String) extends TracingContext {
```
alexandru: Please make all `case class` definitions that you do `final`, because they are not and there's no reason for why they shouldn't be.

monix-execution/jvm/src/main/scala/monix/execution/tracing/CorrelationId.scala
```diff
@@ -0,0 +1,16 @@
+package monix.execution.tracing
+
+import monix.execution.misc.{TracingContext, TracingContextCompanion}
+
+/**
+  * A tracing context that allows to propagate a correlation id
+  * through an execution.
```
alexandru: From the ScalaDoc I don't understand the purpose of this.

If you can give a sample, please do so, or you can mention something like "*Used in X*" or something like that.

monix-execution/jvm/src/main/scala/monix/execution/schedulers/TracingScheduler.scala
```diff
@@ -0,0 +1,89 @@
...
+  extends ReferenceScheduler with BatchingScheduler { self =>
+
+  override def executeAsync(r: Runnable): Unit =
+    executeWithTrace(r)
+
+  /** Executes the given task with tracing.
+    *
+    * @param r is the callback to be executed
+    */
+  def executeWithTrace(r: Runnable): Unit = {
```
alexandru: If `executeAsync` is an alias of `executeWithTrace`, then why does `executeWithTrace` have to exist at all?

monix-execution/jvm/src/main/scala/monix/execution/schedulers/TracingScheduler.scala
```diff
@@ -0,0 +1,89 @@
...
+    * @return a `Cancelable` that can be used to cancel the created task
+    *         before execution.
+    */
+  override def scheduleOnce(initialDelay: Long, unit: TimeUnit, r: Runnable): Cancelable = {
+    if (initialDelay <= 0) {
+      executeWithTrace(r)
+      Cancelable.empty
+    } else {
+      val deferred = new Runnable {
+        override def run(): Unit = executeWithTrace(r)
```
alexandru: This anonymous `Runnable` instance isn't OK, but I just realized there's something I need to fix as well, so await further instructions on this one :-)

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  trait LocalContext
+
+  private[this] def context(k: String, v: Option[LocalContext]): Context =
+    scala.collection.immutable.Map(k -> v)
+
+  private[this] val localContext = ThreadLocal[Context]()
+
+  private def register(): String = synchronized {
+    UUID.randomUUID().toString
```
alexandru: This seems a little wasteful, with the synchronization and the `toString`. Do we really need to use strings for IDs btw? Strings are awful.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -314,6 +314,21 @@ sealed abstract class Task[+A] extends Serializable { self =>
...
+    *
+    * @param s is an injected [[monix.execution.Scheduler Scheduler]]
+    *        that gets used whenever asynchronous boundaries are needed
+    *        when evaluating the task
+    *
+    * @return a [[monix.execution.CancelableFuture CancelableFuture]]
+    *         that can be used to extract the result or to cancel
+    *         a running task.
+    */
+  def runAsyncTraced(implicit s: Scheduler): CancelableFuture[A] = {
```
alexandru: I like this `runAsyncTraced`, however we need to ensure that it works well, as I have doubts.

The problem with the run-loop is that within `Async` tasks users can restart that run-loop in however way they want. And we do that in our own implementation. For example:

```scala
Task.mapBoth(task1, task2) { (a, b) => a + b }
```

Within `mapBoth` we are starting `task1` and `task2` with the normal `runAsync`. Now indeed, according to your implementation, the local context should be restored at some point, probably in that mapping logic.

But will users expect for `task1` and `task2` to not receive that context on execution as well?
The other possibility would have been to introduce an extra option in `Task.Options`.

So what do you think?

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
alexandru: Also, we probably need a `Local` in `monix.eval` that is purely function (i.e. that exposes methods returning `cats.Eval` or `Coeval` ;-)

We'll think about that as it's extra, don't worry about it for now.

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
leandrob13: @alexandru I will change it. I will follow the same implementation as finagle where they use class instances for that. Do you agree?

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
leandrob13: @alexandru Understood. The samples on the tests or in the comments?

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
leandrob13: @alexandru I can leave it as it is and do the macro alternative so we can compare in the future.

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
leandrob13: @alexandru guilty! I will comply.

monix-execution/jvm/src/main/scala/monix/execution/schedulers/TracingScheduler.scala
leandrob13: @alexandru you are right, I did that at the beginning while I was trying to understand the schedulers and how they are used in TaskRunLoop.

monix-execution/jvm/src/main/scala/monix/execution/schedulers/TracingScheduler.scala
leandrob13: @alexandru understood!

monix-eval/shared/src/main/scala/monix/eval/Task.scala
leandrob13: @alexandru you presenting these cases is useful so I can start getting more tests know if this works. I can start testing this case and evaluate.

My first concern about the proposal was about this type of cases. Is it something that we want to support for every constructor of a Task or leave a few ones to construct traced tasks? An additional option in `Task.Options` sounds reasonable if we want to extend this to all the Task API. I wanted to do the TaskTracedRunLoop to not force the tracing as default but to keep it optional, I assume that is still the intention.

monix-eval/shared/src/main/scala/monix/eval/internal/TaskTracedRunLoop.scala
leandrob13: @alexandru testing the CorrelationIdSuite after making this change:

```scala
      def onSuccess(value: Any): Unit = {
        Local.setContext(local)
          // This is where finally the local gets propagated when executing
          if (canCall) {
            canCall = false
            loop(Now(value), context.executionModel, callback, this, bFirst, bRest, runLoopIndex())
          }
        }
```

Throws errors on assertions because in the execution a thread might get reused so its context may still exist and tests like this might fail:

```scala
test("should NOT get CorrelarionId with flatmapped Task with async boundary") {
    // Does not work without TracingScheduler
    // Meaning that when mapping an executed Task, the future will always
    // need the TracingScheduler to propagate the CorrelationId

    val t1 = CorrelationId("0000").asCurrent {
      taskFlatMap.runAsyncTraced.map {
        case (x, v) =>
          (x, CorrelationId.current)
      }
    }
    val (_, cid1) = Await.result(t1, 10.seconds)
    assert(cid1.isEmpty)

    val t2 = CorrelationId("0000").asCurrent {
      taskFlatMap.runAsync.map {
        case (x, v) =>
          (x, CorrelationId.current)
      }
    }
    val (_, cid2) = Await.result(t2, 10.seconds)
    assert(cid2.isEmpty)
  }
```

Because the Local value is still defined in that thread. I reference this comment [Source](http://stevenskelton.ca/threadlocal-variables-scala-futures/) :
```
A small note about garbage collection. As long as a thread exists, it will maintain references
to its ThreadLocal variables. If the thread-pool does not recycle threads and a thread goes
back into the pool without releasing its TLS then those objects will not be freed
```

At least for tracing contexts, we need to use only the current local and restore the previous state so we don't have leaked contexts.

alexandru: I see, OK.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
alexandru: The only thing I care about is the performance impact for when the option is turned off.

When the option is turned off, I assume that's going to simply involve one or two `if` expressions being rejected (for reading and restoring that context respectively on async boundaries).

So if you can do a benchmark to see the impact of such if statements, that would be great.
We have a `benchmarks` sub-project, add a file in `benchmarks/next`, see the other files there because it's not hard to do it.

I like your approach, but I fear gotchas, plus there's always the issue of code duplication, which isn't bad, but then you have to retest all of those branches and it's difficult. So lets do an informed decision here - if it has a performance impact, we'll do it like you did it here, otherwise we simply modify the current run-loop implementation with the needed if statements.

monix-execution/jvm/src/main/scala/monix/execution/misc/Local.scala
andyscott: This PR was sent my way 😄. So I'll weigh in on this:

If you covert this method to a macro, the only difference you'd be able to make is that you can guarantee that it's inlined at call site. I doubt that it's going to make a difference as the method is only called a few times internally. Granted they may be on hot paths-- but the benefit of inlining might be very small, and the JVM might actually do the inlining already.

If we have a way to do benchmarking, it would be trivial to inline this by hand in the few spots it's used.

leandrob13: @alexandru yeah, I brought a friend that knows more about macros than I do!

alexandru: 😃 happy to hear your opinion @andyscott 

Right, so I agree that these are just hunches and before making changes for performance reasons, we should test that it's actually good for performance, because I discovered multiple times that the JVM defies reason 😏

leandrob13:
@alexandru I think I addressed the easiest points to fix of your feedback. I will focus on your last comment, to add an extra option to `Options` and work out a consolidated TaskRunLoop to start comparing. Is it ok if I make a separate one and compare the 3 options?

Gonna leave the scaladoc for last so I can completely focus on documentation.

alexandru:
@leandrob13 sounds good, you can make a separate PR for comparison if you want.

I'll also pull this on my localhost today or tomorrow to play with it a little.

It's a tough feature, but I'm really glad it's shaping up and I'm positive on merging it when ready.

leandrob13:
@alexandru added the scaladocs you requested, waiting for your review on those changes.

I leave to your judgement if this is enough to be approved or you want me to do the comparison benchmarks on this one.

alexandru:
@leandrob13 I'll play with it, however, if you don't mind, I'd like another PR that introduces this as an option in `Task.Options`, to be activated by `executeWithOptions` and thus reuse the current run-loop implementation.

If it's deactivated by default, the the cost might be about two `if` statements and only on asynchronous boundaries, which shouldn't be too high.

The reason for why I'm asking is that my original motivation for introducing [Task.Options](https://monix.io/api/2.3/monix/eval/Task$$Options.html) was this use-case. Now we've got `autoCancelableLoops` in there, but we might as well have a second one:

```scala
 final case class Options(
  localContextPropagation: Boolean
)
```

alexandru:
And the motivation is that I want this option to propagate to child tasks as well, in case it is activated. As I said, if you execute a `Task` with `runAsyncTraced`, that's fine for that `Task`, but it doesn't propagate to async tasks that compose, like what `Task.gather` does and propagation is important, because the ideal is to have a single `runAsync` per program, with everything else being composition.

leandrob13:
@alexandru no problem, I will branch my current work so I can do the necessary changes. Thanks for the feedback.

alexandru:
This PR was superseded by the other one in #444, closing.

Issue #435, opened 2017-09-19
Iterant .zipWithIndex


alexandru:
Need the `zipWithIndex` operator on `Iterant`:

```scala
sealed abstract class Iterant[F[_], A] {
  //...
  def zipWithIndex(implicit F: Sync[F]): Iterant[F, (A, Long)]
}
```

Avasil:
@alexandru I would love to attempt this issue in next few days, as soon as I get more familiar with ```Iterant```

alexandru:
This was solved in #437 — btw @Avasil you can put `Fixes #issue-nr` in the description of the PR, then when we merge it, it automatically closes the ticket.

Avasil:
@alexandru 
Sorry for that, I'll have that in mind, we need to also close #434 and #421

alexandru:
@Avasil no need to apologize, it was just an FYI about how GitHub works.

PR #437, opened 2017-09-20
feat(tail) Iterant.zipWithIndex

Avasil:
My take on #435, looking forward to your feedback

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantZipWithIndex.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantZipWithIndex.scala
@@ -32,7 +32,6 @@ private[tail] object IterantZipWithIndex {
     * Implementation for `Iterant#zipWithIndex`
     */
   def apply[F[_], A](source: Iterant[F, A])(implicit F: Sync[F]): Iterant[F, (A, Long)] = {
-    
     def processSeq(index: Long, ref: NextCursor[F, A]): NextCursor[F, (A, Long)] = {
       val NextCursor(cursor, rest, stop) = ref
       val buffer = ArrayBuffer.empty[(A, Long)]
@@ -48,7 +47,6 @@ private[tail] object IterantZipWithIndex {
       }
 
       val next: F[Iterant[F, A]] = if (cursor.hasNext()) F.pure(ref) else rest
-
       NextCursor(BatchCursor.fromAnyArray(buffer.toArray[Any]), next.map(loop(idx)), stop)
     }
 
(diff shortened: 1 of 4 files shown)


monix-tail/shared/src/main/scala/monix/tail/internal/IterantZipWithIndex.scala
```diff
@@ -0,0 +1,87 @@
...
+                    (implicit F: Sync[F]): Iterant[F, (A, Long)] = {
+
+
+    def processSeq(index: Long, ref: NextCursor[F, A]): NextCursor[F, (A, Long)] = {
+      val NextCursor(cursor, rest, stop) = ref
+      val buffer = ArrayBuffer.empty[(A, Long)]
+
+      var idx = index
+
+      while (cursor.hasNext()) {
```
alexandru: This is close to what we need, however the problem here is that a `Cursor` can be infinite and if it is, then it means that it will never get out of this while loop.

To cope with such infinite sequences described by `Cursor`, we have a value on it called `recommendedBatchSize`. So you need to do ...

```scala
var toProcess = cursor.recommendedBatchSize

while (toProcess > 0 && cursor.hasNext) {
  // ...
  toProcess -= 1
}
```

Naturally, if we do this, then we can end up with an unfinished cursor, so we then need to check for that and continue from where we've left in that case:

```scala
val next = if (cursor.hasNext) F.pure(ref) else rest
NextCursor(BatchCursor.fromAnyArray(buffer.toArray[Any]), next.map(loop(idx)), stop)
```

monix-tail/shared/src/main/scala/monix/tail/internal/IterantZipWithIndex.scala
```diff
@@ -0,0 +1,87 @@
...
+        case empty@Halt(_) =>
+          empty.asInstanceOf[Iterant[F, (A, Long)]]
+      }
+      catch {
+        case NonFatal(ex) => signalError(source, ex)
+      }
+    }
+
+    source match {
+      case Suspend(_, _) | Halt(_) => loop(0)(source)
```
alexandru: In this case we don't need to suspend the execution for `Next` or `Last`, just for `NextCursor` and `NextBatch`. This is because in comparison with other operators, this one does not involve user provided functions, so we know that whatever processing we are doing for `Next` or `Last` is pure and doesn't break anything, no need to suspend those.

So this condition should be ... if NextBatch or NextCursor, then suspend, otherwise loop directly.

alexandru:
Thanks @Avasil, this is close to what's needed, fix the two comments I've added above and then it will be OK to merge.

And please add yourself to the AUTHORS file as well.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantZipWithIndex.scala
Avasil: Okay, just to make sure I understand:
```NextCursor``` and ```NextBatch``` aren't pure because ```cursor.next()``` can throw exceptions, right?

alexandru:
Thanks @Avasil, I did 2 minor whitespace fixes, but everything is fine, will merge as soon as Travis finishes.

Issue #434, opened 2017-09-19
Add Iterant.dump method


alexandru:
`Observable` has the `.dump` utility that we need for `Iterant`:

- [API Documentation](https://monix.io/api/2.3/monix/reactive/Observable.html#dump(prefix:String,out:java.io.PrintStream):Self[A])
- Source code: [Observable.scala](https://github.com/monix/monix/blob/v3.0.0-M1/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala#L1210) and [DumpObservable](https://github.com/monix/monix/blob/v3.0.0-M1/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DumpObservable.scala)

Signature: 

```scala
sealed abstract class Iterant[F[_], A] {
  // ...
  def dump(prefix: String, out: PrintStream = System.out): Iterant[F, A]
}
```

Usage:

```scala
Iterant[Task].range(0, 4)
  .dump("O")
  .completeL.runAsync
```

What this does is to dump the incoming events to standard output, like so:

```scala
O --> 0
O --> 1
O --> 2
O --> 3
O completed
```

Avasil:
I'd like to take this one as well :)

alexandru:
Ah, right, yeah, do the output like what we have with `Observable`, those index numbers are important, although you might add some spacing into `Observable` as well.

This is output meant for debugging, so people shouldn't have parsed that output anyway.

alexandru:
@Avasil and yes, feel free to work on it.

alexandru:
Sovled in #439.

PR #439, opened 2017-09-24
feat(tail) Iterant.dump

Avasil:
PR for #434 

This is pretty straightforward, I've used ``` map ``` for printing because there is no ``` foreach ```. On second thought I should probably write

``` scala
case NextCursor(cursor, rest, stop) =>
          cursor.map { el =>
              out.println(s"$pos: $prefix --> $el")
              pos += 1
          }
          NextCursor[F, A](cursor, rest.map(loop), stop)
```
instead of 
``` scala
case NextCursor(cursor, rest, stop) =>
          val dumped = cursor.map { el =>
              out.println(s"$pos: $prefix --> $el")
              pos += 1
              el
          }
          NextCursor[F, A](dumped, rest.map(loop), stop)
```

but I will wait for your opinion and fix with any other comments

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantDump.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantDump.scala
@@ -32,33 +32,36 @@ private[tail] object IterantDump {
     */
   def apply[F[_], A](source: Iterant[F, A], prefix: String, out: PrintStream = System.out)
                     (implicit F: Sync[F]): Iterant[F, A] = {
-    var pos = 0
 
-    def loop(source: Iterant[F, A]): Iterant[F, A] =
+    def loop(pos: Int)(source: Iterant[F, A]): Iterant[F, A] =
       try source match {
         case Next(item, rest, stop) =>
           out.println(s"$pos: $prefix --> $item")
-          pos += 1
-          Next[F, A](item, rest.map(loop), stop)
+
+          Next[F, A](item, rest.map(loop(pos + 1)), stop)
 
         case NextCursor(cursor, rest, stop) =>
+          var cursorPos = pos
+
           val dumped = cursor.map { el =>
-            out.println(s"$pos: $prefix --> $el")
-            pos += 1
+            out.println(s"$cursorPos: $prefix --> $el")
+            cursorPos += 1
             el
           }
-          NextCursor[F, A](dumped, rest.map(loop), stop)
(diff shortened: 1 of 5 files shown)


monix-tail/shared/src/main/scala/monix/tail/internal/IterantDump.scala
```diff
@@ -0,0 +1,88 @@
...
+import monix.tail.Iterant.{Halt, Last, Next, NextBatch, NextCursor, Suspend}
+import monix.tail.internal.IterantUtils.signalError
+
+private[tail] object IterantDump {
+  /**
+    * Implementation for `Iterant#dump`
+    */
+  def apply[F[_], A](source: Iterant[F, A], prefix: String, out: PrintStream = System.out)
+                    (implicit F: Sync[F]): Iterant[F, A] = {
+    var pos = 0
```
alexandru: I don't like this `var`, for one because it breaks referential transparency even when the `source` is a `Suspend`.

Think of referential transparency as the ability to execute the thing multiple times and get the same result. By closing over this `var` you'll get different results on different runs, as you have shared mutable state.

You don't really need a `var` with your `loop`, when you can simply do:

```scala
def loop(pos: Long)(source: Iterant[F, A]): Iterant[F, A]
```

alexandru:
@Avasil thanks for sending another PR 😀

On your question, it doesn't work, because a `Cursor` can have either strict or lazy behavior in `map`, depending on the underlying implementation ... so you can have cursors over arrays, which are strict, but you can have cursors over `Iterator`, which are lazy.

Avasil:
@alexandru 
You're right about ``` var ```, actually this was my first approach but I've had some weird behaviour for batches/cursor but it turns out I was printing wrong index... *facepalm*

I still have some vars, but completely local, so it should be referentially transparent now but i can also write it like this, I'm not sure what's better:

``` scala
val newPos = cursor.foldLeft(pos) { (cursorPos, el) =>
    out.println(s"$cursorPos: $prefix --> $el")
    cursorPos + 1
}
```

And I'm happy to contribute, if there are more newbie-friendly functions (but progressively harder!) you'd like to delegate to someone I'd love to help

Avasil:
@alexandru Well, after some thought I came to the conclusion that I can't use ``` foldLeft ``` for incrementing updated position for batch/cursor because ```foldLeft``` is side-effecting (consumes cursor) and ```dump``` shouldn't do any side-effects apart from ```IO```

PR #441, opened 2017-09-26
Observable.mapParallelUnordered needs configurable overflow strategy

### LINKED ISSUE #440: Observable.mapParallelUnordered needs configurable overflow strategy
The `mapParallelUnordered` operator needs a configurable overflow strategy, much like `merge`.

So we need to take it as an implicit, with a default value, just like `merge`, with a signature like:

```scala
final def mapParallelUnordered[B](parallelism: Int)(f: A => Task[B])
  (implicit os: OverflowStrategy[B] = OverflowStrategy.Default): Observable[B]
```

It's easy to change the implementation, although some tests might need to be fixed.

TIP: besides the signature in `Observable.scala`, the implementation that needs to change is in `monix.reactive.internal.operators.MapParallelUnorderedObservable` 😉

Wogan:
Fixes #440.

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -1582,7 +1582,7 @@ abstract class Observable[+A] extends Serializable { self =>
     * @see [[mapTask]] for serial execution
     */
   final def mapParallelUnordered[B](parallelism: Int)(f: A => Task[B])
-                                   (implicit os: OverflowStrategy[B] = OverflowStrategy.Default): Observable[B] =
+    (implicit os: OverflowStrategy[B] = OverflowStrategy.Default): Observable[B] =
     new MapParallelUnorderedObservable[A, B](self, parallelism, f, os)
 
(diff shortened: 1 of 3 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -1581,8 +1581,9 @@ abstract class Observable[+A] extends Serializable { self =>
     *
     * @see [[mapTask]] for serial execution
     */
-  final def mapParallelUnordered[B](parallelism: Int)(f: A => Task[B]): Observable[B] =
-    new MapParallelUnorderedObservable[A, B](self, parallelism, f)
+  final def mapParallelUnordered[B](parallelism: Int)(f: A => Task[B])
+                                   (implicit os: OverflowStrategy[B] = OverflowStrategy.Default): Observable[B] =
```
alexandru: I don't like this indenting style for params, there's something that annoys me greatly about things that dangle on the right, which is why this codebase doesn't do it. I know IntelliJ IDEA does this by default, but it can be configured to not do it.

Please use normal indentation for this line.

PR #457, opened 2017-11-04
Upgrade to Cats 1.0.0-RC1

alonsodomin:
This is my attempt at migrating the master branch (v3.0.0) to the latest version of Cats and Cats Effect.

Fixes #456

PR #458, opened 2017-11-07
Updates to cats 1.0-RC1 and cats-effect 0.5

### LINKED ISSUE #456: Migrate to Cats 1.0.0-RC1 and Cats Effect 0.5
Latest versions of Cats and Cats Effect have been released, it would be good to migrate the master branch dependencies to those versions.

alexandru:
Fixes #456.

This is a fork of PR #457 — which isn't sufficient because:

1. `cats.Parallel` is now in Cats
2. [cats-effect/#82](https://github.com/typelevel/cats-effect/issues/82) — `ap` has to be consistent with `flatMap` and thus preserve ordering for all `Monad` implementations

This means that what we did with the extra `import Task.nondeterminism` no longer works, however we can now defer all our parallelism needs to the new `cats.Parallel`, which is simpler and easier, simplifying our implicits a lot.

And now that we can use `cats.Parallel`, introducing:

1. `Iterant.parZip`
2. `Iterant.parZipMap`

/cc @alonsodomin


monix-eval/shared/src/main/scala/monix/eval/instances/CatsEffectForTask.scala
```diff
@@ -0,0 +1,72 @@
...
+    import CatsEffectForTask.noop
+    IO(fa.runAsync(new Callback[A] {
+      def onSuccess(value: A): Unit =
+        cb(Right(value)).unsafeRunAsync(noop)
+      def onError(ex: Throwable): Unit =
+        cb(Left(ex)).unsafeRunAsync(noop)
+    }))
+  }
+
+  /** We need to mixin [[CatsAsyncForTask]], because if we
```
Wogan: Is this true even with the priority of the traits mixed into Task? Does this also mean you could have conflicts for Monad (etc) which both CatsEffectForTask and CatsAsyncForTask implement?

alexandru: Yes, because unfortunately Scala's implicit resolution is a complex thing that nobody really knows how it works.

But the gist is this — the compiler basically calculates a score. And so this works:

```scala
class Foo
class Bar extends Foo

trait Level1 extends Level0 {
  implicit bar: Bar = new Bar
}

trait Level0 {
  implicit foo: Foo = new Foo
}
```

But this does not:

```scala
trait Level1 extends Level0 {
  implicit foo: Foo = new Foo
}

trait Level0 {
  implicit bar: Bar = new Bar
}
```

This is because `Bar <: Foo` and somehow the compiler when calculating the score of those 2 implicits will end up with the same number, the different cake levels being balanced against the subtyping relationship we have here.

Fortunately what I did works, even for `Monad`:

```scala
scala> import monix.execution.Scheduler.Implicits.global

scala> import monix.eval._
import monix.eval._

scala> implicitly[cats.Monad[Task]]
res0: cats.Monad[monix.eval.Task] = monix.eval.instances.CatsAsyncForTask$@2f600d06
```

This is because the compiler does not see `CatsEffectForTask` as being a subtype of `CatsAsyncForTask`. A weird rule, but there you have it.

monix-eval/shared/src/test/scala/monix/eval/BaseLawsSuite.scala
```diff
@@ -118,4 +118,9 @@ trait ArbitraryInstancesBase extends monix.execution.ArbitraryInstances {
     Cogen[Unit].contramap(_ => ())
   implicit def cogenForCoeval[A](implicit A: Numeric[A]): Cogen[Coeval[A]] =
     Cogen((x: Coeval[A]) => A.toLong(x.value))
+
+//  implicit val isoTask: Isomorphisms[Task] =
```
Wogan: remove?

alexandru: 👍 forgot that junk

monix-execution/shared/src/test/scala/monix/execution/BaseLawsSuite.scala
```diff
@@ -18,8 +18,9 @@
 package monix.execution
 
 import cats.Eq
-import cats.laws.IsEq
-import cats.kernel.laws._
+import cats.laws._
+//import cats.kernel.laws._
```
Wogan: also here

PR #462, opened 2017-11-11
cancel source task directly after timeout

Danny02:
Why:
The source task may still suceed after a timeout if the backup task
does not complete immediatily,
e.g. et=execution time, source-et < timeout + backup-et

This is contrary to the documentation and the intended behaviour.

Side-effects:
This is a breaking change, because as stated above code will return
in some cases the result of a different task.


alexandru:
Interesting. I'm not convinced, your changes seem to be equivalent to the old code, but I have to look at those test cases more carefully.

If you're right, I'll merge it of course.

Danny02:
For example:
* the source task takes 3 seconds to complete
* the backup task takes 2 seconds
* and the timeout is also 2 seconds

With the current implementation:
* source will start at 0:00
* backup will start (delayed by 2 seconds) at 0:02
* source will finish at 0:03 while backup is still running

The result will be that the backup task will get canceled and the calculation continues with the result of the source task while the whole computation took 3 seconds which is longer as the specified timeout.

With my code change the source task would race against an already completed task (unit) which gets delayed by the specified timeout (don't know if delayedresult would be more performand).

So if the timeout is reached without the source task completing in time, it will get canceled before the backup task is even started. It also doesn't matter anymore how long the execution of the backup task takes.

PR #463, opened 2017-11-12
Update to SBT 1.0

alexandru:
Also updates:

- Scala to 2.12.4
- JCTools to 2.1.1
- Minitest to 2.0.0
- Scala.js to 0.6.21


alexandru:
Unfortunately we've got an SBT-related problem, standing by for fix: https://github.com/sbt/sbt/issues/3730

PR #471, opened 2017-12-03
WIP: StateSubject

massung:
This is a subject pattern I have used multiple times in my projects and thought it might be useful to others, so I'm putting it here for review by others. The comments in the code should explain the usage, but here's an example:

```scala
sealed trait StackTransform

case class  Push[T](x: T) extends StackTransform
case object Pop           extends StackTransform

val stack = StateSubject[StackTransform, List[Int]](List.empty) {
  case (xs, Push(x: Int)) => x :: xs
  case (xs, Pop)          => xs drop 1
}

val watcher = stack foreach (println _)

stack onNext Push(1)
stack onNext Push(2)
stack onNext Push(3)
stack onNext Pop
stack onNext Pop
```

The above should print out:

```
List()
List(1)
List(2, 1)
List(3, 2, 1)
List(2, 1)
List(1)
```

I put together a test as well, but I haven't been able to get `sbt test` to work (a get `Error: Could not find or load main class #`). If someone didn't mind pointing out the error of my ways so I could ensure the test works?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -30,3 +30,6 @@ https://github.com/alonsodomin
 
 Leandro Bolivar
 https://github.com/leandrob13
+
+Ryo Fukumuro
+https://github.com/rfkm
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/AsyncStateActionObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/AsyncStateActionObservable.scala
@@ -46,12 +46,18 @@ class AsyncStateActionObservable[S,A](seed: => S, f: S => Task[(A,S)]) extends O
   }
 
   def loop(subscriber: Subscriber[A], state: S): Task[Unit] =
-    try f(state).flatMap { case (a, newState) =>
-      Task.fromFuture(subscriber.onNext(a)).flatMap {
-        case Continue => loop(subscriber, newState)
-        case Stop => Task.unit
+    try f(state).transformWith(
+      { case (a, newState) =>
+        Task.fromFuture(subscriber.onNext(a)).flatMap {
+          case Continue => loop(subscriber, newState)
+          case Stop => Task.unit
+        }
+      },
+      { ex =>
+        subscriber.onError(ex)
+        Task.unit
       }
-    } catch {
+    ) catch {
       case NonFatal(ex) =>
         Task.raiseError(ex)
     }
--- a/monix-reactive/shared/src/main/scala/monix/reactive/subjects/ActorSubject.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/subjects/ActorSubject.scala
@@ -36,12 +36,20 @@ package monix.reactive.subjects
   *
   *     val watcher = stack foreach (println _)
   *
-  *     stack onNext Push(1)   // prints List(1)
-  *     stack onNext Push(2)   // prints List(2, 1)
-  *     stack onNext Push(3)   // prints List(3, 2, 1)
-  *     stack onNext Pop       // prints List(2, 1)
-  *     stack onNext Pop       // prints List(1)
-  *     stack onNext Pop       // prints List()
+  *     stack onNext Push(1)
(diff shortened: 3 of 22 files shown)


Avasil:
> I put together a test as well, but I haven't been able to get `sbt test` to work (a get `Error: Could not find or load main class #`). If someone didn't mind pointing out the error of my ways so I could ensure the test works?

I have similiar problem when using Windows, my workaround is using sbt shell provided by IntelliJ. Select it instead of terminal and type:
```
project reactiveJVM
test
```

massung:
Thanks for the IntelliJ suggestion @Avasil (I am on Windows). At this point everything is good except the StateSubjectSuite. I'm not quite sure how the `alreadyTerminatedTest` is supposed to function, so what I've implemented is *not* a valid test.

Could someone explain to me the goal of the test? It would appear as though it's expecting each the subject to get a list of random numbers and then an observer of the test subject is summing up those values (expecting them to be passed through). However, my test subject isn't receiving any of the elements. I'm assuming this is a bug in my implementation of `StateSubject`, but I'm not seeing it.

If this subject is something others find useful and would like in the library, I'd like to understand a bit more so I can finish this up and add to the community. ;-)

alexandru:
@massung I apologize for procrastinating on a reply for your PR, I know you've put effort in it, 

I've been busy unfortunately with other things and forgot about it and when you ignore things on your plate it becomes like a vicious circle 🙂

The thing is, like your code shows, this isn't anything more than a `PublishSubject` + `state` and the code isn't much different than just doing ...

```scala
sealed trait StackTransform

case class  Push[T](x: T) extends StackTransform
case object Pop           extends StackTransform

val stackIn = ConcurrentSubject.publish[StackTransform]
val stackOut = stack.state(List.empty) {
  case (xs, Push(x: Int)) => x :: xs
  case (xs, Pop)          => xs drop 1
}

val watcher = out foreach (println _)

stack onNext Push(1)
stack onNext Push(2)
stack onNext Push(3)
stack onNext Pop
stack onNext Pop
```

When dealing with Subjects, having an `(in, out)` pair is very common, we actually have this formalized in the `Pipe` type, which is the `Subject` factory, but instead of giving you a `Subject`, it gives you an `(in, out)` pair.

So your subject is very cool, unfortunately it just gives a name to a pattern that can be accomplished with about the same level of effort.

And I would like to add it, but when adding stuff to a library, the problem is that it's way harder to take it out, due to backwards compatibility concerns, so you're then stuck with it. Plus giving a name to a pattern is a way of signaling to users that they should learn this stuff, but it also increases the amount of stuff that users need to learn. 

Right now there are 5 `Subject` types and one of them, the `AsyncSubject`, is totally unnecessary, distracting and I'm now sorry for implementing it, the reason being that it was in Rx.NET and thought that it is useful. So now we're stuck with it. But at least it's welcoming for Rx.NET users migrating over and we have a couple of those.

So adding another `Subject` in that package increases the surface area that needs to be maintained and that users need to learn and I'd rather not do it.

I see this was your "First Time Contribution" and the code quality is pretty good. So I hope my rejection and big latency in responding won't turn you off completely from making future contributions 😐

Cheers,

oleg-py:
> `Error: Could not find or load main class #`

BTW, I had this issue on Windows too. This happens because of hash-comments in `.jvmopts` file, which default SBT launcher uses, while Intellij does not.

The solution I use is to remove comments from .jvmopts and then do `git update-index --assume-unchanged .jvmopts` to make git ignore that file in commits.

massung:
> I see this was your "First Time Contribution" and the code quality is pretty good. So I hope my rejection and big latency in responding won't turn you off completely from making future contributions.

LOL. Not at all and when I put this together it was only my 2nd time using Monix and found that I had implemented the same pattern twice in a row. I just figured I'd "formalize" it for myself and see if it'd be useful to others.

Thanks for the kind feedback, and - knowing what I know now - I'm 100% in agreement and (in your place) would do the same thing.

alexandru:
@massung cool and thanks.

Will close this for now.

Issue #380, opened 2017-07-01
Observable recovery does not work


gabor-aranyossy:
After playing around with the reactive extensions I noticed a weird behaviour when trying to recover an exploding `Observable`. Instead of calling the recovery code it stops the stream and calls the `Scheduler`'s `reportFailure`. My interpretation might by wrong, but [this](https://github.com/gabor-aranyossy/monix/commit/d84884e9c11d7c5084b8266f03e98ddd12745f79) I think should work.

My expectation was that this works the same way as [scalaz-stream](https://github.com/scalaz/scalaz-stream) and other similar libraries.

alexandru:
@gabor-aranyossy throwing exceptions is problematic and sometimes we can't recover, or to be more precise, we can only recover in case the final `onComplete` event wasn't sent already.

However if an onComplete already happened, then we can no longer signal a subsequent onError event, because that would break the contract, so the only thing we can do is to log it somewhere. And sometimes the behavior is undefined.

Note that Scalaz-stream uses a different communication protocol so in the presence of exceptions it will break in different ways.

Please paste a code sample, so that I can help.

Note that I remember answering this on the Gitter channel. If it wasn't you, then sorry for the delay, it's summer time, vacations, etc 😁

gabor-aranyossy:
No problem for the delay. :) Here is the code I was testing with: [ErrorHandling.scala](https://gist.github.com/gabor-aranyossy/dc8205044e990e832a2c2651ab1a3487)

I was using "io.monix" %% "monix" % "2.3.0"

alexandru:
This is no longer reproducible on `master`. I think it's fixed in #473.

gabor-aranyossy:
thanks for the fix! 👍

PR #473, opened 2017-12-09
Fix issue where `fromAsyncStateAction` is not safe for user code

rfkm:
An observable created by `fromAsyncStateAction` never completes even if a state function returns a failed task since it doesn't propagate its error to downstream.

This might be related to:  https://github.com/monix/monix/issues/380

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -30,3 +30,6 @@ https://github.com/alonsodomin
 
 Leandro Bolivar
 https://github.com/leandrob13
+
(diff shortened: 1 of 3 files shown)


alexandru:
Thanks @rfkm

alexandru:
Would you mind backporting your fix to the 2.3.x branch as well?

rfkm:
@alexandru Thank you for merging! I've created backport PR: #475

PR #482, opened 2017-12-19
Added task.cancelable

### LINKED ISSUE #477: Add .interruptible on Task
This would be the equivalent of ...

```scala
task.executeWithOptions(_.enableAutoCancelableRunLoops)
```

Avasil:
Fixes #477 .


alexandru:
Sorry for not saying anything, I'm still thinking about this. The problem with adding stuff is that it's hard to remove later should we change our mind.

alexandru:
Please rename to `.cancelable` and merge with `master`.

Issue #484, opened 2017-12-23
Upgrade to cats 1.0.0-RC2 and cats-effect 0.6


lorandszakacs:
Well, everything is in the header really.

Hopefully by doing this the switch to 1.0.0 will be easier.

LukaJCB:
To clarify, 1.0.0 will have zero source code changes over RC2, only tests and documentation are affected :)

alexandru:
Hey, latest `3.0.0-M3` depends on Cats 1.0.1 and cats-effect 0.8. I'm closing this.

Sorry for the latency, I've been on my winter holiday.

Cheers,

PR #485, opened 2017-12-23
Upgrade to cats 1.0.0 rc2 and cats-effect 0.6

lorandszakacs:
Addressing issue #484. Ought to be quite straightforward.

Replaced one single deprecated operation `Apply.followedBy` with its equivalent, syntactically symbolic, cousin, `*>`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,8 +9,8 @@ addCommandAlias("ci-jvm",     ";clean ;coreJVM/test:compile ;coreJVM/test")
 addCommandAlias("ci-js",      ";clean ;coreJS/test:compile  ;coreJS/test")
 addCommandAlias("release",    ";project monix ;+publishSigned ;sonatypeReleaseAll")
 
-val catsVersion = "1.0.0-RC1"
-val catsEffectVersion = "0.5"
+val catsVersion = "1.0.0-RC2"
+val catsEffectVersion = "0.6"
 val jcToolsVersion = "2.1.1"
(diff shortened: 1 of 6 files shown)


lorandszakacs:
Current failure of the 2.11.11 build, smells of a concurrency bug to me. Since I can't reproduce the failure. Will investigate after I decorate a tree 😄 

```
[0m[[0m[0minfo[0m] [0m[0m[32mmonix.eval.TaskBlockingSuite[0m[0m
2017-12-23 13:54:06,886 Log4j2-TF-1-AsyncLogger[AsyncContext@dc9876b]-1 ERROR An exception occurred processing Appender testbuffer-229 java.lang.NullPointerException
	at scala.collection.mutable.ListBuffer.copy(ListBuffer.scala:444)
	at scala.collection.mutable.ListBuffer.$plus$eq(ListBuffer.scala:171)
	at sbt.internal.util.BufferedAppender.append(BufferedLogger.scala:44)
	at org.apache.logging.log4j.core.config.AppenderControl.tryCallAppender(AppenderControl.java:156)
       ....
```

alexandru:
The timeout error in `AsyncSchedulerSuite` might be a problem due to Travis — when we have actual async tests with timeouts, Travis can be really slow when running the test, leading to errors, which isn't reproducible on our localhost due to our localhost not slowing down the tests. In such cases we can simply increase the timeout.

lorandszakacs:
@alexandru. I see.

PR #488, opened 2017-12-26
Add note about execution model for Observable.fromInputStream.

olafurpg:
I was playing around with Observable.fromInputStream to read from
System.in and was struggling to understand why calling `.subscribe()`
blocked when using the default global scheduler. I expected the default
scheduler to block on one of its own threads leaving my program to
continue after `.subscribe()`.

I have no idea how is best to explain this, feel free to close if you have a
better way to put it. Just wanted to send a PR since this took me a while to
figure out and thought it might be worth a mention in the otherwise fantastic docs.

PS. Thanks a lot for your amazing work, I am learning so much reading the Monix
sources and docs.


alexandru:
Thanks @olafurpg — as a matter of policy the current preference is to not start with an automatic fork if not needed.

However you can trigger a fork by yourself:

```scala
Observable.fromInputStream(in).executeWithFork
```

olafurpg:
Thank you @alexandru for clarifying the description! LGTM 👍

PR #494, opened 2018-01-07
Task.cancel as a pure action with .start, .race and .uncancelable

alexandru:
New additions to `Task`, making its API purer and more powerful out of the box:

```scala
sealed abstract class Task[+A]
  // ...
  def cancel: Task[Unit]

  def start: Task[Task[A]]

  def uncancelable: Task[A]
}

object Task {
  // ...
  def race[A, B](fa: Task[A], fb: Task[B]): Task[Either[A, B]] = ???

  def raceMany[A](tasks: TraversableOnce[Task[A]]): Task[A] = ???

  def racePair[A,B](fa: Task[A], fb: Task[B]): Task[Either[(A, Task[B]), (Task[A], B)]] = ???
} 
```

NOTE — no changes to the run-loop or internal encoding have been made, these are simply operations that were added on top of the existing implementation, but needed nevertheless for a good out of the box experience.

In the future changes to the internal encoding might happen. For example at this point there's no special state that describes cancellation, the result of `.cancel` being a `Task.Async` that makes use of the existing cancellation protocol. But in the future we might add a special state for it, depending on feedback.

## Pure cancellation

It is now possible to do this:

```scala
val ta: Task[A] = ???
val tb: Task[B] = ???

Task.racePair(ta, tb).flatMap {
  case Left((a, taskB)) =>
    taskB.cancel.map(_ => a)
  case Right((taskA, b)) =>
    taskA.cancel.map(_ => b)
}
```
 
You might recognize this operation as the old `Task.chooseFirstOf`, however the old signature was returning `CancelableFuture` as the reference for the future result of the losing task, whereas this one returns a pure `Task` (linked to an underlying `Promise`, but that's an implementation detail). And because now `Task` has a pure `.cancel` operation that describes cancellation, you can cancel these tasks in `racePair` without side effects being involved.

```scala
val task = Task.eval(println("Hello!"))
  .delayExecution(1.second)
  .doOnCancel(Task.eval(println("Cancelled!")))

task.cancel.runAsync
//=> Cancelled!
```

## .start

Inspired by [FS2](https://github.com/functional-streams-for-scala/fs2), we can now trigger `start` as a pure operation, allowing operations like this:

```scala
def par2[A, B](ta: Task[A], tb: Task[B]): Task[(A, B)] =
  for {
    fa <- ta.start
    fb <- tb.start
     a <- fa
     b <- fb
  } yield (a, b)
```

Note this is not exactly equivalent with `parMap2` due to insufficient behavior when an error happens. To make this behave like `parMap2` is harder, so usage of `parMap2` (and `parMap3`, etc) is still recommended, this being just a pedagogical example for what `start` might be useful for.

Also compared with the equivalent from FS2, Monix's `start` does not introduce an automatic `fork` (i.e. it does not shift the processing on another thread, although shifting might happen due to batched processing), being consistent with other operations in Monix.

## .uncancelable

Sometimes we need to ensure that tasks are not cancelable, in order to ensure that some actions are executed no matter what and now you can use the new `uncancelable` operator to mark tasks that cannot be canceled:

```scala
Task(println("Hello!")).delayExecution(5.seconds)
  .doOnFinish(_ => println("Executed!"))
  .uncancelable
```

Normally the above task is cancelable and there's plenty of time to trigger cancellation (due to the `delayExecution`), so if we want to ensure that this task executes as an atomic unit (all or nothing), we can call `.uncancelable` on it.

## race, racePair, raceMany

The old `.chooseFirstOf` was too powerful. Most of the time you just want to pick the winner and cancel the loser. Plus `race` as a name is much more standard, having been used both in Haskell and JavaScript's standard libraries at least.

`race` is the new version that automatically cancels the loser.

`racePair` is the new version that behaves like the old `chooseFirstOf`, but is pure, giving you a `Task` for the loser, since tasks can now be cancelled directly as a pure action.

`raceMany` is just the old `chooseFirstOfList`, to be consistent with the other two.

## Other Changes

Refactoring to `monix.execution.cancelables`:

- rename `MultiAssignmentCancelable` to `OrderedCancelable`
- introduce new `MultiAssignCancelable` with a simpler implementation
- rename `SingleAssignmentCancelable` to `SingleAssignCancelable`
- add deprecated symbols in `cancelable` package object

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -396,6 +396,73 @@ sealed abstract class Task[+A] extends Serializable {
   final def attempt: Task[Either[Throwable, A]] =
     FlatMap(this, AttemptTask.asInstanceOf[A => Task[Either[Throwable, A]]])
 
+  /** Introduces an asynchronous boundary at the current stage in the
+    * asynchronous processing pipeline.
+    *
+    * Consider the following example:
+    *
+    * {{{
+    *   import monix.execution.Scheduler
+    *   val io = Scheduler.io()
+    *
+    *   val source = Task(1).executeOn(io).map(_ + 1)
+    * }}}
+    *
+    * That task is being forced to execute on the `io` scheduler,
+    * including the `map` transformation that follows after
+    * `executeOn`. But what if we want to jump with the execution
+    * run-loop on the default scheduler for the following
+    * transformations?
+    *
+    * Then we can do:
+    *
+    * {{{
+    *   source.asyncBoundary.map(_ + 2)
+    * }}}
+    *
+    * In this sample, whatever gets evaluated by the `source` will
+    * happen on the `io` scheduler, however the `asyncBoundary` call
+    * will make all subsequent operations to happen on the default
+    * scheduler.
+    */
+  final def asyncBoundary: Task[A] =
+    this.flatMap(r => Task.shift.map(_ => r))
+
+  /** Introduces an asynchronous boundary at the current stage in the
+    * asynchronous processing pipeline, making processing to jump on
+    * the given [[monix.execution.Scheduler Scheduler]] (until the
+    * next async boundary).
+    *
+    * Consider the following example:
+    * {{{
+    *   import monix.execution.Scheduler
+    *   val io = Scheduler.io()
+    *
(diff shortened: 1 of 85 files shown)


alexandru:
Merging, so I can release 3.0.0-M3, will address potential issues later.

PR #503, opened 2018-01-08
Add Iterant.liftF builder for lifting monadic values

### LINKED ISSUE #487: Iterant.liftF (F[A] => Iterant[F, A]) builder
I've been doing a port from `fs2.Stream` to `monix.Iterant` on one of my personal projects. One of usage patterns I encountered was including effectful values (that are not streams) in the middle of `for`-comprehension:
```scala
for {
  a <- /* streaming operation */
  b <- Stream.eval(/* non-streaming operation dep. on a */)
  c <- /* another streaming operation */
  _ <- Stream.eval(/* another non-streaming operation of b and c */)
} yield c
```
Where `Stream.eval` has the following type signature:

```scala
def eval[F[_],O](fo: F[O]): Stream[F,O] = ???
```

I propose to add such builder to `Iterant`. Proposing name `liftF` based on cats naming for monad transformers' method that does similar thing (see typelevel/cats#2033).

If you don't mind, I'd take a run at making PR for this.

oleg-py:
Closes #487


monix-tail/shared/src/test/scala/monix/tail/IterantBasicSuite.scala
```diff
@@ -59,6 +59,20 @@ object IterantBasicSuite extends BaseTestSuite {
     assertEquals(f.value, Some(Success(1)))
   }
 
+  test("Iterant.liftF") { implicit s =>
```
alexandru: We probably need a test using ScalaCheck too.

```scala
test("any F can be lifted) { implicit s =>
  check1 { (fa: IO[Int]) =>
    Iterant[IO].liftF(fa).headOptionL <-> fa.map(Some.apply)
  }
}
```

Or something along those lines.

alexandru:
Thanks @oleg-py, looks good, see my comment on also adding a ScalaCheck-powered test.

Let me know if you need help.

After that we can merge.

oleg-py:
Good idea. Done 😄

PR #504, opened 2018-01-08
add Iterant.switchIfEmpty operator

### LINKED ISSUE #499: Add Iterant.switchIfEmpty
Need an op for `Iterant` that switches to an alternative stream if the source is empty.

```scala
  // ...
  def switchIfEmpty(fallback: Iterant[F, A])(implicit F: Sync[F]): Iterant[F, A]
}
```

Similar with the `Observable.switchIfEmpty`, see [the implementation](https://github.com/monix/monix/blob/v3.0.0-M3/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/SwitchIfEmptyObservable.scala) for inspiration.

oleg-py:
Closes #499


monix-tail/shared/src/main/scala/monix/tail/internal/IterantSwitchIfEmpty.scala
```diff
@@ -0,0 +1,27 @@
...
+import cats.effect.Sync
+import monix.tail.Iterant
+import monix.tail.Iterant._
+
+private[tail] object IterantSwitchIfEmpty {
+  def apply[F[_], A](primary: Iterant[F, A], backup: Iterant[F, A])(implicit F: Sync[F]): Iterant[F, A] = {
+    def loop(source: Iterant[F, A]): Iterant[F, A] =
+      source match {
+        case Suspend(rest, stop) => Suspend(rest.map(loop), stop)
+        case NextBatch(batch, rest, stop) if batch.cursor().isEmpty =>
```
alexandru: Here we have a problem. As an implementation detail of `Iterant`, the `NextBatch` and `NextCursor` are wrapping the equivalent of an `Iterable` and `Iterator`, but once you start operating on iterators, you end up with side-effects.

So `batch.cursor().isEmpty` is side-effectful and we can't allow that without suspending it.
Take a look for example at [.take](https://github.com/monix/monix/blob/v3.0.0-M3/monix-tail/shared/src/main/scala/monix/tail/internal/IterantTake.scala#L88) where we are doing something like this:

```scala
    source match {
      case NextBatch(_, _, _) | NextCursor(_, _, _) =>
        // We can have side-effects with NextBatch/NextCursor
        // processing, so suspending execution in this case
        Suspend(F.delay(loop(n)(source)), source.earlyStop)
      case _ =>
        loop(n)(source)
    }
}
```

We need the same here.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantSwitchIfEmpty.scala
```diff
@@ -0,0 +1,27 @@
...
+          Suspend(rest.map(loop), stop)
+        case NextCursor(cursor, rest, stop) if cursor.isEmpty =>
+          Suspend(rest.map(loop), stop)
+
+        case Halt(None) =>
+          Suspend(primary.earlyStop.as(backup), F.unit)
```
alexandru: We don't need to call `primary.earlyStop` here, because `primary` is already at `Halt(None)` according to your pattern matching and there is no "early stop" remaining to trigger. We know the stream is done, so at this point you can just return `backup`.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantSwitchIfEmpty.scala
alexandru: Another thing — once you evaluate `batch.cursor()`, given that it is side-effectful, it's wasteful to still return `source` in case the generated cursor is NOT empty.

It's better at this point to transform that node into a `NextCursor`. So...

```scala
case NextBatch(batch, rest, stop) =>
  val cursor = batch.cursor()
  if (cursor.isEmpty) 
     Suspend(rest.map(loop), stop)
  else
    NextCursor(cursor, rest, stop)
```

alexandru: And yet another implementation detail ... we assume that `batch.cursor()` is hostile and can throw (after all we expect side effects from it), the general idea being that you can take any dirty Java Iterable and wrap it into a batch or cursor.

So we also need a `try / catch` that wraps this pattern matching and that returns `Halt(error)`.
See [IterantTake](https://github.com/monix/monix/blob/v3.0.0-M3/monix-tail/shared/src/main/scala/monix/tail/internal/IterantTake.scala#L81) for this one as well.

oleg-py:
Thank you @alexandru for your insightful comments. I've adjusted the PR.

My initial assumptions on how `stop` values were used were not exactly correct. I thought it should've executed always on completion, which was not right, and so my tests were not corrected.

However, once I added suspending, I was not able to construct a test case which would make the try/catch required (i.e. everything seems to work the same way without it (for EitherT[Eval, Throwable, A], IO, Coeval and Task)

alexandru:
@oleg-py that `earlyStop` is for early termination; when you receive a `Halt` then it means the producer has already terminated whatever underlying data source it had.

Think about it like this — when you're traversing a plain iterator like the following, the producer that gave you the iterator will probably close any connections when the iterator's `hasNext` returns `false`.

```scala
iterator.foreach { x => println(x) }
```

However if you do this ...

```scala
iterator.take(10).foreach(println)
```

Then in this case you're in trouble, because you're just taking `10` elements and ignore the rest. Which means the stream will never complete and the underlying data source will never close.

With `Iterant` however, this is an "early termination". The end of the stream hasn't been reached, but by using `earlyStop` you're signaling to the producer that you're no longer interested in the rest of the stream.

alexandru:
For generating errors, built a fake `Batch` with [ThrowExceptionBatch](https://github.com/monix/monix/blob/master/monix-tail/shared/src/test/scala/monix/tail/ThrowExceptionBatch.scala) or a fake `Cursor` with [ThrowExceptionCursor](https://github.com/monix/monix/blob/master/monix-tail/shared/src/test/scala/monix/tail/ThrowExceptionCursor.scala).

oleg-py:
@alexandru yes, I've figured that out already 👍 .
I was assuming that

    Suspend(F(stream), effect) <-> stream.doOnComplete(_ => effect)

 instead of being

     Suspend(F(stream), effect) <-> stream.doOnEarlyStop(effect)

---

I do have tests with broken batches & cursors. These have demonstrated to me where `try ... catch` is not enough and I need to use `suspend`. I'm saying that I was not able to construct a test case that would show that `suspend` is not enough and `try ... catch` is required (although it doesn't hurt either).

alexandru:
The purpose of a `try ... catch` is to treat the error using a `Halt` instead of letting the error be caught by the underlying `F[_]` (e.g. `Task`).

I think you can prove it by seeing if `earlyStop` is called in case of an error.

PR #508, opened 2018-01-09
Add Iterant.intersperse operation

### LINKED ISSUE #501: Add Iterant.intersperse
We need an `intersperse` similar with the one on `Observable`, with the following description:

```scala
  // ...

  /** Creates a new stream from the source that will emit a specific `separator`
    * between every pair of elements.
    *
    * @param separator is the separator
    */
  def intersperse(separator: A)(implicit F: Sync[F]): Iterant[F, A]

  /** Creates a new stream from the source that will emit the `start` element
    * followed by the upstream elements paired with the `separator` 
    * and lastly the `end` element.
    *
    * @param start is the first element emitted
    * @param separator is the separator
    * @param end the last element emitted
    */
  def intersperse(start: A, separator: A, end: A)(implicit F: Sync[F]): Iterant[F, A]
```

oleg-py:
Closes #501 

Hope you're not annoyed by my pull requests yet 😅 

Not too many tests for this one. I found that ScalaCheck-powered tests check pretty much every stream structure, and since `intersperse` signature requires pure values, there's not many erroneous cases possible.

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantIntersperse.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantIntersperse.scala
@@ -108,7 +108,8 @@ private[tail] object IterantIntersperse {
     source match {
       case NextCursor(_, _, _) | NextBatch(_, _, _) =>
         Suspend(F.delay(loop(prepend = false)(source)), source.earlyStop)
-      case _ => loop(prepend = false)(source)
+      case _ =>
+        loop(prepend = false)(source)
     }
   }
 }
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantIntersperse.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantIntersperse.scala
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2014-2018 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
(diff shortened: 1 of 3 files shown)


alexandru:
@oleg-py not annoyed, no way, keep them coming 😀

I might have a latency in reviewing PRs though, due to work. Will take a look tonight.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantIntersperse.scala
```diff
@@ -0,0 +1,114 @@
...
+        case ex if NonFatal(ex) =>
+          val stop = source.earlyStop
+          Suspend(stop.map(_ => Halt(Some(ex))), stop)
+      }
+    }
+
+    source match {
+      case NextCursor(_, _, _) | NextBatch(_, _, _) =>
+        Suspend(F.delay(loop(prepend = false)(source)), source.earlyStop)
+      case _ => loop(prepend = false)(source)
```
alexandru: A minor matter of style / indentation — consistency is good as it makes it easier to read, probably because the arrangement is more predictable, so in a `match` statement if you have a branch with the expression on a new line above, it's better if you keep that for the rest of the branches, e.g...

```scala
source match {
  case NextCursor(_, _, _) | NextBatch(_, _, _) =>
    Suspend(F.delay(loop(prepend = false)(source)), source.earlyStop)
  case _ => 
    loop(prepend = false)(source) 
}
```

PR #509, opened 2018-01-09
implementation for Iterant#dropLast

### LINKED ISSUE #495: Add Iterant.dropLast
We need an `Iterant` op that drops the last `n: Int` elements from the stream.

The method signature should be something like this:

```scala
  // ...
  def dropLast(n: Int)(implicit F: Sync[F]): Iterant[F, A]
}
```

For inspiration check out [DropLastOperator](https://github.com/monix/monix/blob/v3.0.0-M3/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DropLastOperator.scala), the implementation for `Observable`.

Avasil:
Fixes #495.

There are a lot of mutable things but it should mantain referential transparency if I'm not missing anything. 

I would appreciate feedback about used data structures in particular. I have tried to make it sort of efficient but I might have overcomplicated it. :D

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
@@ -57,13 +57,13 @@ private[tail] object IterantDropLast {
 
     def finalCursor(toDrop: Int, length: Int, queue: mutable.Queue[A]): Iterant[F, A] = {
       var i = length
-      var result = List[A]()
+      val buffer = mutable.Buffer[A]()
 
       while (i > toDrop) {
-        result ::= queue.dequeue()
+        buffer.append(queue.dequeue())
         i -= 1
       }
-      val cursor = BatchCursor.fromSeq(result)
+      val cursor = BatchCursor.fromSeq(buffer)
       NextCursor(cursor, F.pure(Halt(None)), F.unit)
     }
 
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
@@ -33,15 +33,19 @@ private[tail] object IterantDropLast {
   def apply[F[_], A](source: Iterant[F, A], n: Int)(implicit F: Sync[F]): Iterant[F, A] = {
 
     def processCursor(toDrop: Int, length: Int, queue: mutable.Queue[A],
-                      cursor: BatchCursor[A], rest: F[Iterant[F, A]], stop: F[Unit]): Iterant[F, A] = {
-
+                      ref: NextCursor[F, A]): Iterant[F, A] = {
+      val NextCursor(cursor, rest, stop) = ref
       val limit = cursor.recommendedBatchSize
-      var queueLength = length
       val buffer = mutable.Buffer[A]()
-      var bufferLength = 0
 
-      while (cursor.hasNext()) {
+      var bufferLength = 0
+      var queueLength = length
+      var toProcess = limit
+
+      // protects against infinite cursors
+      while (toProcess > 0 && cursor.hasNext()) {
         queue.enqueue(cursor.next())
+        toProcess -= 1
         if (queueLength >= toDrop && bufferLength < limit) {
           buffer.append(queue.dequeue())
           bufferLength += 1
@@ -49,10 +53,9 @@ private[tail] object IterantDropLast {
         else queueLength += 1
       }
 
-      if (bufferLength > 0)
-        NextBatch(Batch.fromSeq(buffer), rest.map(loop(toDrop, queueLength, queue)), stop)
-      else
-        Suspend(rest.map(loop(toDrop, queueLength, queue)), stop)
+      val next: F[Iterant[F, A]] = if (cursor.hasNext()) F.pure(ref) else rest
+
+      NextBatch(Batch.fromSeq(buffer), next.map(loop(toDrop, queueLength, queue)), stop)
     }
 
     def finalCursor(toDrop: Int, length: Int, queue: mutable.Queue[A]): Iterant[F, A] = {
@@ -76,10 +79,10 @@ private[tail] object IterantDropLast {
             Next(queue.dequeue(), rest.map(loop(toDrop, length, queue)), stop)
           else
             Suspend(rest.map(loop(toDrop, length + 1, queue)), stop)
-        case NextCursor(cursor, rest, stop) =>
-          processCursor(toDrop, length, queue, cursor, rest, stop)
+        case ref@NextCursor(_, _, _) =>
+          processCursor(toDrop, length, queue, ref)
         case NextBatch(batch, rest, stop) =>
-          processCursor(toDrop, length, queue, batch.cursor(), rest, stop)
+          processCursor(toDrop, length, queue, NextCursor(batch.cursor(), rest, stop))
         case Suspend(rest, stop) =>
           Suspend(rest.map(loop(toDrop, length, queue)), stop)
         case Last(item) =>
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
@@ -22,8 +22,10 @@ import cats.syntax.all._
 import monix.execution.misc.NonFatal
 import monix.tail.Iterant
 import monix.tail.Iterant.{Halt, Last, Next, NextBatch, NextCursor, Suspend}
-import monix.tail.batches.{Batch, BatchCursor}
+import monix.tail.batches.BatchCursor
 
+import scala.annotation.tailrec
+import scala.collection.immutable.Queue
 import scala.collection.mutable
 
 private[tail] object IterantDropLast {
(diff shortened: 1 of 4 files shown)


monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
```diff
@@ -0,0 +1,109 @@
...
+
+      var bufferLength = 0
+      var queueLength = length
+      var toProcess = limit
+
+      // protects against infinite cursors
+      while (toProcess > 0 && cursor.hasNext()) {
+        queue.enqueue(cursor.next())
+        toProcess -= 1
+        if (queueLength >= toDrop && bufferLength < limit) {
```
alexandru: I don't think `bufferLength` can ever go over `limit`, since you already guard for `toProcess > 0`.  I might be mistaken, but I think you can get rid of both `bufferLength` and `limit`.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
```diff
@@ -0,0 +1,109 @@
...
+        case halt@Halt(Some(_)) =>
+          halt
+      } catch {
+        case ex if NonFatal(ex) =>
+          val stop = source.earlyStop
+          Suspend(stop.map(_ => Halt(Some(ex))), stop)
+      }
+    }
+
+    source match {
```
alexandru: Well, if you're going to suspend the whole thing, you don't need a `match`.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
```diff
@@ -0,0 +1,109 @@
...
+
+import scala.collection.mutable
+
+private[tail] object IterantDropLast {
+  /**
+    * Implementation for `Iterant#dropLast`
+    */
+  def apply[F[_], A](source: Iterant[F, A], n: Int)(implicit F: Sync[F]): Iterant[F, A] = {
+
+    def processCursor(toDrop: Int, length: Int, queue: mutable.Queue[A],
```
alexandru: Using a mutable data structure, if properly encapsulated, is fine in Monix's codebase 😀

However I try to avoid optimizations without having some proof that they work. Because if say this optimization here only brings marginal improvements, then we shouldn't do it because an immutable `Queue` is more reasonable.

So you can either revert to an immutable `Queue` for now, or make a JMH test in `benchmarks` that proves a mutable queue is better for a stream of 100000 events where you're dropping the last 1, 10 or 50 items lets say.

Don't commit the benchmark, just something to report back that indeed this is worth it.

alexandru:
Thanks @Avasil, see my comments.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
Avasil: I think you're right, about 15 minutes ago I realized I have bugs related to infinite cursors and it became irrelevant but I've forgotten to get rid of it, will do it in the evening, thanks

monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropLast.scala
Avasil: I generally prefer to follow the same rule but I feel like operating on ```immutable.Queue``` is pretty weird, although it might be better if there is implementation similar to how ```State``` works. 

Below are results from benchmarks but please tell me if I did some benchmarking "faux pas", I know there are a lot of catches which make results false.

Results:
``` scala
jmh:run -i 50 -wi 10 -f 2 -t 1 monix.benchmarks.IterantDropLastBenchmark

mutable.Queue
[info] Benchmark                         Mode  Cnt   Score    Error  Units
[info] IterantDropLastBenchmark.drop1   thrpt  100  73,519 â–’ 4,664  ops/s
[info] IterantDropLastBenchmark.drop10  thrpt  100  79,834 â–’ 1,388  ops/s
[info] IterantDropLastBenchmark.drop50  thrpt  100  76,795 â–’ 3,076  ops/s

immutable.Queue
[info] Benchmark                         Mode  Cnt    Score   Error  Units
[info] IterantDropLastBenchmark.drop1   thrpt  100  113,273 â–’ 2,272  ops/s
[info] IterantDropLastBenchmark.drop10  thrpt  100  117,181 â–’ 1,738  ops/s
[info] IterantDropLastBenchmark.drop50  thrpt  100  119,765 â–’ 1,839  ops/s
```
Assuming those results are true... well that's a good lesson :D I was sure that all those extra tuple allocations in ```immutable.Queue``` would show in performance. I guess ```immutable.Queue``` is better implemented?

Benchmark code:
```scala
@State(Scope.Thread)
@BenchmarkMode(Array(Mode.Throughput))
@OutputTimeUnit(TimeUnit.SECONDS)
class IterantDropLastBenchmark {

  @Benchmark
  def drop1(): List[Int] = IterantDropLastBenchmark.test(100000, 1)

  @Benchmark
  def drop10(): List[Int] = IterantDropLastBenchmark.test(100000, 1)

  @Benchmark
  def drop50(): List[Int] = IterantDropLastBenchmark.test(100000, 1)

}

object IterantDropLastBenchmark {
  def test(size: Int, toDrop: Int): List[Int] = {
    val list = List.range(1, size)
    val stream = Iterant[Task].fromList(list)

    Await.result(stream.dropLast(toDrop).toListL.runAsync, Duration.Inf)
  }
}
```
This is code for ```Iterant#dropLast``` using ```immutable.Queue``` (I have left it as a mutable-immutable hybrid monster to make this comparison more fair):
```scala
package monix.tail.internal

import cats.effect.Sync
import cats.syntax.all._
import monix.execution.misc.NonFatal
import monix.tail.Iterant
import monix.tail.Iterant.{Halt, Last, Next, NextBatch, NextCursor, Suspend}
import monix.tail.batches.{Batch, BatchCursor}

import scala.collection.immutable.Queue
import scala.collection.mutable

private[tail] object IterantDropLast {
  /**
    * Implementation for `Iterant#dropLast`
    */
  def apply[F[_], A](source: Iterant[F, A], n: Int)(implicit F: Sync[F]): Iterant[F, A] = {

    def processCursor(toDrop: Int, length: Int, queue: Queue[A],
                      ref: NextCursor[F, A]): Iterant[F, A] = {
      val NextCursor(cursor, rest, stop) = ref
      val limit = cursor.recommendedBatchSize
      val buffer = mutable.Buffer[A]()
      var queueLength = length

      @tailrec
      def queueLoop(queue: Queue[A], toProcess: Int): Queue[A] = {
        if (!(toProcess > 0 && cursor.hasNext())) {
          queue
        } else {
          val newQueue = queue.enqueue(cursor.next())
          if (queueLength >= toDrop) {
            val (a, nextQueue) = newQueue.dequeue
            buffer.append(a)
            queueLoop(nextQueue, toProcess - 1)
          } else {
            queueLength += 1
            queueLoop(newQueue, toProcess - 1)
          }
        }
      }
      val finalQueue = queueLoop(queue, limit)

      val next: F[Iterant[F, A]] = if (cursor.hasNext()) F.pure(ref) else rest
      NextBatch(Batch.fromSeq(buffer), next.map(loop(toDrop, queueLength, finalQueue)), stop)
    }

    def finalCursor(toDrop: Int, length: Int, queue: Queue[A]): Iterant[F, A] = {
      var queueLength = length
      val buffer = mutable.Buffer[A]()

      @tailrec
      def queueLoop(queue: Queue[A]): Queue[A] = {
        if (queueLength <= toDrop) {
          queue
        } else {
            val (a, nextQueue) = queue.dequeue
            buffer.append(a)
            queueLength -= 1
          queueLoop(nextQueue)
        }
      }
      queueLoop(queue)

      val cursor = BatchCursor.fromSeq(buffer)
      NextCursor(cursor, F.pure(Halt(None)), F.unit)
    }

    def loop(toDrop: Int, length: Int, queue: Queue[A])
            (source: Iterant[F, A]): Iterant[F, A] = {
      try if (toDrop <= 0) source else source match {
        case Next(elem, rest, stop) =>
          val newQueue = queue.enqueue(elem)
          if (length >= toDrop){
            val (next, nextQueue) = newQueue.dequeue
            Next(next, rest.map(loop(toDrop, length, nextQueue)), stop)
          }
          else Suspend(rest.map(loop(toDrop, length + 1, newQueue)), stop)
        case ref@NextCursor(_, _, _) =>
          processCursor(toDrop, length, queue, ref)
        case NextBatch(batch, rest, stop) =>
          processCursor(toDrop, length, queue, NextCursor(batch.cursor(), rest, stop))
        case Suspend(rest, stop) =>
          Suspend(rest.map(loop(toDrop, length, queue)), stop)
        case Last(item) =>
          finalCursor(toDrop, length + 1, queue.enqueue(item))
        case Halt(None) =>
          finalCursor(toDrop, length, queue)
        case halt@Halt(Some(_)) =>
          halt
      } catch {
        case ex if NonFatal(ex) =>
          val stop = source.earlyStop
          Suspend(stop.map(_ => Halt(Some(ex))), stop)
      }
    }

    // Suspending execution, because pushing into our queue
    // is side-effecting
    Suspend(F.delay(loop(n, 0, Queue.empty[A])(source)), source.earlyStop)
  }
}
```

Avasil:
btw @alexandru I'm encountering weird behaviour in tests

If I add this to ```IterantDropLastSuite```:
```scala
  test("Iterant.dump works for NextCursor") { implicit s =>
    check2 { (list: List[Int], nr: Int) =>
      val stream = Iterant[Task].nextCursorS(BatchCursor(list: _*), Task.now(Iterant[Task].empty[Int]), Task.unit).onErrorIgnore
      val n = if (nr == 0) 0 else math.abs(math.abs(nr) % 20)

      val stream1 = stream.dropLast(n).toListL
      val stream2 = stream.toListL.map(_.dropRight(n))

      stream1.foreach(x => println(s"s1 $x"))
      stream2.foreach(x => println(s"s2 $x"))

      stream1 <-> stream2
    }
  }
```
it passes but ```println``` returns:
``` scala
s2 List()
s1 List(1297675174, 624554472, -905296189, 2147483647, 837085907)
s2 List()
s1 List(-761977668, -2047144408, -1, -2147483648, -1, -2147483648, 1, -2147483648, 0, -2147483648, 2147483647, 220996074, -308602378, 0, 1, 1, 0, -1, 1, 2137626787)
s2 List()
```
which should fail.
If I remove those ```foreach``` lines from tests - they fail.

alexandru:
@Avasil make sure you suspend execution for side effects, because if evaluating a stream once changes the outcome of its second evaluation it means you probably have unsuspected side effects

alexandru:
Speaking of which, a test like `stream <-> stream` would be good to guard against this.

Avasil:
That explains why second list is always empty, I will investigate in this direction.

Avasil:
@alexandru 
I have investigated it a bit more and it seems it is a bigger problem unless this is desired behavior for ```BatchCursor```.

Tests like that:
```scala
  test("Iterant.takeLast suspends side effects for NextCursor") { implicit s =>
    check2 { (list: List[Int], nr: Int) =>
      val n = if (nr == 0) 0 else math.abs(math.abs(nr) % 20)
      val stream = Iterant[Task].nextCursorS(BatchCursor(list: _*), Task.now(Iterant[Task].empty[Int]), Task.unit)
        .onErrorIgnore

      val s1 = stream.takeLast(n).toListL

      s1 <-> s1
    }
  }
```
fail not only for my ```Iterant.dropLast``` implementation but also ```Iterant.takeLast```, ```Iterant.map``` and probably everything else. Is it desired behavior?

alexandru:
Ah, wait, that's normal.

A `BatchCursor` is effectively an `Iterator` (I actually wanted to use `Iterator` directly, but I wanted strict semantics for arrays for efficiency, smaller API surface + `recommendedBatchSize`). Anyway, once you consume a `BatchCursor` / `Iterator`, then you're done. You can only consume it once. And the `NextCursor` state wraps it raw as it is.

To see that effects get suspended on this one, you have to basically check that the `.isEmpty`, `hasNext` or `next` where not called once you call your `dropLast`.

IMO a pattern match on the return of `NextCursor.dropLast` and `NextBatch.dropLast` to see that the returned state is `Suspend` should be sufficient.

A test like `stream <-> stream` is still useful though, but in that `check` you should auto-generate an `Iterant`:

```scala
check1 { stream: Iterant[Task, Int] => stream <-> stream }
```

And if by off chance our `Arbitrary[Iterant[Task]]` instance is returning `NextCursor` directly, then we need to stop it from doing that, since `NextBatch` will trigger `NextCursor` states anyway and we need to rely on referential transparency.

alexandru:
Btw, what did you decide about the mutable versus immutable `Queue`.
Are we going forward with the immutable one or not?

Avasil:
@alexandru I wasn't sure about this ```Iterant``` behavior and was worried I have buggy code.

According to benchmarks there is no rperformance eason to choose mutable ```Queue``` so I'll go with immutable one as you suggested. Unless I did some mistake in benchmark?

alexandru:
Going with the immutable one is the right approach.

Queues are special anyway. Mutable data structures have considerable performance benefits when backed by arrays, but the mutable queue you've used is probably not backed by arrays. An efficient implementation would be some sort of `ArrayList`. And in our case, given that the size of the queue used is fixed (`n + 1`), for performance you can do better by usage of a "circular buffer", which uses a fixed array internally, with no other heap allocations. 

If you'd like to play, we actually have one in [DropHeadOnOverflowQueue](https://github.com/monix/monix/blob/master/monix-execution/shared/src/main/scala/monix/execution/internal/collection/DropHeadOnOverflowQueue.scala) — it's a simple thing really, should be standard.

But I wouldn't sweat it, performance improvements should come as a result of benchmarking, which is freaking annoying and I'm perfectly happy with an implementation based on an immutable `Queue` right now.

I'm leaving the choice to you btw, you're the owner 😀

Avasil:
@alexandru I'd love to play with something like that, even if just to learn something new but IMO it is better for the project If I focus my "resources" on helping with ```3.0.0``` release which seems pretty close seeing how many people came to contribute!

alexandru:
Thanks @Avasil, looking good.

PR #510, opened 2018-01-10
Add Iterant.takeEveryNth operation

### LINKED ISSUE #500: Add Iterant.takeEveryNth
Need a new `Iterant` op that that emits every n-th event from the source, dropping intermediary events.

```scala
  // ...
  def takeEveryNth(n: Int)(implicit F: Sync[F]): Iterant[F, A]
}
```

See the [observable implementation](https://github.com/monix/monix/blob/v3.0.0-M3/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/TakeEveryNthOperator.scala) for inspiration.

greenhat:
Closes #500

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantTakeEveryNth.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantTakeEveryNth.scala
@@ -1,0 +1,95 @@
+/*
+ * Copyright (c) 2014-2018 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.tail.internal
+
+import monix.tail.Iterant.{Halt, Last, Next, NextBatch, NextCursor, Suspend}
+import monix.tail.Iterant
+import cats.syntax.all._
+import cats.effect.Sync
+import monix.execution.misc.NonFatal
+import monix.tail.batches.BatchCursor
+import monix.tail.internal.IterantUtils._
+
+import scala.collection.mutable.ArrayBuffer
+
+private[tail] object IterantTakeEveryNth {
+  /**
+    * Implementation for `Iterant#takeEveryNth`
+    */
+  def apply[F[_], A](source: Iterant[F, A], n: Int)
+                    (implicit F: Sync[F]): Iterant[F, A] = {
+
+    require(n > 0, "n must be strictly positive")
+
+    // TODO shortcut for n = 1
+
+    def processSeq(index: Int, ref: NextCursor[F, A]): NextCursor[F, A] = {
+      val NextCursor(cursor, rest, stop) = ref
+      val buffer = ArrayBuffer.empty[A]
+
+      var idx = index
+      var toProcess = cursor.recommendedBatchSize
+
+      // protects against infinite cursors
+      while (toProcess > 0 && cursor.hasNext()) {
+        if (idx == 1) {
+          buffer += cursor.next()
+          idx = n
+        } else {
+          cursor.next()
+          idx -= 1
+        }
+        toProcess -= 1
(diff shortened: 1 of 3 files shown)


monix-tail/shared/src/test/scala/monix/tail/IterantTakeEveryNthSuite.scala
```diff
@@ -0,0 +1,121 @@
...
+    val stop = Coeval.eval(effect += 1)
+    val source = Iterant[Coeval].nextCursorS(BatchCursor(1,2,3), Coeval.now(Iterant[Coeval].empty[Int]), stop)
+    val stream = source.takeEveryNth(1)
+    stream.earlyStop.value
+    assertEquals(effect, 1)
+  }
+
+  test("Iterant.takeEveryNth throws on invalid n") { implicit s =>
+    val source = Iterant[Coeval].nextCursorS(BatchCursor(1,2,3), Coeval.now(Iterant[Coeval].empty[Int]), Coeval.unit)
+    intercept[IllegalArgumentException] { source.takeEveryNth(0).toListL.value }
```
alexandru: Just a note — we need to be precise about what throws the error. Given you have a `require` in `takeEveryNth`, then you should delete `.toListL.value`.

monix-tail/shared/src/test/scala/monix/tail/IterantTakeEveryNthSuite.scala
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  test("Iterant.takeEveryNth triggers early stop on exception") { _ =>
+    check1 { (iter: Iterant[Coeval, Int]) =>
+      val cancelable = BooleanCancelable()
+      val dummy = DummyException("dummy")
+      val suffix = Iterant[Coeval].nextCursorS[Int](new ThrowExceptionCursor(dummy), Coeval.now(Iterant[Coeval].empty), Coeval.unit)
+      val stream = (iter.onErrorIgnore ++ suffix).doOnEarlyStop(Coeval.eval(cancelable.cancel()))
+      intercept[DummyException] { stream.takeEveryNth(1).toListL.value }
```
alexandru: Here you need to be precise about how the error gets thrown too. So the test should be:

```scala
assertEquals(stream.takeEveryNth(1).toListL.runTry, Failure(dummy))
```

greenhat:
@alexandru Thanks! Fixed the places where the exception is expected. 
I'm done with implementation and I'm out of ideas for more tests. Let me know if I miss something.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantTakeEveryNth.scala
```diff
@@ -0,0 +1,95 @@
...
+private[tail] object IterantTakeEveryNth {
+  /**
+    * Implementation for `Iterant#takeEveryNth`
+    */
+  def apply[F[_], A](source: Iterant[F, A], n: Int)
+                    (implicit F: Sync[F]): Iterant[F, A] = {
+
+    require(n > 0, "n must be strictly positive")
+
+    if (n == 1) return source
```
alexandru: I don't like usage of `return`, unless it's made for concrete optimization reasons, or in other words it's very low level stuff.

This check should be included before the `source match` below.

greenhat: @alexandru Totally agree. Just could not find a better place for it. Fixed.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantTakeEveryNth.scala
```diff
@@ -0,0 +1,94 @@
...
+import monix.tail.internal.IterantUtils._
+
+import scala.collection.mutable.ArrayBuffer
+
+private[tail] object IterantTakeEveryNth {
+  /**
+    * Implementation for `Iterant#takeEveryNth`
+    */
+  def apply[F[_], A](source: Iterant[F, A], n: Int)
+                    (implicit F: Sync[F]): Iterant[F, A] = {
```
alexandru: Don't like this indentation style and the Monix code is pretty consistent on not doing things dangling to the right like that 😀 personal preference, but it's good to be consistent.

Put this signature all on one line, delete the empty line between it and `require` below and you're done 😀

alexandru:
This looks good @greenhat, thanks, one last comment about indentation and you're done.

Will then merge as soon as the build is green.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantTakeEveryNth.scala
greenhat: Consistency is the king! Fixed. It's my first PR here, need to learn codebase more. Thanks for the fast and thorough review!

PR #512, opened 2018-01-11
Add Iterant.dropWhileWithIndex

### LINKED ISSUE #496: Add Iterant.dropWhileWithIndex
Similar with the `Observable` equivalent, drops the longest prefix of elements that satisfy the given predicate:

```scala
  // ...
  def dropWhileWithIndex(p: (A, Int) => Boolean)(implicit F: Sync[F]): Iterant[F, A]
}
```

For inspiration see the implementation of [DropByPredicateWithIndexOperator](https://github.com/monix/monix/blob/v3.0.0-M3/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DropByPredicateWithIndexOperator.scala), which is for `Observable`.

Avasil:
Fixes #496.

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropWhileWithIndex.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropWhileWithIndex.scala
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2014-2018 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.tail.internal
+
+import cats.effect.Sync
+import cats.syntax.all._
+import monix.execution.misc.NonFatal
+import monix.tail.Iterant
+import monix.tail.Iterant.{Halt, Last, Next, NextBatch, NextCursor, Suspend}
+import monix.tail.batches.BatchCursor
+import scala.annotation.tailrec
+
+private[tail] object IterantDropWhileWithIndex {
+  /**
+    * Implementation for `Iterant#dropWhileWithIndex`
+    */
+  def apply[F[_], A](source: Iterant[F, A], p: (A, Int) => Boolean)
+                    (implicit F: Sync[F]): Iterant[F, A] = {
+
+    // Reusable logic for NextCursor / NextBatch branches
+    @tailrec
+    def evalCursor(ref: F[Iterant[F, A]], cursor: BatchCursor[A], rest: F[Iterant[F, A]],
+                   stop: F[Unit], dropped: Int, index: Int, continueDropping: Boolean): Iterant[F, A] = {
+      if (!cursor.hasNext())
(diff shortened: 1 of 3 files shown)


monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropWhileWithIndex.scala
```diff
@@ -0,0 +1,93 @@
...
+  /**
+    * Implementation for `Iterant#dropWhileWithIndex`
+    */
+  def apply[F[_], A](source: Iterant[F, A], p: (A, Int) => Boolean)
+                    (implicit F: Sync[F]): Iterant[F, A] = {
+
+    // Reusable logic for NextCursor / NextBatch branches
+    @tailrec
+    def evalCursor(ref: F[Iterant[F, A]], cursor: BatchCursor[A], rest: F[Iterant[F, A]],
+                   stop: F[Unit], dropped: Int, index: Int, continueDropping: Boolean): Iterant[F, A] = {
```
alexandru: Personal pet peeve, but I don't like things dangling like that to the right and the codebase is pretty consistent in that regard. This would be better:

```scala
    def evalCursor(
      ref: F[Iterant[F, A]], cursor: BatchCursor[A], rest: F[Iterant[F, A]],
      stop: F[Unit], dropped: Int, index: Int, continueDropping: Boolean): Iterant[F, A] = { 

      // ....
```

monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropWhileWithIndex.scala
```diff
@@ -0,0 +1,93 @@
...
+        else
+          Next(elem, rest.map(loop(index, continueDropping = false)), stop)
+      }
+    }
+
+    def loop(index: Int, continueDropping: Boolean)(source: Iterant[F, A]): Iterant[F, A] = {
+      try source match {
+        case ref@Next(item, rest, stop) =>
+          if (continueDropping && p(item, index)) Suspend(rest.map(loop(index + 1, continueDropping = true)), stop)
+          else ref
```
alexandru: Another personal pet peeve — that line is too long, so something like this maybe?

```scala
if (continueDropping && p(item, index)) 
  Suspend(rest.map(loop(index + 1, continueDropping = true)), stop)
else 
  ref 
```

monix-tail/shared/src/main/scala/monix/tail/internal/IterantDropWhileWithIndex.scala
```diff
@@ -0,0 +1,93 @@
...
+      else if (dropped >= cursor.recommendedBatchSize)
+        Suspend(ref.map(loop(index, continueDropping)), stop)
+      else {
+        val elem = cursor.next()
+        if (continueDropping && p(elem, index))
+          evalCursor(ref, cursor, rest, stop, dropped + 1, index + 1, continueDropping)
+        else if (cursor.hasNext())
+          Next(elem, ref.map(loop(index, continueDropping = false)), stop)
+        else
+          Next(elem, rest.map(loop(index, continueDropping = false)), stop)
```
alexandru: So, all of this looks good, however I believe you can simplify it a little - correct me if I'm wrong ...

This law applies:

```scala
ref.map(loop(index, continueDropping = false)) <-> ref
```

Once `continueDropping` becomes `false` the predicate `p` is no longer called and you no longer drop any events, correct? This means that you can just return `ref` or `rest` without the `.map(loop(...))`.

This also means that you can get rid of `continueDropping` completely. You assume that you have to drop all elements, until the predicate becomes `false`, at which point you just return whatever you have left, but without the `.map(loop(...))`.

alexandru:
Hey @Avasil, sorry about the style-related comments — one of these days I'll configure a code formatting plugin. I resisted until now due to code formatting plugins not doing a very good job, plus I disagree with some standard practices 😀

But until then I have to annoy people in code reviews.

Avasil:
@alexandru No problem, code formatting could be nice because despite doing my best to follow codebase style I don't always get it right and one instinctive ```ctrl + alt + l``` in IntelliJ can ruin it :D

PR #514, opened 2018-01-11
Add Iterant.takeWhileWithIndex

### LINKED ISSUE #497: Add Iterant.takeWhileWithIndex
Similar with [issue #496](https://github.com/monix/monix/issues/496) we need a `takeWhileWithIndex` on `Iterant` that takes elements for as long as the given predicate returns `true`:

```scala
  // ...
  def takeWhileWithIndex(p: (A, Int) => Boolean)(implicit F: Sync[F]): Iterant[F, A]
}
```

The implementation should be very similar actually.

greenhat:
Fixes #497.

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
@@ -1392,6 +1392,24 @@ sealed abstract class Iterant[F[_], A] extends Product with Serializable {
   final def takeWhile(p: A => Boolean)(implicit F: Sync[F]): Iterant[F, A] =
     IterantTakeWhile(self, p)(F)
 
+  /** Takes longest prefix of elements zipped with their indices that satisfy the given predicate
+    * and returns a new iterant that emits those elements.
+    *
+    * Example: {{{
+    *   // Yields 1, 2
+    *   Iterant[Task].of(1, 2, 3, 4, 5, 6).takeWhile((_, idx) => idx < 2)
+    * }}}
+    *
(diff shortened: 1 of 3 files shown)


greenhat:
I'm done. Let me know if I missed something.

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -1392,6 +1392,24 @@ sealed abstract class Iterant[F[_], A] extends Product with Serializable {
   final def takeWhile(p: A => Boolean)(implicit F: Sync[F]): Iterant[F, A] =
     IterantTakeWhile(self, p)(F)
 
+  /** Takes longest prefix of elements zipped with their indices that satisfy the given predicate
+    * and returns a new iterant that emits those elements.
+    *
+    * Example: {{{
+    *   // Yields 1, 2
+    *   Iterant[Task].of(1, 2, 3, 4, 5, 6).takeWhile((_, idx) => idx != 2)
```
alexandru: An, just one very small issue with this sample, you got the name wrong: `takeWhile` => `takeWhileWithIndex`

😀

greenhat: @alexandru Thanks for spotting this one, Alexandru! I was so obsessed to come up with a good example that I missed it. Fixed.

alexandru:
@greenhat I forgot; please add yourself to the `AUTHORS` file. It doesn't have to be in this PR — I'll merge this one once it turns green to get it out of the way.

PR #518, opened 2018-01-12
Add Iterant[Task].intervalAtFixedRate builder

### LINKED ISSUE #516: Add Iterant[Task].intervalAtFixedRate
For `Observable` we have:

No 1 — `Observable.intervalWithFixedDelay` — generates ticks with a fixed delay between them, such that when you have something like this ...

```scala
Observable.intervalWithFixedDelay(5.seconds).mapTask { _ =>
  Task.eval(println("tick")).delayExecution(3.seconds)
}
```

This logic will print `"tick"` every 8 seconds, because the source gets back-pressured and the delay is fixed.

No 2 — `Observable.intervalAtFixedRate` — generates ticks at a fixed rate, like the above but *substracts* the time it took for the consumer to consume the tasks, trying to keep the spacing between the starts fixed:

```scala
Observable.intervalAtFixedRate(5.seconds).mapTask { _ =>
  Task.eval(println("tick")).delayExecution(3.seconds)
}
```

This piece of logic will print `"tick"` every 5 seconds, because the 3 seconds it takes for those tasks to execute is taken into account and subtracted from the 5 seconds period.

## Your Task, Should You Choose to Accept It

We already have [Iterant.intervalWithFixedDelay](https://github.com/monix/monix/blob/v3.0.0-M3/monix-tail/shared/src/main/scala/monix/tail/internal/IterantIntervalWithFixedDelay.scala) implementated.

We need `intervalAtFixedRate` with the same contract as for `Observable`. 

Lets note the duration of the downstream tasks (the `rest` task at each step) with `taskDuration`.
The contract is:

1. if `taskDuration <= period` then the delay is `period - taskDuration`
2. if `taskDuration > period` then the delay is zero
3. obviously two tasks cannot run in parallel, so the actual tick can take longer than the specified period, but that's fine, it's the same contract as on `Observable`

You can also checkout [IntervalFixedRateObservable](https://github.com/monix/monix/blob/v3.0.0-M3/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/IntervalFixedRateObservable.scala) for inspiration.

## Other Considerations

The implementation is specialized for `Task` because with `Task` we can express `.delayExecution` without any `ExecutionContext` or `Scheduler` needed in scope.

This is OK for now, subject to change in the future.

Good luck 😀

oleg-py:
Closes #516

I hope I got this right.


alexandru:
Looks good 👍

PR #523, opened 2018-01-13
Add Iterant.scanMap

### LINKED ISSUE #519: Add Iterant.scanMap
As described in #449 for `Observable`, we also need a new `Iterant` op:

```scala
def scanMap[B](f: A => B)(implicit F: Sync[F], B: Monoid[B]): Iterant[F, B]
```

Note this can be expressed in terms of the existing `scan`.

Avasil:
Fixes #519.


monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -1656,6 +1656,51 @@ sealed abstract class Iterant[F[_], A] extends Product with Serializable {
   final def scanEval[S](seed: F[S])(op: (S, A) => F[S])(implicit F: Sync[F]): Iterant[F, S] =
     IterantScanEval(self, seed, op)
 
+  /** Applies implicitly available `Monoid[B]` to combine elements using
+    * `Monoid[B]` identity value as the initial state and map every incoming
+    * element of this `Iterant` with provided function f: A => B.
```
alexandru: There's no "identity value" in Monoid, that value is called "empty" or "zero" (depending on the implementation).

Also the provided example defines a `Monoid` instance on the fly. That's not how this function will get used because people don't define `Monoid` instances on the fly like that — usually `Monoid` instances are attached to data types, with tests integrated with `cats-laws`, etc. What makes a `Monoid` a `Monoid` isn't only the interface, but the laws as well, which makes monoid instances expensive. So users that don't have a `Monoid[B]` instance already will probably not use this function.

I would say something like this:
```
Given a mapping function that returns a `B` type for which we have 
a [[cats.Monoid]] instance, returns a new stream that folds the incoming
elements of the sources using the provided `Monoid.combine`, with the
initial seed being the `Monoid.empty` value, emitting the generated values
at each step.

Equivalent with [[scan]] applied with the given [[cats.Monoid]], so given 
our `f` mapping function returns a `B`, this law holds:
{{{
  val B = implicitly[Monoid[B]]

  stream.scanMap(f) <-> stream.scan(B.empty)(B.combine)
}}}

Example:
{{{
  // Yields 2, 6, 12, 20, 30, 42
  Iterant[Task].of(1, 2, 3, 4, 5, 6).scanMap(x => x * 2)
}}}
```

Yeah, docs are harder than code 😛

monix-tail/shared/src/test/scala/monix/tail/IterantScanMapSuite.scala
```diff
@@ -0,0 +1,93 @@
...
+        .collect { case Current(a, _) => a }
+
+      fa.toListL <-> source.take(10).toListL.map(_.take(9))
+    }
+  }
+
+  test("Iterant.scanMap equivalence to Iterant.scan") { implicit s =>
+    check1 { (source: Iterant[Coeval, Int]) =>
+
+      val scanned1 = source.scanMap(Current(_, 1): State[Int])
```
alexandru: I also think these tests are too complicated, because we just defer to `scan`.

So the only test we need is this `scanMap equivalence to scan`  and IMO you can just operate with `Int` instead of defining that `State`.

`scan` needs more complicated tests like this one, but not `scanMap`.

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
Avasil: Regarding "identity value" - I see it often (maybe more in Haskell) but good point anyway because it might only confuse people. 

I've changed docs according to your suggestion. :)

Avasil:
@alexandru Done!

PR #524, opened 2018-01-14
Add Iterant.interleave

### LINKED ISSUE #498: Add Iterant.interleave
Need an `Iterant` op that interleaves the elements of 2 streams into a strictly alternating sequence.

For `Iterant` it should be similar with `zip`, an operation that proved a little challenging but doable, see [IterantZipMap](https://github.com/monix/monix/blob/v3.0.0-M3/monix-tail/shared/src/main/scala/monix/tail/internal/IterantZipMap.scala).

```scala
  def interleave[B](rhs: Iterant[F, A])(implicit F: Sync[F]): Iterant[F, A]
}
```

greenhat:
Closes #498.

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantInterleave.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantInterleave.scala
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2014-2018 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.tail.internal
+
+import cats.effect.Sync
+import monix.tail.Iterant
+
+
+private[tail] object IterantInterleave {
+  def apply[F[_], A, B >: A](lh: Iterant[F, A], rh: Iterant[F, B]) (implicit F: Sync[F]): Iterant[F, B] = {
+    rh
+  }
+}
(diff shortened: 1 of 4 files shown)


monix-tail/shared/src/main/scala/monix/tail/internal/IterantInterleave.scala
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package monix.tail.internal
+
+import cats.effect.Sync
+import monix.tail.Iterant
+
+
+private[tail] object IterantInterleave {
+  def apply[F[_], A, B >: A](lh: Iterant[F, A], rh: Iterant[F, B]) (implicit F: Sync[F]): Iterant[F, B] = {
```
alexandru: Just a small comment — you don't need the subtyping relationship to avoid covariance issues here, i.e. you don't need a `B` type parameter.

alexandru: Or in other words, it's OK to have it on the method in `Iterant`, but not on this function here.

greenhat: @alexandru If I remove `B` in `apply`:
```
def apply[F[_], A](lh: Iterant[F, A], rh: Iterant[F, A]) (implicit F: Sync[F]): Iterant[F, A]
```
I get error:
```
[error] monix/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala:1528:23: type mismatch;
[error]  found   : monix.tail.Iterant[F,A]
[error]  required: monix.tail.Iterant[F,B]
[error] Note: A <: B, but class Iterant is invariant in type A.
[error] You may wish to define A as +A instead. (SLS 4.5)
[error]     IterantInterleave(self, rhs)(F)
[error]                       ^
```
Do you want to remove `B` from `Iterant.interleave`?

alexandru: Ah, true, `Iterant` is invariant for now as a design decision having to do with covariance not mixing well with higher kinds, but you can use `self.upcast[B]` in that call.

greenhat: @alexandru Thanks! Done. I saw `upcast` but did not think of using it here.

greenhat:
@alexandru I'm done.

alexandru:
Thanks @greenhat, will review this soon — it has many branches and I have to pay attention 😀 tonight or tomorrow.

alexandru:
Looking good @greenhat, merged!

PR #530, opened 2018-01-15
Add Task.forkAndForget

### LINKED ISSUE #520: Add Task.forkAndForget
For `Task` we now have `.start` and `.fork` with the same semantics, except that `fork` also forces a logical thread fork:

```scala
sealed abstract class Task[+A] {
  // ...
  def start: Task[Task[A]]
  // ...
  def fork: Task[Task[A]] =
    executeAsync.start
}
```

Now we need something like this:

```scala
sealed abstract class Task[+A] {
  // ...
  def forkAndForget: Task[Unit]
}
```

Following law should apply, but the implementation doesn't need to be this:

```scala
task.forkAndForget <-> task.fork.map(_ => ())
```

N.B. this law probably can't be tested since `forkAndForget` is all about triggering side effects.

Avasil:
Fixes #520.
Implementation is very simple, it seems to work but I'm of course open to any suggestions improving it and I think ScalaDoc could use meaningful example which could potentially be also used in one of the tests.


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1092,6 +1092,21 @@ sealed abstract class Task[+A] extends Serializable {
...
+    * running it in the background and discarding the result.
+    *
+    * Similar to [[start]] after mapping result to Unit. Below law holds:
+    *
+    * {{{
+    *   task.startAndForget <-> task.start.map(_ => ())
+    * }}}
+    *
+    * IMPORTANT - In contrast to [[start]] it does automatic fork,
+    * forcing new async boundary.
```
alexandru: Once you refer to `fork` instead of `start`, this comment is not needed.

monix-eval/shared/src/main/scala/monix/eval/internal/TaskStartAndForget.scala
```diff
@@ -0,0 +1,35 @@
...
+    */
+  def apply[A](fa: Task[A]): Task[Unit] =
+    Task.Async[Unit] { (ctx, cb) =>
+      implicit val sc = ctx.scheduler
+      // It needs its own context, its own cancelable
+      val ctx2 = Task.Context(sc, ctx.options)
+      // Starting actual execution of our newly created task forcing new async boundary
+      Task.unsafeStartAsync(fa, ctx2, Callback.empty)
+      // Signal the created Task reference
+      cb.onSuccess(())
```
alexandru: This has the potential of generating stack overflow errors under certain conditions, which is why as policy we never call `cb.onSuccess(())` straight from `Async`.

Please use `cb.asyncOnSuccess(())`.

monix-eval/shared/src/test/scala/monix/eval/TaskStartAndForgetSuite.scala
```diff
@@ -0,0 +1,56 @@
...
+      sc.tick()
+
+      counter <-> i
+    }
+  }
+
+  test("task.startAndForget is stack safe") { implicit sc =>
+    var task: Task[Any] = Task(1)
+    for (_ <- 0 until 5000) task = task.startAndForget
+    for (_ <- 0 until 5000) task = task.flatMap(_ => Task.unit)
```
alexandru: Use `5000` for JavaScript, `100000` for Java. You can differentiate by using `Platform.isJVM` (from `monix.execution.internal`).

monix-eval/shared/src/test/scala/monix/eval/TaskStartAndForgetSuite.scala
```diff
@@ -0,0 +1,56 @@
...
+
+    val f = task.runAsync
+    sc.tick()
+    assertEquals(f.value, Some(Success(())))
+  }
+
+  test("task.startAndForget <-> task.start.map(_ => ())") { implicit sc =>
+    check1 { (task: Task[Int]) =>
+      task.startAndForget <-> task.start.map(_ => ())
```
alexandru: `startAndForget` on evaluation triggers side effects that we cannot notice here; this test is basically a no-op because you could've said:

```scala
task.startAndForget <-> Task.unit
```

We don't need it. Get rid of it.

monix-eval/shared/src/main/scala/monix/eval/internal/TaskStartAndForget.scala
oleg-py: Nitpick: the comment above is not correct, there's no Task reference signalled

monix-eval/shared/src/test/scala/monix/eval/TaskStartAndForgetSuite.scala
```diff
@@ -0,0 +1,56 @@
...
+    check1 { (i: Int) =>
+      var counter = 0
+      val task = Task {
+        counter = i
+        i
+      }
+      task.startAndForget.runAsync
+      sc.tick()
+
+      counter <-> i
```
alexandru: You're using `ScalaCheck` to repeat the test for random `i: Int` values, however that won't do much for us.

Do a simple test and test that:

1. async boundary is needed
2. execution is actually triggered in a background process (its own thread)

```scala
val task = Task.eval { counter += 1; counter }
val task = for {
  _ <- task.delayExecution(1.second).forkAndForget // first
  _ <- task.delayExecution(1.second).forkAndForget // second
} yield ()

val f = task.runAsync
assertEquals(f.value, Some(Success(()))

// Testing that both got executed in parallel
s.tick(1.second)
assertEquals(counter, 2)
```

Well, this doesn't test that an async boundary is forced, since we are doing it anyway with `delayExecution`, but you get the idea.

monix-eval/shared/src/test/scala/monix/eval/TaskStartAndForgetSuite.scala
Avasil: Can I somehow prove this equivalence?

alexandru: I don't think so. Would be nice but `forget` is in the name 😀 and I can't think of anything useful to do here.

Avasil:
@alexandru Thanks, I've used your example and added similar test but with raising exception. Maybe after sleep I will come up with a way to test for async boundary, it seems like other tests use either ```delayExecution``` or executing on different ```Scheduler```

monix-eval/shared/src/main/scala/monix/eval/internal/TaskForkAndForget.scala
```diff
@@ -0,0 +1,34 @@
...
+    * Implementation for `Task.startAndForget`.
+    */
+  def apply[A](fa: Task[A]): Task[Unit] =
+    Task.Async[Unit] { (ctx, cb) =>
+      implicit val sc = ctx.scheduler
+      // It needs its own context, its own cancelable
+      val ctx2 = Task.Context(sc, ctx.options)
+      // Starting actual execution of our newly created task forcing new async boundary
+      Task.unsafeStartAsync(fa, ctx2, Callback.empty)
+      cb.asyncOnSuccess(())
```
alexandru: Implementation looks good 👍

monix-eval/shared/src/test/scala/monix/eval/TaskForkAndForgetSuite.scala
```diff
@@ -0,0 +1,71 @@
...
+    val errorTask = Task.raiseError(new DummyException)
+
+    val result = for {
+      _ <- errorTask.forkAndForget
+      value <- task
+    } yield value
+
+    val f = result.runAsync
+    sc.tick()
+    assertEquals(f.value, Some(Success(20)))
```
alexandru: You're raising the error in that background thread, however you now need to ensure that the error was reported. Something like:

```scala
assertEquals(sc.state.lastReportedError, dummy)
```

alexandru: This happens due to `Callback.empty` if you were wondering, which isn't so "empty" 😀 as it reports incoming errors with the provided `Scheduler`.

Issue #513, opened 2018-01-11
Setup automatic code formatting in SBT


alexandru:
Style is difficult to maintain and a burden in code reviews.

I long maintained that formatting should be done by the developers themselves, due to available tools not understanding context well. But for public projects that receive contributions this does not scale.

My preference would be for [scalafmt](http://scalameta.org/scalafmt/), but it remains to be seen if it can be configured well.

Settings need to be adjusted to match the current style as much as possible.
Some quick rules, probably more to come as the need arises:

1. Automatic copyright header
2. ScalaDoc formatting should be disabled, being completely the responsibility of the developer
3. In general I don't like dangling things to the right and I hate what IntelliJ IDEA does by default with function and class constructor params
4. one line branches in `match` statements are OK, but the style needs to be consistent for all branches of an expression (`if/else`, `match`, etc), so either all branches are one lines, or all of them introduce newlines between condition and expression

## No. 1 — auto copyright header

Being careful about the copyright header is a burden I'd like to avoid.

## No. 2 — ScalaDoc

Monix uses carefully crafted indentation that does not match the formatting rules of IntellIJ IDEA and I'd like to avoid any fuckups of an automated plugin. 

ScalaDoc is documentation and documentation requires manual crafting.

## No. 3 — no dangling stuff to the right

For classes, this is OK:

```scala
case class Foo(
  number: Int,
  name: String)

case class Bar(
  number: Int,
  name: String
)
```

This is not OK:

```scala
case class Foo(number: Int,
               name: String)
```

For functions, this is OK:

```scala
def foo(number: Int,
  name: String): Result = {
  ???
}

def bar(
  number: Int,
  name: String): Result = {

  ???
}

// At call site
foo(
  number,
  name)
```

But this is not OK:

```scala
def foo(number: Int
        name: String): Result = {
  ???
}

// At call site
foo(number
    name)
```

So there's a strict policy against *dangling stuff to the right*.

## No. 4 — indentation of match / if

For `if/else`, this is OK:

```scala
if (condition) success
else failure

if (condition) success else {
  failure
}

if (condition)
  success
else
  failure
```

This is not OK:

```scala
if (condition) success
else {
  failure
}

if (condition) {
  success
} else failure
```

For `match`, this is OK:

```scala
// All branches are one lines
expr match {
  case Foo(_) => success
  case Bar(_) => failure
}

expr match {
  case Foo(_) => 
    success
  case Bar(_) => 
    failure
}

[...]

PR #531, opened 2018-01-15
Add automatic header creation on compilation

oleg-py:
First step to finishing #513.

There's [sbt-header](https://github.com/sbt/sbt-header) just for making headers on files, so I decided to put this together.

I ran the `headerCreate` command on whole project and it only touched few files where the header was outdated. I also set the project in the way that doing `sbt compile` will automatically create headers, so the contributors are not bothered by it. Alternatively, it could be made so that users have to run command manually and CI fails build if header is not correct.

Of other formatting tools, it seems that only Scalastyle has the [header check](http://www.scalastyle.org/rules-1.0.0.html#org_scalastyle_file_HeaderMatchesChecker), but I haven't looked into formatting tool yet, so I decided against committing to Scalastyle for now.


alexandru:
Oh, nice, thanks.

dwijnand:
Btw sbt-header has `headerCheck` which you can invoke in CI to enforce headers.

https://github.com/sbt/sbt-header/tree/v4.1.0#checking-headers

PR #536, opened 2018-01-16
Add a Parallel for Observable for using combineLatest

### LINKED ISSUE #459: Add a Parallel for Observable for using combineLatest
`combineLatestMap` is a super useful combinator! Using it with more than a couple of combinators can be quite cumbersome though, so something like a `Parallel` would be a super useful thing IMO.
Here's some code I wrote:

```scala
class CombineObservable[A](val value: Observable[A]) extends AnyVal
object CombineObservable {
  implicit def combineObservableApplicative: Applicative[CombineObservable] = new Applicative[CombineObservable] {
    def pure[A](x: A): CombineObservable[A] = new CombineObservable(Observable.now(x))

    def ap[A, B](ff: CombineObservable[(A) => B])(fa: CombineObservable[A]) = new CombineObservable(
      ff.value.combineLatestWith(fa.value)((f, a) => f(a))
    )

    override def map[A, B](fa: CombineObservable[A])(f: A => B): CombineObservable[B] =
      new CombineObservable(fa.value.map(f))

    override def product[A, B](fa: CombineObservable[A], fb: CombineObservable[B]): CombineObservable[(A, B)] =
      new CombineObservable(fa.value.combineLatest(fb.value))
  }
}
```

And then inside `object Observable`:
```scala
implicit def observableParallel: Parallel[Observable, CombineObservable] = new Parallel[Observable, CombineObservable] {
  def monad = observableMonad
  def applicative = CombineObservable.combineObservableApplicative
  def sequential = new (CombineObservable ~> Observable) {
    def apply[A](fa: CombineObservable[A]) = fa.value
  }
  def parallel = new (Observable ~> CombineObservable) {
    def apply[A](fa: Observable[A]) = new CombineObservable(fa)
  }
}
``` 

I've checked the laws for my `rxjs` wrapper and they worked fine! However if we find it's not a valid `Applicative`, but only a valid `Apply` we could still make a `NonEmptyParallel` which would still allow us most of the benefits (i.e. `parMapN` syntax)

Avasil:
Closes #459

Thanks to @LukaJCB for most of the implementation. :)


Avasil:
Right now ```CombineObservable``` fails ```ApplicativeTests``` but I'll see if I can make it work and in worst case we'll do with ```Apply``` as @LukaJCB suggested

alexandru:
What laws are failing?

Avasil:
```scala
Applicative[CombineObservable].applicative.applicative interchange
```

``` scala
fa: List(OnNext(1), OnComplete)
fb: List(OnNext(-1152585554), OnNext(1), OnComplete)
```
and

```scala
Applicative[CombineObservable].applicative.monoidal right identity
```
```scala
fa: List(OnNext(964858958), OnComplete)
fb: List(OnNext(-1), OnNext(964858958), OnComplete)
```

Avasil:
As for right identity:
```scala
    val obs = new CombineObservable(Observable(0, 1))
    val F = implicitly[Applicative[CombineObservable]]
    val x = (F.product(obs, F.pure(())), obs)
    val y = F.imap(x._1) { case (a, _) => a } { a => (a, ()) }

    // List(1)
    val lhs = y.value.toListL.runAsync
    // List(0, 1)
    val rhs = obs.value.toListL.runAsync
    // List((1, ()))
    val product = F.product(obs, F.pure(())).value.toListL.runAsync
```

I don't think it can pass because it compares entire ```Observable``` to result of ```combineLatest``` which discards unpaired elements and will take only the latest.

Avasil:
On the other hand, I feel like ```CombinedObservable``` should only take the most recent element from ```Observable``` so it might work this way

Avasil:
@alexandru The best thing I came up with so far is:
```scala
final class CombineObservable[A] private (val value: Observable[A], val latestValue: Observable[A])

object CombineObservable {
  def apply[A](value: Observable[A]): CombineObservable[A] = {
    new CombineObservable(value, value.combineLatestMap(Observable.now(()))((a, _) => a))
  }

  implicit def combineObservableApplicative: Applicative[CombineObservable] = new Applicative[CombineObservable] {
    def pure[A](x: A): CombineObservable[A] = new CombineObservable(Observable.now(x), Observable.now(x))

    def ap[A, B](ff: CombineObservable[(A) => B])(fa: CombineObservable[A]) = CombineObservable(
      ff.value.combineLatestMap(fa.value)((f, a) => f(a))
    )

    override def map[A, B](fa: CombineObservable[A])(f: A => B): CombineObservable[B] =
      CombineObservable(fa.value.map(f))

    override def product[A, B](fa: CombineObservable[A], fb: CombineObservable[B]): CombineObservable[(A, B)] = {
      CombineObservable(fa.value.combineLatest(fb.value))
    }
  }
}
```
there might be a better way to take the latest element but that's the general idea. This way all ```Applicative``` laws pass but I feel like it is too much of a hack. :P Without storing additional value ```NonEmptyParallelLaws.parallelRoundTrip``` check is failing which is:
```scala
def parallelRoundTrip[A](ma: M[A]): IsEq[M[A]] =
  P.sequential(P.parallel(ma)) <-> ma
// translates to
CombineObservable(ma).value <-> ma
```

alexandru:
`combineLatest` is very non-deterministic.  Resembles `parMap2` on `Task`, but there only the effects can come in a different order, yet the result will always be the same, whereas for `combineLatest` this law does not hold for any two streams with a size greater than one:

```scala
combineLatestMap(oa, ob) <-> combineLatestMap(oa, ob)
```

The operation that produces deterministic results in this case is actually `zip`, which back-pressures one stream until the other emits a value. But then `zip` isn't as useful.

```scala
final class CombineObservable[A] private (val value: Observable[A], val latestValue: Observable[A])
```

We can't do this, no, as this is no longer a `newtype`.

Try to see if it passes the laws of `ApplyLaws` and `NonEmptyParallel`.
If it doesn't, then there's nothing that we can do.

Avasil:
@alexandru ```NonEmptyParallel``` passes without any issues

LukaJCB:
:+1: Thanks for this!

monix-reactive/shared/src/main/scala/monix/reactive/observables/CombineObservable.scala
```diff
@@ -0,0 +1,42 @@
...
+package monix.reactive.observables
+
+import cats.Apply
+import monix.reactive.Observable
+
+/** A `CombineObservable` is an observable that wraps a regular
+  * [[Observable]] and provide [[cats.Apply]] instance
+  * which uses [[Observable.combineLatest]] to combine elements.
+  */
+final class CombineObservable[A](val value: Observable[A]) extends AnyVal
```
alexandru: This is OK-ish, but I'd like to avoid any boxing, plus I think we should be standardizing on a newtype encoding in the Cats ecosystem.

Please use the newtype encoding that I used for `IO.Par` in this PR: https://github.com/typelevel/cats-effect/pull/115/

alexandru: To make it clear:

- add `monix.execution.internal.Newtype1` (private to Monix, I hope will standardize on something in the community)
- add `object CombineObservable` like you are doing here, but extend it from `Newtype1` and so the `newtype` will be `CombineObservable.Type[+A]` (and also, there won't be a corresponding `class`)
- on this object add `apply` and `unwrap` functions that work with `.asInstanceOf`

Avasil: @alexandru 
Okay, I implemented something that works but I think it could be nice to organize it into something like what you did in you PR with ```IO.Par``` instead of ```CombineObservable.Type```.

Also I have followed your guidelines but is there any reason against implementing ```Newtype1``` ( based on your ```IONewtype``` ) as:
``` scala
private[monix] abstract class Newtype1[F[_]] { self =>
  type Base
  trait Tag extends Any
  type Type[+A] <: Base with Tag

  def apply[A](fa: F[A]): Type[A] =
    fa.asInstanceOf[Type[A]]

  def unwrap[A](fa: Type[A]): F[A] =
    fa.asInstanceOf[F[A]]
}
```

alexandru: Right, if it works, it's fine. Note that's probably an `F[+_]` in the case of `Observable`, but then I don't see a reason for why invariance wouldn't work here.

Avasil: It actually works with just ```F[_]``` for this use case

monix-reactive/shared/src/main/scala/monix/reactive/observables/CombineObservable.scala
```diff
@@ -21,9 +21,9 @@ import cats.Apply
 import monix.execution.internal.Newtype1
 import monix.reactive.Observable
 
-/** A `CombineObservable` is an observable that wraps a regular
-  * [[Observable]] and provide [[cats.Apply]] instance
-  * which uses [[Observable.combineLatest]] to combine elements.
+/** Newtype encoding for an [[Observable]] datatype that has a [[cats.Apply]]
+  * instance which uses [[Observable.combineLatest]] to combine elements
```
alexandru: Do a `unidoc` command in `sbt`, make sure these references are resolved - I know that it has problems with resolving imports, so if you're not in the same package you have to do the full path, but check first:

```
[[monix.reactive.Observable.combineLatest Observable.combineLatest]]
```

Avasil: Done :+:

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -4470,6 +4470,23 @@ object Observable {
...
+      import CombineObservable.unwrap
+      import CombineObservable.{apply => wrap}
+
+      def flatMap: FlatMap[Observable] = implicitly[FlatMap[Observable]]
+      def apply: Apply[CombineObservable.Type] = CombineObservable.combineObservableApplicative
+
+      def sequential = new (CombineObservable.Type ~> Observable) {
+        def apply[A](fa: CombineObservable.Type[A]): Observable[A] = unwrap(fa)
+      }
+      def parallel = new (Observable ~> CombineObservable.Type) {
```
alexandru: Please make `parallel` and `sequential` to be `val` instead of `def`, because we can reuse the `FunctionK` instances.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -4470,6 +4470,23 @@ object Observable {
     override def empty[A]: Observable[A] =
       Observable.empty[A]
   }
+
+  /** [[cats.NonEmptyParallel]] instance for [[Observable]]. */
+  implicit def observableNonEmptyParallel: NonEmptyParallel[Observable, CombineObservable.Type] =
```
alexandru: Instances with no dependencies should be `val` and not `def`, otherwise it's going to create an unnecessary new instance every time it's required. Make this a `val` too.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -4470,6 +4470,23 @@ object Observable {
...
+    new NonEmptyParallel[Observable, CombineObservable.Type] {
+      import CombineObservable.unwrap
+      import CombineObservable.{apply => wrap}
+
+      def flatMap: FlatMap[Observable] = implicitly[FlatMap[Observable]]
+      def apply: Apply[CombineObservable.Type] = CombineObservable.combineObservableApplicative
```
alexandru: These can remain `def`, since all they do is to return an already initialized instance (hopefully).

monix-reactive/shared/src/main/scala/monix/reactive/observables/CombineObservable.scala
```diff
@@ -0,0 +1,43 @@
...
+import monix.reactive.Observable
+
+/** Newtype encoding for an [[Observable]] datatype that has a [[cats.Apply]]
+  * instance which uses [[Observable.combineLatest]] to combine elements
+  * needed for implementing [[cats.NonEmptyParallel]]
+  */
+
+object CombineObservable extends Newtype1[Observable] {
+
+  implicit def combineObservableApplicative: Apply[CombineObservable.Type] = new Apply[CombineObservable.Type] {
```
alexandru: Make this a `val` too.

monix-reactive/shared/src/main/scala/monix/reactive/observables/CombineObservable.scala
```diff
@@ -0,0 +1,43 @@
...
+    import CombineObservable.{apply => wrap}
+
+    def ap[A, B](ff: CombineObservable.Type[(A) => B])(fa: CombineObservable.Type[A]) =
```
alexandru: Override `map2` as well.

And use `override` on implemented methods, even if they are abstract, because it's easier to protect against API breakage that way.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -4470,6 +4470,23 @@ object Observable {
...
       Observable.empty[A]
   }
+
+  /** [[cats.NonEmptyParallel]] instance for [[Observable]]. */
+  implicit def observableNonEmptyParallel: NonEmptyParallel[Observable, CombineObservable.Type] =
+    new NonEmptyParallel[Observable, CombineObservable.Type] {
+      import CombineObservable.unwrap
+      import CombineObservable.{apply => wrap}
+
+      def flatMap: FlatMap[Observable] = implicitly[FlatMap[Observable]]
```
alexandru: Use `override` even on abstract methods, because we need to protect against API breakage.

monix-reactive/shared/src/main/scala/monix/reactive/observables/CombineObservable.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import cats.Apply
+import monix.execution.internal.Newtype1
+import monix.reactive.Observable
+
+/** Newtype encoding for an [[Observable]] datatype that has a [[cats.Apply]]
+  * instance which uses [[Observable.combineLatest]] to combine elements
+  * needed for implementing [[cats.NonEmptyParallel]]
+  */
+
```
alexandru: Please delete this newline here. Personal pet peeve, plus Scaladoc might malfunction.

monix-reactive/shared/src/main/scala/monix/reactive/observables/CombineObservable.scala
```diff
@@ -0,0 +1,46 @@
...
+    import CombineObservable.{apply => wrap}
+
+    override def ap[A, B](ff: CombineObservable.Type[(A) => B])(fa: CombineObservable.Type[A]) =
+      wrap(unwrap(ff).combineLatestMap(unwrap(fa))((f, a) => f(a)))
+
+    override def map[A, B](fa: CombineObservable.Type[A])(f: A => B): CombineObservable.Type[B] =
+      wrap(unwrap(fa).map(f))
+
+    override def map2[A, B, C](fa: CombineObservable.Type[A],
+      fb: CombineObservable.Type[B])(f: (A, B) => C): CombineObservable.Type[C] =
```
Avasil: There might be nicer way to format it

alexandru: My preference is this:

```scala
override def map2[A, B, C](fa: CombineObservable.Type[A], fb: CombineObservable.Type[B])
  (f: (A, B) => C): CombineObservable.Type[C] = 
  wrap(unwrap(fa).combineLatestMap(unwrap(fb))(f))
```

Not much of a difference though.

PR #537, opened 2018-01-16
Make MVar constructors return Task

### LINKED ISSUE #535: Should MVar constructors return `Task`s?
Monix MVar constructors are not referentially transparent:

```scala
def mVar = MVar.empty[Int]
mVar.put(0).flatMap(_ => mVar.put(0)) // <-> Task.unit

// vs

val mVar = MVar.empty[Int]
mVar.put(0).flatMap(_ => mVar.put(0)) // <-> Task.never
```

[Haskell](https://hackage.haskell.org/package/base-4.0.0.0/docs/Control-Concurrent-MVar.html#v%3AnewMVar) and [Scalaz](https://github.com/scalaz/scalaz/blob/series/7.3.x/concurrent/src/main/scala/scalaz/concurrent/MVar.scala#L31) constructors both return `IO[MVar[A]]`, since allocating a mutable cell is an effect.

Is it intentional that Monix MVar does not wrap creation in `Task`?

oleg-py:
Closes #535

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/MVar.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/MVar.scala
@@ -89,26 +89,26 @@ abstract class MVar[A] {
 
 object MVar {
   /** Builds an [[MVar]] instance with an `initial` value. */
-  def apply[A](initial: A): MVar[A] =
-    new AsyncMVarImpl[A](AsyncVar(initial))
+  def apply[A](initial: A): Task[MVar[A]] =
+    Task.eval(new AsyncMVarImpl[A](AsyncVar(initial)))
 
   /** Returns an empty [[MVar]] instance. */
-  def empty[A]: MVar[A] =
-    new AsyncMVarImpl[A](AsyncVar.empty)
+  def empty[A]: Task[MVar[A]] =
+    Task.eval(new AsyncMVarImpl[A](AsyncVar.empty))
 
   /** Builds an [[MVar]] instance with an `initial`  value and a given
     * [[monix.execution.atomic.PaddingStrategy PaddingStrategy]]
     * (for avoiding the false sharing problem).
     */
-  def withPadding[A](initial: A, ps: PaddingStrategy): MVar[A] =
-    new AsyncMVarImpl[A](AsyncVar.withPadding(initial, ps))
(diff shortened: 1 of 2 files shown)


Avasil:
What are your thoughts on making it polymorphic in effect type like ```Iterant```?

Instead of:
``` scala
def apply[A](initial: A): Task[MVar[A]] =
  Task.eval(new AsyncMVarImpl[A](AsyncVar(initial)))
```
something like:
``` scala
def apply[F[_], A](initial: A)(implicit F: Sync[F]): F[MVar[A]] =
  F.suspend(new AsyncMVarImpl[A](AsyncVar(initial)))
```

Then you could use it with other types if you don't want to use ```Task``` for some reason

alexandru:
@Avasil that's a cool idea, however the transition wouldn't be smooth.

E.g. one pretty cool thing about `Task` is that the `Scheduler` gets injected. Thus you can describe async actions that need an `ExecutionContext` without taking it explicitly from the current scope:

```scala
    def take: Task[A] =
      Task.unsafeCreate { (context, callback) =>
        implicit val s = context.scheduler
        // Execution could be synchronous (e.g. result is null or not)
        av.unsafeTake(callback) match {
          case null => () // do nothing
          case a => callback.asyncOnSuccess(a)
        }
      }
```

This, plus `Task` is cancellable, although I didn't bother to describe a cancel action for `MVar`, because well, I don't think it's that useful plus it would make the implementation less performant.
But we can do that too, without waiting on `cats-effect`, which is an ongoing and slower process.

Another reason is consistency. The `monix-eval` project is about `Task` and `Coeval` and so you have `TaskSemaphore`, `TaskLocal` and so on. You can't describe `TaskLocal` for `cats-effect` and you'll probably never be able to do it, since the implementation is very tied to `Task`'s run-loop.
And this consistency is important, as people know that in `monix-eval` they've got abstractions optimized for `Task`.

That said I do want us to work on abstractions, like `MVar`, that can work on any `F[_]` implementing `cats-effect` stuff (n.b. for `MVar` you need `Async` and not `Sync`).

But those are going to be in a separate sub-project. And maybe we should make it non-monix (in `org.typelevel`). Maybe it can be a sub-project of `cats-effect`.

alexandru:
@oleg-py can't merge it because Travis-ci is currently having issues unrelated to your code - their queue is messed up.

PR #540, opened 2018-01-17
Make all Task abstractions referentially transparent

### LINKED ISSUE #539: All Task abstractions allocating mutable vars need to be built in the Task context
The `monix-eval` sub-project is about purity by means of `Task` and `Coeval`, therefore the abstractions exposed have to be referentially transparent with no exceptions.

All the ugly, mutable stuff goes into `monix-execution`, some concessions are acceptable for `monix-reactive`, but `monix-eval` and `monix-tail` need to be good FP citizens beyond reproach.

Currently these are:

- `MVar` (see discussion at #535 and PR #537)
- `TaskLocal`
- `TaskSemaphore`
- `TaskCircuitBreaker`

Also see discussion on the `cats-effect` chat channel: [January 17, 2018 1:18 PM](https://gitter.im/typelevel/cats-effect?at=5a5f30f26117191e6177201c)

Unfortunately this is a breaking change, but one that is done in order to correct a design error and that benefits users.

In Monix 3.0 such breaking changes are acceptable and better to do it now, since we've got an excuse in the deep integration with Cats.

oleg-py:
Closes #539.

I found existing circuit breaker & semaphore tests hard to convert since they check async things in sequence, so I used quick and easy `runSyncMaybe`. Let me know if it's not good enough.

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/MVar.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/MVar.scala
@@ -87,18 +87,33 @@ abstract class MVar[A] {
   def read: Task[A]
 }
 
+/** Builders for [[MVar]]
+  *
+  * @define refTransparent [[Task]] returned by this operation
+  *         produces a new [[MVar]] each time it is evaluated.
+  *         To share a state between multiple consumers, pass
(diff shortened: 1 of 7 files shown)


alexandru:
The fix with `runSyncMaybe` is fine, I'd do the same.

PR #543, opened 2018-01-18
Observable.take(0) shouldn't subscribe to the source at all

### LINKED ISSUE #541: Observable.take(0) shouldn't subscribe to the source at all
[January 17, 2018 11:37 PM](https://gitter.im/monix/monix?at=5a5fc21b5a9ebe4f759d4ba5):

```scala
var counter = 0
def inc() ={
  counter += 1
  1
}

val o = Observable.repeatEval( inc() ).take(0)
```

The fix should be that `take(0)` should simply return `Observable.empty`.

Fix should be back-ported to 2.3.x as well.

Avasil:
Fixes #541 
I went ahead and also fixed ```Observable.takeLast(0)```

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -1514,7 +1514,7 @@ abstract class Observable[+A] extends Serializable { self =>
     * buffer gets dropped and the error gets emitted immediately.
     */
   final def takeLast(n: Int): Observable[A] =
-    self.liftByOperator(new TakeLastOperator(n))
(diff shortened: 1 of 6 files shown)


Avasil:
Oh it seems like ```DelayBySelectorObservable``` relied on ```take(0)```
``` scala
      def onNext(elem: A): Future[Ack] = {
        currentElem = elem
        ack = Promise()

        var streamErrors = true
        try {
          val obs = selector(elem).take(0)
          streamErrors = false
          task := obs.unsafeSubscribeFn(trigger)
          ack.future
        }
        catch {
          case NonFatal(ex) if streamErrors =>
            onError(ex)
            Stop
        }
      }
```

PR #544, opened 2018-01-18
ported Observable.take(0) fix to 2.3.x

### LINKED ISSUE #541: Observable.take(0) shouldn't subscribe to the source at all

Avasil:
Fixes #541 
Hopefully this is right branch.

I've noticed that Copyright is outdated in this branch


alexandru:
Thanks @Avasil

PR #545, opened 2018-01-18
Add newtype for Task.Par, using same encoding used in cats-effect

### LINKED ISSUE #538: Add newtype for Task.Par, using same encoding used in cats-effect
The `cats.Parallel` instance is defined for `cats.Parallel[Task, Task]`.

This isn't OK, best practice is to define a newtype for that second type parameter.

In my PR on [Parallel[IO, ?]](https://github.com/typelevel/cats-effect/pull/115) I introduced a newtype encoding inspired by [alexknvl/newtypes](https://github.com/alexknvl/newtypes) ... we can't depend on that particular project, because (for Cats) it doesn't support Scala 2.10, plus I'm wary introducing dependencies. But we can use the same trick.

To coordinate with PR #536 that's doing the same for `Observable`.

Avasil:
Closes #538

### DIFF
--- a/monix-eval/shared/src/test/scala/monix/eval/TypeClassLawsForTaskWithCallbackSuite.scala
+++ b/monix-eval/shared/src/test/scala/monix/eval/TypeClassLawsForTaskWithCallbackSuite.scala
@@ -17,10 +17,11 @@
 
 package monix.eval
 
-import cats.Eq
 import cats.effect.laws.discipline.{AsyncTests, EffectTests}
 import cats.kernel.laws.discipline.MonoidTests
-import cats.laws.discipline.{CoflatMapTests, ParallelTests}
+import cats.laws.discipline.{ApplicativeTests, CoflatMapTests, ParallelTests}
+import cats.{Applicative, Eq}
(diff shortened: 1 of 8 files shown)


alexandru:
Looks good 👍

monix-eval/shared/src/test/scala/monix/eval/TypeClassLawsForTaskWithCallbackSuite.scala
```diff
@@ -29,6 +30,8 @@ import scala.concurrent.Promise
   * use of Task's `runAsync(callback)`.
   */
 object TypeClassLawsForTaskWithCallbackSuite extends BaseLawsSuite {
+  implicit val ap: Applicative[Task.Par] = CatsParallelForTask.applicative
```
alexandru: Ah, wait — so here we need a special equality defined for `Task.Par`, just like we have for `Task`. The purpose of this test suite is to test task's run-loop by using the `runAsync(callback)` method, instead of the normal `runAsync: CancelableFuture` that we are using in these tests.

alexandru:
Looking good @Avasil, merging.

PR #546, opened 2018-01-18
add scalafmt.conf + apply formatting

lorandszakacs:
Attempt to solve issue #513 

Now, I've added a sample `.scalafmt.conf` file, and applied it only to `Task.scala` for the sake of this PR. I glanced over other files, and it's mostly the same kind of formatting.

Leave any specific comments, and I'll try to hammer them out. Unforunately I have no idea how to keep all access modifiers on a separate line from top-level type definitions.
```
private [pack]
class Test {
//...
}
```
Always gets collapsed to:
```
private[pack] class Test {
//...
}
```

N.B.
You eliminate dangling stuff to the right by moving the closing parenthesis `)` on a newline, manually. Otherwise it — sometimes — exhibits the behavior of IntelliJ.

P.S.
`sbt scalafmt` formats everything. For tight trial and error cycles you can use the [IntelliJ plugin](https://plugins.jetbrains.com/plugin/8236-scalafmt).

P.P.S
If you want to check formatting via `scalafmt` as part of the CI you can simply add `scalafmtCheck` to the `ci` aliases in `build.sbt`


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -795,7 +795,9 @@ sealed abstract class Task[+A] extends Serializable {
     * obviously nothing gets executed at this point.
     */
   final def foreachL(f: A => Unit): Task[Unit] =
-    this.map { a => f(a); () }
+    this.map { a =>
+      f(a); ()
+    }
```
alexandru: Old version is better, sometimes one liners are better, but I can live with it.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1007,9 +1009,11 @@ sealed abstract class Task[+A] extends Serializable {
     * will be `maxRetries + 1`.
     */
   final def onErrorRestart(maxRetries: Long): Task[A] =
-    this.onErrorHandleWith(ex =>
-      if (maxRetries > 0) this.onErrorRestart(maxRetries-1)
-      else raiseError(ex))
+    this.onErrorHandleWith(
+      ex =>
```
alexandru: Meh.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1640,8 +1645,9 @@ object Task extends TaskInstancesLevel1 {
     *
     *  It's a simple version of [[traverse]].
     */
-  def sequence[A, M[X] <: TraversableOnce[X]](in: M[Task[A]])
-    (implicit cbf: CanBuildFrom[M[Task[A]], A, M[A]]): Task[M[A]] =
+  def sequence[A, M[X] <: TraversableOnce[X]](
+    in: M[Task[A]]
+  )(implicit cbf: CanBuildFrom[M[Task[A]], A, M[A]]): Task[M[A]] =
```
alexandru: I don't like this one 😑 

Are the styles available documented somewhere?

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -285,7 +285,7 @@ import scala.util.{Failure, Success, Try}
   *
   * @define optionsDesc a set of [[monix.eval.Task.Options Options]]
   *         that determine the behavior of Task's run-loop.
-  *         
+  *
```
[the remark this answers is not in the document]
alexandru: Well, in this case it isn't so bad, it just removed some whitespace, but I've seen IntelliJ do some weird shit in regards to the indenting of those `@define` paragraphs.

lorandszakacs:
@alexandru the space in `private[pack]` was my fault, I just wrote that block of code manually.

I'll fiddle with the config a bit more now.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
lorandszakacs: Yeah... unfortunately this is done by it's attempt to break up lines longer than 120 (as configured). You can usually get different results by manually putting certain curried parameter lists on separate lines. But automatically it always gives you these results... I don't think there is a way to configure which parameter lists ought to be moved to a newline, but I'll double check

The only documentation is here:
http://scalameta.org/scalafmt/#Configuration

From there onward it's trial and error. And I try to stay away from the list at the end, where certain properties can be removed without notice in future versions.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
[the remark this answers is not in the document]
lorandszakacs: Unfortunately it cannot be disabledd, you can either choose `docstrings=ScalaDoc` or `docstrings=JavaDoc`.

But fortunately, it never messes with the content of the scaladoc, only how the `*` char is aligned — and in this example I see it removed some whitespace (this is new to me).

Basically, the only real harm it does is this:
```scala
  /** Returns a string representation of this task meant for
   * debugging purposes only.
   */
```
To:
```scala
  /** Returns a string representation of this task meant for
    * debugging purposes only.
    */
```

So no need to worry about it messing around with docs like IntelliJ does. If in the future it ever supports it, then probably it will be configurable (at least disable-able).

monix-eval/shared/src/main/scala/monix/eval/Task.scala
lorandszakacs: Unfortunately you cannot configure it to always leave the function parameter after the `(`. And it moves stuff to the next line because the line was too long. But this particular case can be solved by using `{` instead of `(`.

lorandszakacs: I've converged a bit more towards the existing style, but this cannot be solved without using the `{ }` lambda syntax.

lorandszakacs:
@alexandru 
If you check the latest commit, you can see what kinds of manual fixes would be necessary to avoid really awkward formatting. I even think I got most of them. I skimmed through the rest of the project diff, and noticed nothing horrible.

62ae4a3974a3a8588135bc9a749d8817b8efbb74

Two important things to note:
1) run scalafmt twice, on a non-formatted codebase, because it might surprise you

2) In case classes and methods with many parameters, always put the ')' on a new line. It's the difference between this:
```scala
  def zip6[A1, A2, A3, A4, A5, A6](
    oa1: Observable[A1],
    oa2: Observable[A2],
    oa3: Observable[A3],
    oa4: Observable[A4],
    oa5: Observable[A5],
    oa6: Observable[A6]
  ): Observable[(A1, A2, A3, A4, A5, A6)] =
```
And this:
```scala
  def zip6[A1, A2, A3, A4, A5, A6](oa1: Observable[A1],
                                   oa2: Observable[A2],
                                   oa3: Observable[A3],
                                   oa4: Observable[A4],
                                   oa5: Observable[A5],
                                   oa6: Observable[A6]): Observable[(A1, A2, A3, A4, A5, A6)] =
```

lorandszakacs:
@alexandru, did you run `scalafmt` twice on that last commit?

In a minority of cases it is not idempotent for files previously not formatted with `scalafmt`
https://github.com/scalameta/scalafmt/issues/339

lorandszakacs:
@alexandru noticed some awkward right-leaning formatings. Fixing them now.

lorandszakacs:
@alexandru well, I think I got most awkward code formattings. I'm removed the "wip" status of this PR.

lorandszakacs:
Ok, this time I'm pretty sure I caught everything. Did an IntelliJ search for the regex: `\([^,\n)]*,\n`. I think this heuristic is enough to find all cases. It yields false positives, but I think no false negatives (of course, I can't be sure).

alexandru:
@lorandszakacs I reformatted the whole project and pushed the changes into your branch.

That was a mistake.

Can you please get rid of my commit with a rebase/reset, whatever works? :-)
Also, revert the test formatting that you did, I just want the configuration.

I'm thinking that we should merge the configuration but without reformatting the files at the moment.

Can scalafmt be configured in SBT to do incremental reformatting of changed files? Think I saw in the documentation that it cannot be, but just want to check a second opinion.

lorandszakacs:
@alexandru kept only the config and the sbt plugin update.

Unfortunately you cannot configure scalafmt to run in such an incremental way. Not even with the alternate [neo-sbt-scalafmt](https://github.com/lucidsoftware/neo-sbt-scalafmt) plugin.

alexandru:
@lorandszakacs thanks.

So we can enable automatic reformatting, but once we get some annoyances out of the way.

E.g. I asked on the Gitter channel about a newline rule that annoys me: https://gitter.im/scalameta/scalafmt?at=5a62508a9cdc721e4fa2aa5b

lorandszakacs:
@alexandru 

Good news, now it should be able to format the entire codebase without introducing abominations like this:
```scala
  def zip6[A1, A2, A3, A4, A5, A6](oa1: Observable[A1],
                                   oa2: Observable[A2],
                                   oa3: Observable[A3],
                                   oa4: Observable[A4],
                                   oa5: Observable[A5],
                                   oa6: Observable[A6]): Observable[(A1, A2, A3, A4, A5, A6)] =
```

Thanks to two configs I had previously missed:
```
align {
  openParenCallSite = false
  openParenDefnSite = false
}
```

alexandru:
Looks good, now all I'm waiting for is for this option to get released: https://github.com/scalameta/scalafmt/pull/1101 😀

alexandru:
It will take some time before a new stable version of scalafmt is released and given that we aren't doing this automatically yet, I'll just merge this configuration for now and activate it to do automatic formatting once that `blankLineBeforeDocstring` option gets released.

PR #549, opened 2018-01-19
Iterant.fromStateAction

### LINKED ISSUE #548: Add Iterant.fromStateAction / fromStateActionL
`Observable` has them and so should `Iterant` 😀

```scala
object Iterant {

  def fromStateAction[F[_], S, A](f: S => (A, S))(seed: => S)
    (implicit F: Sync[F]): Iterant[F, A] 

  def fromStateActionL[F[_], S, A](f: S => F[(A, S)])(seed: F[S])
    (implicit F: Sync[F]): Iterant[F, A]
}
```

Note: the simple version, `fromStateAction`, should emit `NextBatch` items with the default `recommendedBatchSize`.

The `fromStateActionL` overload should emit `Next` items, because presumably the evaluation of `F[_]` is expensive and we can't delay it.

Avasil:
Closes #548 
I've noticed mistake in one of my previous PR so just fixed this one line in this commit, I hope it's ok

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
@@ -25,11 +25,13 @@ import cats.{Applicative, CoflatMap, Eq, MonadError, Monoid, MonoidK, Order, Par
 import monix.eval.instances.{CatsAsyncForTask, CatsBaseForTask, CatsSyncForCoeval}
 import monix.eval.{Coeval, Task}
 import monix.execution.Scheduler
+import monix.execution.misc.NonFatal
 import monix.tail.batches.{Batch, BatchCursor}
 import monix.tail.internal._
 import org.reactivestreams.Publisher
 
 import scala.collection.immutable.LinearSeq
+import scala.collection.mutable
 import scala.reflect.ClassTag
 
 /** The `Iterant` is a type that describes lazy, possibly asynchronous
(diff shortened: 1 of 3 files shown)


monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -2046,6 +2048,66 @@ object Iterant extends IterantInstances {
...
+    * `Iterant` that keeps generating `NextBatch` items produced
+    * by our generator function with default `recommendedBatchSize`.
+    *
+    * Example: {{{
+    *   val f = (x: Int) => (x + 1, x * 2)
+    *   val seed = 1
+    *   val stream = Iterant.fromStateAction[Task, Int, Int](f)(seed)
+    *
+    *   // Yields 2, 3, 5, 9
+    *   stream.take(5)
```
alexandru: You know what would be cool?

To give a sample that generates prime numbers, using that function from the red book. Here it is:

```scala
  /**
    * Given a `seed`, generates a pseudo-random integer.
    */
  def int(seed: Seed): (Int, Seed) = {
    // `&` is bitwise AND. We use the current seed to generate a new seed.
    val newSeed = (seed * 0x5DEECE66DL + 0xBL) & 0xFFFFFFFFFFFFL
    // The next state, which is an `RNG` instance created from the new seed.
    val nextRNG = newSeed
    // `>>>` is right binary shift with zero fill. The value `n` is our new pseudo-random integer.
    val n = (newSeed >>> 16).toInt
    // The return value is a tuple containing both a pseudo-random integer and the next `RNG` state.
    (n, nextRNG)
}
```

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -2046,6 +2048,66 @@ object Iterant extends IterantInstances {
...
+  def fromStateActionL[F[_], S, A](f: S => F[(A, S)])(seed: => F[S])(implicit F: Sync[F]): Iterant[F, A] = {
+    import cats.syntax.all._
+
+    def loop(state: S): F[Iterant[F, A]] =
+      try {
+        f(state).map { case (elem, newState) =>
+          Next(elem, F.suspend(loop(newState)), F.unit)
+        }
+      } catch {
+        case e if NonFatal(e) => F.delay(Halt(Some(e)))
```
alexandru: This can be `F.pure(Halt(Some(e))`

alexandru:
Also see why Travis fails.

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
Avasil: I think you mean something else - it doesn't generate prime numbers :D

alexandru: Sure it does, take that function above and do ...

```scala
Iterant[Coeval].fromStateAction(int)(System.currentTimeMillis())
```

Now you've got a stream of prime numbers 😀

Anyway, we can put this in the doc article on monix.io 🙃

Avasil: I must be doing something wrong then because checking it out in worksheet gives me results like
```res0: (Int, Long) = (2033103875,133241495584024)```
 :D

Comment (alexandru) [reply monix-tail/shared/src/main/scala/monix/tail/Iterant.scala]: Yes, because it generates the current result + the next seed, which is how the "state monad" works, `S => (A, S)` being the signature, which is where the `fromStateAction` name comes:

```scala
import monix.reactive._

def randInt(seed: Long): (Int, Long) = {
  val newSeed = (seed * 0x5DEECE66DL + 0xBL) & 0xFFFFFFFFFFFFL
  val nextRNG = newSeed
  val n = (newSeed >>> 16).toInt
  (n, nextRNG)
}

Observable.fromStateAction(randInt)(System.currentTimeMillis())
  .dump("O")
  .take(10)
  .subscribe()

// 0: O --> -558270268
// 1: O --> 999348358
// 2: O --> 86307229
// 3: O --> -1618809049
// 4: O --> -491414624
// 5: O --> 2067091276
// 6: O --> -1907838351
// 7: O --> 93144131
// 8: O --> -203141602
// 9: O --> -1331681589
// 11: O stopped
```

alexandru: There, that's a purely functional stream of random integers 😀

Avasil: @alexandru 
I've corrected ```F.delay(Halt(Some(e)))``` but I'm not sure about this example, I'm afraid it might be hard to understand if you're not familiar with bit manipulation and you just browse scaladoc but at the same time my example might be too simple :D

Avasil: I meant that I don't see how it generates prime numbers (2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, ...)

alexandru: OK, it's too much for ScalaDoc 😀

Once green I'll merge it.

Avasil: I did it for test and it looks good in API documentation but it's hard to read when reading from IDE because of the comments so I agree it will be the coolest in monix.io

alexandru: Sure, no problem, we'll do it on some article in the documentation.

Avasil:
I don't know why 1 out of 5 builds is failing , I suppose it is not related to the code
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

PR #554, opened 2018-01-21
Add Iterant#bracket operation

### LINKED ISSUE #479: Add a .bracket operation for Iterant
FS2 has something like this:

```scala
def bracket(fa: F[A])(use: A => Stream[F, B], release: A => F[Unit]): Stream[F, B]
```

oleg-py:
Closes #479 

This is just the first pass for now. I'll look into implementing `bracket` as its own operator instead of combining bunch of existing ones, and there are few other things to consider too.


monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -507,6 +507,10 @@ sealed abstract class Iterant[F[_], A] extends Product with Serializable {
   final def doOnFinish(f: Option[Throwable] => F[Unit])(implicit F: Sync[F]): Iterant[F, A] =
     IterantStop.doOnFinish(this, f)(F)
 
+  // @TODO determine the destiny of this method
+  private[tail] final def doOnFullConsumption(f: Option[Throwable] => F[Unit])(implicit F: Sync[F]): Iterant[F, A] =
+    IterantStop.doOnFullConsumption(this, f)(F)
```
oleg-py: I added that operation because currently there's no way to handle early stop and full consumption separately (well, without closing over `var`s and checking them, or something like that)

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -1874,6 +1878,12 @@ sealed abstract class Iterant[F[_], A] extends Product with Serializable {
   *         the stream when evaluated
   */
 object Iterant extends IterantInstances {
+  // @TODO move somewhere else?
+  sealed trait BracketResult
+  case object Completed extends BracketResult
+  case object EarlyStop extends BracketResult
+  case class Error(e: Throwable) extends BracketResult
+
```
oleg-py: I was going to do something similar with typelevel/cats-effect#113. Tho here for `BracketResult`:
* The error type parameter is not needed, since `Iterant` is bound to `Throwable`
* Signaling the result in `Completed` probably does not make sense:
  * It's not possible to always signal a `B` from `use` because it might return an empty Iterant
  * It's possible to signal a `Seq[B]` (which would potentially require unlimited buffering, I'd rather not have an operator do that implicitly) or an `Option[B]` with the first element (for which I could not find any compelling use case)

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -1888,6 +1898,25 @@ object Iterant extends IterantInstances {
...
+  def bracket[F[_], A, B](acquire: F[A])(
+    use: A => Iterant[F, B],
+    release: (A, BracketResult) => F[Unit]
+  )(
+    implicit F: Sync[F]
+  ): Iterant[F, B] = {
+    Iterant.liftF(acquire).flatMap { a =>
+      Iterant.suspend(F.map(F.attempt(F.delay(use(a))))(_.fold(
+        ex => Iterant.raiseError[F, B](ex),
+        iterant => iterant
```
oleg-py: This weird machinery is required for lifting errors `use` might throw directly into `Iterant`. As a result, it will be reported as `Halt(Some(ex))` inside `Iterant[F, ?]` context . Alternatively, `release` could be called directly after `use` fails, and the error would be reported inside `F` itself. I went for the former.

See the test
> Bracket releases resource if `use` throws

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
alexandru: Indeed, we don't have a result for `Completed`. This type should be top-level in `monix.tail`.

But make `BracketResult` a `sealed abstract class` and the `case` items members of `object BracketResult`. Oh and that `case class` needs to be `final`. We think of them as `final`, but the Scala compiler doesn't do it for legacy reasons.

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -1888,6 +1898,25 @@ object Iterant extends IterantInstances {
...
+      )))
+        .doOnEarlyStop(F.suspend(release(a, EarlyStop)))
+        .doOnFullConsumption {
+          case None => F.suspend(release(a, Completed))
+          case Some(ex) => F.suspend(release(a, Error(ex)))
+        }
```
alexandru: I think you're trying too much to describe `bracket` with available operations.

`bracket` is kind of a primitive operation. I would do an ugly tail-recursive loop for it because you'd then have full control over error handling and so on.

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
alexandru: Actually this highlights that `doOnFinish` is not correctly defined.

If we are going to use the semantics of `Task.doOnFinish`, since after all that's where I got the name from, then `doOnFinish` shouldn't handle early stop.

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
oleg-py: Yeah, I kinda said that 😄 Just wanted to get tests right on the first pass

oleg-py:
Everything should be better now 😅 . Tho `doc` command seems to not like my example, failing the build:
```
/home/travis/build/monix/monix/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala:1906:58: Variable i undefined in comment for method bracket in object Iterant
 *         .mapEval(i => IO { writer.println(s"Line #$i") }),
                                                      ^
```

monix-tail/shared/src/main/scala/monix/tail/internal/IterantBracket.scala
```diff
@@ -0,0 +1,82 @@
...
+    release: (A, BracketResult) => F[Unit]
+  )(implicit F: Sync[F]): Iterant[F, B] = {
+
+    def earlyRelease(a: A) =
+      F.suspend(release(a, EarlyStop))
+
+    def loop(a: A)(source: Iterant[F, B]): Iterant[F, B] =
+      source match {
+        case Next(item, rest, stop) =>
+          Next(item, rest.map(loop(a)), earlyRelease(a) *> stop)
```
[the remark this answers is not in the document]
oleg-py: Ah, I think I get it now, after encountering a bug with my implementation 😞.

Seems the error check is unavoidable at least in `Suspend` case if we want guarantees to hold, which I assume we do.
I'll see tomorrow if `earlyStop` requires similar treatment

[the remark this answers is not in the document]
alexandru: So the thing is what happens if `F[_]` evaluates to `F.raiseError(e)`?

I think we need a global policy for dealing with error.
But we can't be defensive about it everywhere, such an implementation isn't feasible.

I also worry that the type classes in Cats do not allow for usage of our `Task.transformWith`, which is much better than succumbing to `attempt`.

oleg-py:
I went ahead and made everything more defensive in `bracket`, but it seems it requires #562 / #563 to handle broken batches or cursors 🤕 . Tho more tests won't hurt and error handling can be removed later if not really needed.

> I also worry that the type classes in Cats do not allow for usage of our Task.transformWith, which is much better than succumbing to attempt.

If it's so much better, and would not benefit only Monix, maybe it's worth proposing to add it to `cats.MonadError` with a default implementation based on `attempt`, so it can be overriden in Task and Coeval instances?

alexandru:
Yes, I think a proposal for a `transformWith` on `MonadError` would be worth it, but we have to take care of #562 and #563 in order to prove the need for it.

monix-tail/shared/src/main/scala/monix/tail/BracketResult.scala
```diff
@@ -0,0 +1,37 @@
...
+  * e.g. by `take` operation
+  *  - [[BracketResult$.Error Error]], for a stream which evaluation
+  * resulted in an error
+  */
+sealed abstract class BracketResult
+
+object BracketResult {
+  case object Completed extends BracketResult
+  case object EarlyStop extends BracketResult
+  final case class Error(e: Throwable) extends BracketResult
```
alexandru: I am very conflicted about this ADT because the same type might happen in that Bracket PR.

Are you OK with waiting to see what becomes of that PR?

oleg-py: Yeah, no problem in waiting.

I could add a weaker version of `bracket` that does not provide an ADT, and make the ADT version private for the time being (it should probably be renamed to be consistent with `Task`/`Coeval`).

oleg-py:
Per my testing, handling errors in left folds removes the need for doing it in `bracket`. Although I found few folds where it was not added yet (#577) - these still work with current version, but break if bracket does not handle errors.

monix-tail/shared/src/main/scala/monix/tail/BracketResult.scala
alexandru: @oleg-py a weaker version, along with the private full version / ADT sounds good.

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -1888,6 +1888,38 @@ object Iterant extends IterantInstances {
...
+    *     )
+    *
+    *   // Write 100 numbered lines to the file
+    *   // closing the writer when finished
+    *   writeLines.take(100).completeL.unsafeRunSync()
+    * }}}
+    */
+  def bracket[F[_], A, B](acquire: F[A])(
+    use: A => Iterant[F, B],
+    release: (A, BracketResult) => F[Unit]
```
alexandru: Btw, do you think it would make sense for `use` and `release` to be curried?
Maybe it's too much, but for `Task` it works with curly brackets, e.g..

```scala
task.bracket { in =>
  // use part
  ???
} { in =>
  // release part
  ???
}
```

Sort of resembles a `try/finally` statement to me and right now in Jan 2018 doesn't look ugly, but maybe I'll change my opinion in a month or so 😀

What do you think?

oleg-py: I copied over the currying on `acquire` from fs2 because without it inference does not work.

Semantics of this `Iterant.bracket` are different from `try/finally` and bracket discussed at `cats-effect`, as here you can continue `flatMap`ping safely. Actually, because of this property, my `use` blocks in fs2 bracket are often one-liners:

```scala
fs2.Stream.bracket(delay { file.open() })(
  file => fs2.Stream(file),
  file => delay { file.close() }
)
```

I think it's nicer this way than

```scala
fs2.Stream.bracket(delay { file.open() }) { file =>
  fs2.Stream(file)
} { file =>
  delay { file.close() }
}
```

although I would not mind having it curried just for the sake of consistency, as `Task` has it this way and it seems to be where `cats-effect` is going too.

alexandru: I think you're right that the semantics are different and that the `use` function gets to be very light.

So leave it as it is.

alexandru:
@oleg-py when you think it's ready for merge, pull it out of WIP state.

oleg-py:
@alexandru I'm pretty much done (or at least lacking more ideas for improvements / tests). #581 needs to be merged to ensure safety, as I already removed error handling on each step.

PR #555, opened 2018-01-21
Task's CoflatMap implementation should do `Task!.fork`

### LINKED ISSUE #553: Task's CoflatMap implementation should do `Task!.fork`
Task implements `CoflatMap`, but the implementation is basically wrapping a task with `pure`, being mostly useless.

We should change the implementation to be based on `fork` such that:

```scala
coflatMap(task) <-> task.fork
```

Avasil:
Fixes #553.

Do you want some tests with that?


alexandru:
No tests needed, as we should test for the laws in Cats and those are sufficient.

Please re-merge with `master`.

PR #558, opened 2018-01-22
WIP: Add Observable.mapParallelOrdered

### LINKED ISSUE #329: Add Observable.mapAsyncOrdered for parallel mapping over an observable
I've briefly discussed this on gitter before, but can't find a search feature in gitter to keep track of the conversation so moving to github.

I'm already making use of the awesome `Observable.mapAsync` but would love it if it would preserve the order of the elements.

Here is a use case as to why I'd like this operator.
I'm processing a large dataset which can potentially fail.  As the processing progresses, I'd like to periodically output the id of the record which was last successfully processed.  If the processing fails, I can then continue from the last id that was successfully processed (the source of my data provides the ability to resume from a known point).

If the Observable is using mapAsync, the ordering is no longer preserved and I can't easily record what the id of the last processed record is.  

If ordered parallel operations is painful, I'd be open to other patterns or suggestions to handle.

And for reference, this is the equivalent functionality in akka-streams
http://doc.akka.io/docs/akka/2.4/scala/stream/stages-overview.html#mapAsync (The default in akka is order preserving).

Avasil:
Closes #329 

This is my first attempt.
Tests are the same as for ```Observable#mapParallelUnordered``` minus backpressuring on buffer and I've added test for parallel processing and preserving order. 

It passes all of the test cases so that looks good but it is pretty likely I've got lost in proper cancelation, cleaning resources etc despite basing implementation on ```Observable#mapParallelUnordered```. :)

Things to consider:

- After my experience with default ```mutable.Queue``` in #509 I think we might need to use something else to get easiest boost in performance. Probably dedicated solution. Right now it is probably best to have anything working but I think this operator will be used a lot so we will have to tune its performance sooner or later.
- Sempahore is released as soon as computation is complete - not after getting ack from downstream. This is because there might be situation when we have ```parallelism = 1000``` and 999 tasks are idle but the first element is still doing some work, "wasting" time. I think it can lead to higher memory consumption but better throughput but that's the tradeoff to make. It should probably be limited with some kind of ```BlockingQueue``` so it doesn't consume memory forever if oldest task never finish.

I need to come back to this code after I clear my head but in the meantime I'm opening PR so you could see how is my progress. :)

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
@@ -1,0 +1,213 @@
+/*
+ * Copyright (c) 2014-2018 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.reactive.internal.operators
+
+import monix.eval.Task
+import monix.execution.Ack.{Continue, Stop}
+import monix.execution.cancelables.{CompositeCancelable, SingleAssignCancelable}
+import monix.execution.misc.{AsyncSemaphore, NonFatal}
+import monix.execution.{Ack, Cancelable, CancelableFuture}
+import monix.reactive.observers.{BufferedSubscriber, Subscriber}
+import monix.reactive.{Observable, OverflowStrategy}
+
+import scala.collection.mutable
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+private[reactive] final class MapParallelOrderedObservable[A, B](
+  source: Observable[A],
+  parallelism: Int,
+  f: A => Task[B],
+  overflowStrategy: OverflowStrategy[B]) extends Observable[B] {
+
+  override def unsafeSubscribeFn(out: Subscriber[B]): Cancelable = {
+    if (parallelism <= 0) {
+      out.onError(new IllegalArgumentException("parallelism > 0"))
+      Cancelable.empty
+    }
+    else if (parallelism == 1) {
+      // optimization for one worker
+      new MapTaskObservable[A, B](source, f).unsafeSubscribeFn(out)
+    }
+    else {
+      val composite = CompositeCancelable()
+      val subscription = new MapAsyncParallelSubscription(out, composite)
+      composite += source.unsafeSubscribeFn(subscription)
+      subscription
+    }
+  }
+
+  private final class MapAsyncParallelSubscription(
+    out: Subscriber[B], composite: CompositeCancelable)
+    extends Subscriber[A] with Cancelable { self =>
+
+    implicit val scheduler = out.scheduler
+    // Ensures we don't execute more then a maximum number of tasks in parallel
+    private[this] val semaphore = AsyncSemaphore(parallelism)
+    // Reusable instance for releasing permits on cancel, but
+    // it's debatable whether this is needed, since on cancel
+    // everything gets canceled at once
+    private[this] val releaseTask = Task.eval(semaphore.release())
+    // Buffer with the supplied  overflow strategy.
+    private[this] val buffer = BufferedSubscriber[B](out, overflowStrategy)
+
+    // Flag indicating whether a final event was called, after which
+    // nothing else can happen. It's a very light protection, as
+    // access to it is concurrent and not synchronized
+    private[this] var isDone = false
+    // Turns to `Stop` when a stop acknowledgement is observed
+    // coming from the `buffer` - this indicates that the downstream
+    // no longer wants any events, so we must cancel
+    private[this] var lastAck: Ack = Continue
+    // Buffer for signaling new elements to downstream preserving original order
+    private[this] val queue = mutable.Queue[CancelableFuture[B]]()
+
+    private def sendHeadIfPossible(): Unit = {
+      while (queue.nonEmpty && queue.head.isCompleted && !isDone) {
+        val head = queue.head
+        head.value match {
+          case Some(Success(value)) =>
+            buffer.onNext(value).syncOnComplete {
+              case Success(Stop) =>
+                lastAck = Stop
+                composite.cancel()
+                queue.dequeue()
+              case Success(Continue) =>
+                queue.dequeue()
+                composite -= head.cancelable
+              case Failure(ex) =>
+                lastAck = Stop
+                composite -= head.cancelable
+                self.onError(ex)
+            }
+
+          case Some(Failure(error)) =>
+            lastAck = Stop
+            composite -= head.cancelable
+            self.onError(error)
+
+          case None => // shouldnt get here
+        }
+      }
+    }
+
+    private def process(elem: A) = {
+      // For protecting against user code, without violating the
+      // observer's contract, by marking the boundary after which
+      // we can no longer stream errors downstream
+      var streamErrors = true
+      try {
+        // We need a forward reference, because of the
+        // interaction with the `composite` below
+        val subscription = SingleAssignCancelable()
+        composite += subscription
+
+        val task = f(elem).doOnCancel(releaseTask)
+        // No longer allowed to stream errors downstream
+        streamErrors = false
+        // Start execution
+        val future = task.runAsync
+        future.onComplete {
+          case Success(_) =>
+            // Computation is complete so we can accept new ones
+            semaphore.release()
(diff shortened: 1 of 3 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
```diff
@@ -0,0 +1,213 @@
...
+        // No longer allowed to stream errors downstream
+        streamErrors = false
+        // Start execution
+        val future = task.runAsync
+        future.onComplete {
+          case Success(_) =>
+            // Computation is complete so we can accept new ones
+            semaphore.release()
+            // We can only send current head element to downstream subscriber
+            sendHeadIfPossible()
```
Avasil: I see that I will need to synchronize it to avoid race condition

alexandru:
Sorry for not reviewing this @Avasil, this is a difficult one.

Avasil:
@alexandru 
No worries, I don't expect it going through very quickly, there is probably a lot of work ahead and it is important to get this right. 

I will come back to it during/after weekend to fix one bug I mentioned (should be easy to fix) and whatever else I find.

monix-reactive/jvm/src/test/scala/monix/reactive/internal/operators/MapParallelOrderedConcurrencySuite.scala
```diff
@@ -0,0 +1,53 @@
...
+
+package monix.reactive.internal.operators
+
+import cats.laws._
+import cats.laws.discipline._
+import monix.eval.Task
+import monix.reactive.{BaseConcurrencySuite, Observable}
+
+object MapParallelOrderedConcurrencySuite extends BaseConcurrencySuite {
+  test("mapParallelUnordered works concurrently") { implicit s =>
```
Wogan: fix the test name

monix-reactive/jvm/src/test/scala/monix/reactive/internal/operators/MapParallelOrderedConcurrencySuite.scala
```diff
@@ -0,0 +1,53 @@
...
+      }
+
+      val task1 = Observable.fromIterable(list).mapParallelOrdered(parallelism)(x => Task(x)).sumL
+      val task2 = Task.eval(list.sum)
+
+      task1 <-> task2
+    }
+  }
+
+  test("mapParallelUnordered works concurrently preserving order") { implicit s =>
```
Wogan: fix the test name

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
```diff
@@ -0,0 +1,218 @@
...
+    // access to it is concurrent and not synchronized
+    private[this] var isDone = false
+    // Turns to `Stop` when a stop acknowledgement is observed
+    // coming from the `buffer` - this indicates that the downstream
+    // no longer wants any events, so we must cancel
+    private[this] var lastAck: Ack = Continue
+    // Buffer for signaling new elements downstream preserving original order
+    private[this] val queue = new ConcurrentLinkedQueue[CancelableFuture[B]]
+    // This lock makes sure that only one thread at the time sends processed elements downstream
+    private[this] val sendDownstreamLock = new ReentrantLock()
```
alexandru: What is the advantage of a `ReentrantLock`, versus intrinsic locks with `synchronize`?

The later has the advantage that the JVM can do escape analysis in some instances and eliminate it. I once had a `SpinLock` in `monix.execution` btw, but got rid of it because I could not prove that it is better than plain `synchronize`, as the JVM does some weird shit with it.

Avasil: I needed `tryLock()` function because we only need one thread at a time to poll queue so if it is currently occupied the other thread doesn't have to wait. I don't think it's possible with `synchronized` but there might be alternative way to do it (assuming my approach is correct, I don't have full trust in tests with so much concurrency involved)

Avasil:
Thanks for some comments @alexandru (and @Wogan!) - I also need to come back to this with fresh head and see how can I simplify it

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
```diff
@@ -0,0 +1,218 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.reactive.internal.operators
+
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.locks.ReentrantLock
```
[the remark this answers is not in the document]
Avasil: I'm not too familiar with Scala.js - should I only rely on `monix` primitives / implement something new if needed?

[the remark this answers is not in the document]
alexandru: For `ConcurrentLinkedQueue` I'm pretty sure we have an alias or wrapper somewhere in `monix.execution` that you can use.

In the case of `ReentrantLock`, I have no idea what the Scala.js version is doing. Should be a dummy. I guess that if the tests pass, then it's not bad. Was surprised to see it in shared code, guess Scala.js has evolved its Java library.

So you can leave it, but was just mentioning that in case you introduce it in shared code, then you need to understand what the Scala.js implementation is doing, to verify that it's OK.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
alexandru: Ah, OK, `tryLock` is an interesting use case.

alexandru: OK, I understood what you did there 🙂 it's very interesting.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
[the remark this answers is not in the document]
Avasil: Leaving note about Scala.JS:

Reentrant lock:
https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/concurrent/locks/ReentrantLock.scala

So it just returns `true` for `tryLock` every time, sounds safe to use.

ConcurrentLinkedQueue:
https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala

This is a bit more complicated but glancing through the code it looks like ordinary `Queue`. 
I've found [Monix's ConcurrentQueue](https://github.com/monix/monix/blob/master/monix-reactive/jvm/src/main/scala/monix/reactive/observers/buffers/ConcurrentQueue.scala) but it's `private[buffer]`.

BTW @alexandru  

I'm trying to see how I can "merge" or remove any of those structures:
1. `AsyncSemaphore` - I need to limit incoming tasks.
2. `BufferedSubscriber` - I call `subscriber.onNext` concurrently.
3. `ConcurrentLinkedQueue` - order elements, need synchronized access because when one thread is checking `Queue` for head completion, new elements can arrive and they should be added to the queue.
4. `ReentrantLock` - I use it to ensure that only one thread at the time checks `Queue`, I only use `tryLock` so it shouldn't be big overhead


I'm trying to figure out if I can get rid of `BufferedSubscriber` since I call `onNext` only from one place but no luck so far - it's failing few tests.

Other than that I have only desparate idea which is implementing custom semi-bounded thread-safe  `Queue` which could track completion of its items and only allow up to N not-completed elements. Then backpressuring on it instead of `AsyncSemaphore` :D But even if that makes any sense I don't think I have capacity to do it soon and we probably want this operator in Monix 3.0.0 - a lot of people are requesting this.

alexandru:
I've been delaying this PR for months.

@Avasil please merge with `master` and re-test. Do you feel confident in the implementation?

I will take it for a spin, but I'd appreciate a re-review from you, maybe you get some fresh ideas.

Avasil:
@alexandru 
I will do it in the next couple of days, update and let you know. :)

alexandru:
👍 sounds good

Avasil:
@alexandru
I've merged to the latest changes and added more comments, hopefully it will make review easier. 
Unfortunately I didn't manage to come up with better design. Though I feel like this one should work. :) The only improvement I'm trying to get working is getting rid of `BufferedSubscriber`.

Most of the stuff is the same as unordered version, the difference is `sendDownstreamOrdered` with while loop and queue. I'll benchmark it to see if it's not 10x times slower than unordered variant but review in the meantime would be much appreciated!

alexandru:
We can merge it for the next release I think.

But please re-merge master.

Avasil:
Thanks @alexandru , I've re-merged with master. I was playing with getting rid of `BufferedSubscriber` in the implementation, I think it's doable but I didn't get anywhere at the time. Fortunately tuning internals later should be possible without breaking compatibility. :)

alexandru:
Indeed, we can tune the internals later, at this point it should just be correct.

PR #561, opened 2018-01-23
Bracket for Task/Coeval, Task.cancelable and Task.onCancelRaiseError

alexandru:
Introducing:

1. `task.cancelable` for activating the auto-cancelable run-loop for a certain task
2. `task.onCancelRaiseError` for throwing an error in case the source task gets cancelled, in a thread-safe way
3. `task.bracket` and `task.bracketE` for resource handling

## Bracket

The `bracket` operation would be the equivalent of `try {} finally {}` in many mainstream languages.
For both `Task` and `Coeval` it comes in two flavors. One would be the simple version:

```scala
import java.io._

def readFile(file: File): Task[String] = {
  // Opening a file handle for reading text
  val acquire = Task.eval(new BufferedReader(
    new InputStreamReader(new FileInputStream(file), "utf-8"))
  )

  acquire.bracket { in =>
    // Usage part
    Task.eval {
      // Yes, ugly Java, non-FP loop;
      // side-effects are suspended though
      var line: String = null
      val buff = new StringBuilder()
      do {
        line = in.readLine()
        if (line != null) buff.append(line)
      } while (line != null)
      buff.toString()
    }
  } { in =>
    // The release part
    Task.eval(in.close())
  }
}
```

And `bracketE` which allows distinguishing between successful termination, termination in error or, in the case of `Task`, cancellation:

```scala
def readFile(file: File): Task[String] = {
  // Opening a file handle for reading text
  val acquire = Task.eval(new BufferedReader(
    new InputStreamReader(new FileInputStream(file), "utf-8"))
  )

  acquire.bracketE { in =>
    // Usage part
    Task.eval {
      // Yes, ugly Java, non-FP loop;
      // side-effects are suspended though
      var line: String = null
      val buff = new StringBuilder()
      do {
        line = in.readLine()
        if (line != null) buff.append(line)
      } while (line != null)
      buff.toString()
    }
  } { (in, result) =>
    // The release part
    Task.eval {
      result match {
        case Right(_) => println("Successful!")
        case Left(Some(e)) => println("Raised error: " + e)
        case Left(None) => println("Canceled!")
      }
      in.close()
    }
  }
}
```

Also see the discussion at: https://github.com/typelevel/cats-effect/pull/113

## task.onCancelRaiseError

In Monix normally tasks are non-terminating on cancellation. This method transforms a task into one that terminates with error on completion:

```scala
import java.util.concurrent.CancellationException

val tenSecs = Task.sleep(10.seconds)
  .onCancelRaiseError(new CancellationException)

val task = tenSecs.fork.flatMap { fa =>
  // Triggering pure cancellation, then trying to get its result
  fa.cancel.flatMap(_ => fa)
}
```

## task.cancelable

This is the opposite of `uncancelable`, activating "auto-cancelable flatMap chains" for the given task on evaluation. Consider this loop that calculates the n-th Fibonacci number:

```scala
def fib(n: Int): Task[Long] = {
  def loop(n: Int, a: Long, b: Long): Task[Long] =
    Task.suspend {
      if (n > 0)
        loop(n - 1, b, a + b)
      else
        Task.now(a)
    }

  loop(n, 0, 1).cancelable
}
```

By specifying `cancelable` on the loop, we are ensuring that the internal cancellation flag is checked on asynchronous boundaries, automatically interrupting the evaluation if needed.


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -570,6 +570,112 @@ sealed abstract class Task[+A] extends Serializable {
...
+    * @param use is a function that evaluates the resource yielded by the source,
+    *        yielding a result that will get generated by this function on
+    *        evaluation
+    *
+    * @param release is a function that gets called after `use` terminates,
+    *        either normally or in error, or if it gets cancelled, receiving
+    *        as input the resource that needs that needs release, along with
+    *        the result of `use` (cancellation, error or successful result)
+    */
+  final def bracketE[B](use: A => Task[B])(release: (A, Either[Option[Throwable], B]) => Task[Unit]): Task[B] =
```
LukaJCB: Just to be 100% clear, `Either[Option[Throwable], B]` here means, that
`Left(None)` is cancelled, 
`Left(Some(ex))` is an error, and
`Right(b)` is success, correct? :)

alexandru: Yes. I'm also waiting on that ADT defined in `cats-effect`, for whenever it will happen, to introduce another overload.

LukaJCB: 👍

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -570,6 +570,112 @@ sealed abstract class Task[+A] extends Serializable {
...
+    * an error in the background with nowhere to go but in
+    * [[monix.execution.Scheduler.reportFailure Scheduler.reportFailure]].
+    *
+    * In this particular example, given that we are reading from a file,
+    * it doesn't matter. But in other cases it might matter, as concurrency
+    * on top of the JVM when dealing with I/O might lead to corrupted data.
+    *
+    * For those cases you might want to do synchronization (e.g. usage of
+    * locks and semaphores) and you might want to use [[bracketE]], the
+    * version that allows you to differentiate between nuormal termination
```
oleg-py: Typo: `nuormal` 😛

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -863,21 +969,65 @@ sealed abstract class Task[+A] extends Serializable {
...
+    * For example this is a function that calculates the n-th Fibonacci element:
     * {{{
-    *   task.executeWithOptions(_.enableAutoCancelableRunLoops)
+    *   def fib(n: Int): Task[Long] = {
+    *     def loop(n: Int, a: Long, b: Long): Task[Long] =
+    *       Task.suspend {
+    *         if (n > 0)
+    *           loop(n - 1, b, a + b)
+    *         else
+    *           a
```
oleg-py: Shouldn't it be `Task.now(a)`?

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -863,21 +969,65 @@ sealed abstract class Task[+A] extends Serializable {
...
+    * 
+    * This operation represents the opposite of [[uncancelable]]. And note
+    * that it works even for tasks that are uncancelable / atomic, because
+    * it blocks the rest of the `flatMap` loop from executing, functioning
+    * like a sort of cancellation boundary:
     *
+    * {{{
+    *   Task(println("Hello ..."))
+    *     .cancelable
+    *     .flatMap(_ => println("World!"))
```
oleg-py: I thikn `println` needs to be wrapped in `Task`

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1017,6 +1167,46 @@ sealed abstract class Task[+A] extends Serializable {
...
+    * cancellation.
+    *
+    * This `onCancelRaiseError` operator transforms a task that would yield
+    * [[Task.never]] on cancellation into one that yields [[Task.raiseError]].
+    *
+    * Example:
+    * {{{
+    *   import java.util.concurrent.CancellationException
+    *
+    *   val tenSecs = Task.sleep(10.seconds).onCancelRaiseError
```
oleg-py: Forgotten parameter for `onCancelRaiseError`

monix-eval/shared/src/main/scala/monix/eval/internal/TaskBracket.scala
```diff
@@ -0,0 +1,50 @@
...
+    */
+  def apply[A, B](
+    acquire: Task[A],
+    use: A => Task[B],
+    release: (A, Either[Option[Throwable], B]) => Task[Unit]): Task[B] = {
+
+    acquire.flatMap { a =>
+      val next = try use(a) catch { case NonFatal(e) => Task.raiseError(e) }
+      next.onCancelRaiseError(isCancel).transformWith[B](
+        b => release(a, Right(b)).map(_ => b),
```
oleg-py: So as I understand that correctly, if `release` throws or return a failed task, it will shadow the exception from `use`. Maybe it should use `reportFailure` from current scheduler instead?

monix-eval/shared/src/main/scala/monix/eval/Task.scala
alexandru: 👍

monix-eval/shared/src/main/scala/monix/eval/Task.scala
alexandru: 👍

monix-eval/shared/src/main/scala/monix/eval/Task.scala
alexandru: 👍

monix-eval/shared/src/main/scala/monix/eval/Task.scala
alexandru: 👍

oleg-py:
My last comment got hidden, but it kinda was important: what happens if both `use` and `release` result in a failure?

alexandru:
@oleg-py that's a very good point.

I added a detailed reply here: https://github.com/typelevel/cats-effect/pull/113#issuecomment-360386552

Currently the behavior, if both `use` and `release` throw, is this:

1. the error sent downstream is the one in `release`
2. the other error is completely silent

I'll modify the behavior to be consistent with the one in Haskell, that I just tested:

1. the error sent downstream should be the one in `use`
2. the error in `release` should be reported using `Scheduler.reportError`

alexandru:
@oleg-py I have implemented the changes.

alexandru:
@olafurpg given your involvement in `bracket`, I'd really like your approval for merging this PR.

So what say you? Does it look good after the latest changes?

alexandru:
Oh, above I misspelled the name, sorry @olafurpg, I wanted to mention @oleg-py, both of you have usernames starting with "ol" 😜

oleg-py:
@alexandru looks great 👍

ngbinh:
wow! This is awesome!

olafurpg:
`bracket` looks amazing, looking forward to give it a try!

Issue #559, opened 2018-01-22
Iterant error handling is leaky


oleg-py:
Found those cases while I was looking for correct implementation for #554 

These throw:

```scala
Iterant[IO]
  .eval { throw new Exception }
  .onErrorIgnore.completeL.unsafeRunSync()

Iterant[IO].of(1)
  .mapEval(_ => IO { throw new Exception })
  .onErrorIgnore.completeL.unsafeRunSync()

Iterant[IO].liftF(IO { throw new Exception })
  .onErrorIgnore.completeL.unsafeRunSync()

Iterant[IO].of(1)
  .mapEval(_ => IO(???))
  .attempt.toListL.unsafeRunSync()
```

although these do not:
```scala
Iterant[IO].raiseError(new Exception)
  .onErrorIgnore.completeL.unsafeRunSync()

Iterant[IO].of(1).scanEval(IO[Int](???))((_, _) => ???)
  .onErrorIgnore.completeL.unsafeRunSync()
```

alexandru:
Thanks for investigating.

We need a policy for dealing with errors triggered in the `F[_]` context.

My feeling right now is that the `L` operations, that are folding the iterant and returning `F[_]`, should be the ones in charge.

My only worry is that we'll suffer from performance issues, however I think we can come up with a trick to not do a `handleWith` on each `F[_]` boundary.

We could do a single `handleWith` only once, that wraps everything, the problem is that we need to know the current `earlyStop`, so I wonder if an implementation like the following would work:

```scala
private[tail] object IterantCompleteL {
  /**
    * For boxing stuff. We could close of a plain `var`, but Scala does this
    * under the hood as well (with `ObjectRef`) and I'd rather do this explicitly.
    */
  final class Box[A](var value: A)

  /**
    * Implementation for `Iterant#completeL`
    */
  final def apply[F[_], A](source: Iterant[F, A])
    (implicit F: Sync[F]): F[Unit] = {

    def loop(box: Box[F[Unit]], self: Iterant[F, A]): F[Unit] = {
      try self match {
        case Next(_, rest, stop) =>
          box.value = stop
          rest.flatMap(loop(box, _))
        case NextCursor(cursor, rest, stop) =>
          box.value = stop
          while (cursor.hasNext()) cursor.next()
          rest.flatMap(loop(box, _))
        case NextBatch(gen, rest, stop) =>
          box.value = stop
          val cursor = gen.cursor()
          while (cursor.hasNext()) cursor.next()
[...]

Issue #563, opened 2018-01-23
Iterant fold left operators (yielding F[_]) need to handle errors thrown in F[_] context


alexandru:
See #559

PR #566, opened 2018-01-25
Improve safety of `attempt` & `onErrorHandleWith`

### LINKED ISSUE #562: Iterant.onErrorHandleWith needs to lift F[_] errors into the Iterant context
See #559.

And all error-related operators that should depend (or that should depend) on `onErrorHandleWith` need to be verified.

oleg-py:
Closes #562.

I went ahead and made `attempt` safer too.

Incorporates tests from #559 comments and my latest work on `bracket`

### TODO:
- [x] Use `earlyStop` in guards

---

@alexandru should `earlyStop` be called if the evaluation of `tail` results on error? (I suppose it should).

I noticed some inconsitency that seems to not match my understanding of `earlyStop`:

```scala
private[tail] object IterantCompleteL {
  /**
    * Implementation for `Iterant#completeL`
    */
  final def apply[F[_], A](source: Iterant[F, A])
    (implicit F: Sync[F]): F[Unit] = {

    def loop(self: Iterant[F, A]): F[Unit] = {
      try self match { /* ...evicted... */ } catch {
        case ex if NonFatal(ex) =>
          source.earlyStop *> F.raiseError(ex)
      }
    }
```
Notice that it calls `source.earlyStop`. But in your comment for #559 you said:

> ... use a freaking var to hold the latest earlyStop reference 

and in this implementation the latest reference would be `self.earlyStop`. Most other operators name parameter inside loop `source` instead of `self`, shadowing the one that `apply` accepts, so they use latest `earlyStop`. Is that a bug?

Another question is the one I raised a number of times already 😄 .

```scala
Suspend(IO.raiseError(ex1), IO.raiseError(ex2)).onErrorHandleWith { ex =>
   /* is ex == ex1 or ex == e2 ? */
}
```

I assume per discussion in typelevel/cats-effect#113 we are going to swallow the one in earlyStop (which is akin to `release`), and there's nowhere it can go.


alexandru:
> Notice that it calls source.earlyStop. But in your comment for #559 you said:

That's not an inconsistency. It's a bug ☹️ as it should call `self.earlyStop`.

Unfortunately such bugs are easy when using inner functions and it's hard to test for it in this case.

alexandru:
About this:

```scala
Suspend(IO.raiseError(ex1), IO.raiseError(ex2)).onErrorHandleWith { ex =>
   /* is ex == ex1 or ex == e2 ? */
}
```

We can't swallow the one in `onErrorHandleWith` because it would probably violate the `MonadError` laws. `onErrorHandleWith` behaves like `flatMap` for errors. When the user returns an `F.raiseError` from `onErrorHandleWith`, we have to assume that's what he wants. And if an error gets thrown by the function given to `onErrorHandleWith`, we have to treat that as if the user returned an `F.raiseError`.

This is a different situation to `bracket`, as `onErrorHandleWith` is the equivalent of `catch`, not the equivalent of `finally`.

alexandru:
FYI — I am fixing `completeL`.

Avasil:
> FYI — I am fixing completeL.
Including entire #563?

oleg-py:
@alexandru I don't *return* raiseError 😛. As my Iterant, I have a `Suspend` of `rest` and `earlyStop` that both fail. Which error of two do I pass into `onErrorHandleWith`?

Fix for `completeL` will be nice - it should make `bracket` handle broken cursors

alexandru:
@Avasil not the entire issue, if you want you can take care of the others. Was thinking to put into practice my `var` idea to see if it works.

alexandru:
@oleg-py that's a tough call — prefer `ex1` (the error from `rest`), but do what's easy.

alexandru:
Created the PR - https://github.com/monix/monix/pull/567/files

Avasil:
Ok, I can take care of the others tomorrow, should be easy if your idea works, I'll do comparison benchmark too

monix-tail/shared/src/main/scala/monix/tail/internal/IterantOnError.scala
```diff
@@ -48,18 +48,23 @@ private[tail] object IterantOnError {
       Suspend[F, A](next, stop)
     }
 
+    def tailGuard(ffa: F[Iterant[F, A]]): F[Iterant[F, A]] =
+      ffa.handleError(f)
```
alexandru: PS: gone ahead and opened a ticket for a `transformWith` in `MonadError`, fingers crossed — https://github.com/typelevel/cats/issues/2161

We can't wait for it though. It will be ages before that happens, if at all 😀

monix-tail/shared/src/main/scala/monix/tail/internal/IterantOnError.scala
```diff
@@ -101,16 +106,20 @@ private[tail] object IterantOnError {
 
   /** Implementation for `Iterant.attempt`. */
   def attempt[F[_], A](fa: Iterant[F, A])(implicit F: Sync[F]): Iterant[F, Either[Throwable, A]] = {
+    type ErrOrA = Either[Throwable, A]
```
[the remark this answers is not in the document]
oleg-py: 👍

monix-tail/shared/src/main/scala/monix/tail/internal/IterantOnError.scala
oleg-py: That won't hurt. In the meantime, I think, users will appreciate correctness anyway 😄

monix-tail/shared/src/test/scala/monix/tail/IterantOnErrorSuite.scala
```diff
@@ -83,4 +85,64 @@ object IterantOnErrorSuite extends BaseTestSuite {
     }
   }
 
+  test("attempt & onErrorHandleWith should protect against broken continuations") { _ =>
```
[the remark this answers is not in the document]
oleg-py: I hope it's okay if I just extract common code to a method? 😏

oleg-py:
Ok, that should be it. I decided not to use the `var` trick - it seems to complicate the code with no benefit (as I cannot reduce the number of calls to `handleErrorWith`).

monix-tail/shared/src/main/scala/monix/tail/internal/IterantOnError.scala
```diff
@@ -48,18 +49,23 @@ private[tail] object IterantOnError {
...
     }
 
+    def tailGuard(stop: F[Unit])(ffa: F[Iterant[F, A]]): F[Iterant[F, A]] =
+      ffa.handleError(sendError(stop, _))
+
     def loop(fa: Iterant[F, A]): Iterant[F, A] =
       try fa match {
         case Next(a, lt, stop) =>
-          Next(a, lt.map(loop), stop)
+          Next(a, tailGuard(stop)(lt.map(loop)), stop)
```
alexandru: Everything looks good except that I feel this call is overreaching. It should probably not include the `.map(loop)`:

```scala
tailGuard(lt, stop).map(loop)
```

For one this is because it's unnecessary and also because we're avoiding optimization opportunities — e.g. `lt` could be a `pure` value for which the `F[_]` implementation could bypass error handling.

alexandru:
@oleg-py is there some channel to reach you? Can we chat on Gitter? 😀

oleg-py:
I moved `map` outside the guard for `handleError`, but not for `attempt` since that way types don't align. Also, removed currying, it seems totally redundant now

alexandru:
Thanks @oleg-py

PR #567, opened 2018-01-25
Issue #563 - completeL should handle F[_] errors, mapEval should not

alexandru:
Related to #563 — fixed `completeL` with a reasonable penalty for doing that error handling, due to mutating a plain var.

In case of this loop, it doesn't matter due to the inner workings of the loop being suspended and completely encapsulated. However this isn't an optimization that we can afford to do for operations returning `Iterant`.

Therefore, except for operations meant specifically for error handling, like `onErrorHandlingWith`, the errors raised in the `F[_]` context should get handled by folding operations (ending with the `L` suffix), one of which being `completeL`.


alexandru:
@oleg-py Not waiting any longer on this PR, merging it directly to unblock your testing.

Implementation seems fine to me - the internal variable is completely encapsulated.

One note - `mapEval` no longer handles `F[_]` errors by itself, even if it still catches exceptions thrown directly from its function parameter (maybe we shouldn't do that either, since `Task` or `IO` will catch that exception anyway). The convention going forward is that the fold operations (like `completeL`) are in charge of handling `F[_]` errors. Or the `onError*` ops.

PR #569, opened 2018-01-26
Iterant fold left operators need to handle errors thrown in F[_] context

### LINKED ISSUE #563: Iterant fold left operators (yielding F[_]) need to handle errors thrown in F[_] context

Avasil:
Fixes #563

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantFoldLeft.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantFoldLeft.scala
@@ -19,49 +19,65 @@ package monix.tail.internal
 
 import cats.effect.Sync
 import cats.syntax.all._
+import monix.execution.misc.NonFatal
 import monix.tail.Iterant
 import monix.tail.Iterant.{Halt, Last, Next, NextBatch, NextCursor, Suspend}
+
 import scala.collection.mutable
-import monix.execution.misc.NonFatal
+import scala.runtime.ObjectRef
 
 private[tail] object IterantFoldLeft {
   /**
     * Implementation for `Iterant#foldLeftL`
     */
   final def apply[F[_], S, A](source: Iterant[F, A], seed: => S)(op: (S,A) => S)
-    (implicit F: Sync[F]): F[S] = {
+                             (implicit F: Sync[F]): F[S] = {
 
-    def loop(self: Iterant[F, A], state: S): F[S] = {
+    def loop(stopRef: ObjectRef[F[Unit]])(self: Iterant[F, A], state: S): F[S] = {
       try self match {
-        case Next(a, rest, _) =>
+        case Next(a, rest, stop) =>
+          stopRef.elem = stop
           val newState = op(state, a)
-          rest.flatMap(loop(_, newState))
-        case NextCursor(cursor, rest, _) =>
+          rest.flatMap(loop(stopRef)(_, newState))
+        case NextCursor(cursor, rest, stop) =>
+          stopRef.elem = stop
           val newState = cursor.foldLeft(state)(op)
-          rest.flatMap(loop(_, newState))
-        case NextBatch(gen, rest, _) =>
+          rest.flatMap(loop(stopRef)(_, newState))
+        case NextBatch(gen, rest, stop) =>
+          stopRef.elem = stop
           val newState = gen.foldLeft(state)(op)
-          rest.flatMap(loop(_, newState))
-        case Suspend(rest, _) =>
-          rest.flatMap(loop(_, state))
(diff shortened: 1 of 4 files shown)


Avasil:
I did benchmark for```Iterant#toListL``` to satisfy my curiosity:
```scala
Before:
[info] Benchmark                                 Mode  Cnt       Score      Error  Units
[info] IterantFoldLeftLBenchmark.toListL100     thrpt  100  391092.509 ± 3795.786  ops/s
[info] IterantFoldLeftLBenchmark.toListL10000   thrpt  100    4020.913 ±   30.503  ops/s
[info] IterantFoldLeftLBenchmark.toListL100000  thrpt  100     425.652 ±    1.370  ops/s

After:
[info] Benchmark                                 Mode  Cnt       Score      Error  Units
[info] IterantFoldLeftLBenchmark.toListL100     thrpt  100  361662.963 ± 4809.333  ops/s
[info] IterantFoldLeftLBenchmark.toListL10000   thrpt  100    3952.470 ±   48.976  ops/s
[info] IterantFoldLeftLBenchmark.toListL100000  thrpt  100     376.399 ±    3.289  ops/s
```

monix-tail/shared/src/main/scala/monix/tail/internal/IterantFoldLeft.scala
```diff
@@ -31,37 +33,51 @@ private[tail] object IterantFoldLeft {
   final def apply[F[_], S, A](source: Iterant[F, A], seed: => S)(op: (S,A) => S)
     (implicit F: Sync[F]): F[S] = {
 
-    def loop(self: Iterant[F, A], state: S): F[S] = {
+    def loop(stopRef: ObjectRef[F[Unit]])(self: Iterant[F, A], state: S): F[S] = {
```
alexandru: Lets do currying here. Put the `state` on the left, such that you can then do ... 

```scala
rest.flatMap(loop(stopRef, newState))
```

monix-tail/shared/src/main/scala/monix/tail/internal/IterantFoldWhileLeft.scala
```diff
@@ -45,26 +47,30 @@ private[tail] object IterantFoldWhileLeft {
       if (hasResult)
         stop.map(_ => s)
       else
-        rest.flatMap(loop(s))
+        rest.flatMap(loop(stopRef)(s))
     }
 
-    def loop(state: S)(self: Iterant[F, A]): F[S] = {
+    def loop(stopRef: ObjectRef[F[Unit]])(state: S)(self: Iterant[F, A]): F[S] = {
```
alexandru: We don't need 3 parameter groups for this curried function. `stopRef` and `state` can be on the same level so to speak.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantFoldWhileLeft.scala
```diff
@@ -45,26 +47,30 @@ private[tail] object IterantFoldWhileLeft {
...
+          process(stopRef)(state, cursor, rest, stop)
 
-        case Suspend(rest, _) =>
-          rest.flatMap(loop(state))
+        case Suspend(rest, stop) =>
+          stopRef.elem = stop
+          rest.flatMap(loop(stopRef)(state))
 
         case Last(a) =>
           F.pure(f(state, a) match {
```
alexandru: Calling `f` can throw, which can trigger an exception that's going to trigger our `handleErrorWith`, which is going to call an `earlyStop` reference that's no longer in service.

Do `stopRef.elem = null`.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantFoldWhileLeft.scala
```diff
@@ -74,55 +80,69 @@ private[tail] object IterantFoldWhileLeft {
...
           }
 
-        case Suspend(rest, _) =>
-          rest.flatMap(loop(state))
+        case Suspend(rest, stop) =>
+          stopRef.elem = stop
+          rest.flatMap(loop(stopRef)(state))
 
         case Last(a) =>
           f(state, a).map {
```
alexandru: Same problem here. `stopRef.elem = null`. If unsure then it's easier to think that this assignment needs to happen on all branches, unless you can reason about why it isn't needed.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantFoldWhileLeft.scala
Avasil: I see, I've fixed it and added tests for this case - please see if I understand desired behavior correctly. :)

PR #575, opened 2018-01-28
Add Iterant.repeat (method and builder)

### LINKED ISSUE #571: Add Iterant.repeat (method)
Need the following builder for `Iterant`:

```scala

  def repeat(implicit F: Sync[F]): Iterant[F, A]
}
```

Should be equivalent of [RepeatSourceObservable](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/RepeatSourceObservable.scala).

Should work like this:

1. mirrors the source exactly, until `Halt(None)`
2. when `Halt(None)` happens, this operator should swallow it and return the source again, thus restarting it
3. the objective being to create a loop that goes over the source again and again

### LINKED ISSUE #572: Add Iterant.repeat (builder)
We need an `Iterant` builder with this signature:

```scala
def repeat[F[_], A](elems: A*)(implicit F: Sync[F]): Iterant[F, A]
```

Could depend on #571 but it would be more efficient to just wrap those `elems` in a `NextBatch` (or `Next` if its `length` is 1) and then have `rest` go in loop.

Avasil:
Fixes #572.
Fixes #571.

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
@@ -2138,8 +2138,16 @@ object Iterant extends IterantInstances {
     */
   def repeat[F[_], A](elems: A*)(implicit F: Sync[F]): Iterant[F, A] = elems match {
     case Seq() => Iterant.empty
-    case Seq(elem) => Next[F, A](elem, F.delay(repeat(elem)), F.unit)
-    case _ => NextBatch[F, A](Batch(elems: _*), F.delay(repeat(elems: _*)), F.unit)
+    case Seq(elem) =>
+      // trick to optimize recursion, see Optimisation section at:
+      // https://japgolly.blogspot.com.by/2017/12/practical-awesome-recursion-ch-02.html
+      var result: Iterant[F, A] = null
+      result = Next[F, A](elem, F.delay(result), F.unit)
+      result
+    case _ =>
+      var result: Iterant[F, A] = null
+      result = NextBatch(Batch(elems: _*), F.delay(result), F.unit)
+      result
   }
 
   /** Returns an empty stream. */
--- a/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
@@ -1235,6 +1235,13 @@ sealed abstract class Iterant[F[_], A] extends Product with Serializable {
   final def reduceL(op: (A, A) => A)(implicit F: Sync[F]): F[Option[A]] =
     IterantReduce(self, op)
 
+  /** Repeats the items emitted by the source continuously
+    *
+    * It terminates either on error or if the source is empty.
+    */
+  final def repeat(implicit F: Sync[F]): Iterant[F, A] =
+    IterantRepeat(self)
+
   /** Returns an `Iterant` that mirrors the behavior of the source,
     * unless the source is terminated with an error, in which case
     * the streaming of events continues with the specified backup
@@ -2124,6 +2131,20 @@ object Iterant extends IterantInstances {
     */
   def range[F[_]](from: Int, until: Int, step: Int = 1)(implicit F: Applicative[F]): Iterant[F, Int] =
     NextBatch(Batch.range(from, until, step), F.pure(empty[F, Int]), F.unit)
+
+  /** Builds a stream that repeats the items provided in argument.
+    *
+    * It terminates either on error or if the source is empty.
+    */
+  def repeat[F[_], A](elems: A*)(implicit F: Sync[F]): Iterant[F, A] = {
+    try elems match {
+      case Seq() => Iterant.empty
+      case Seq(elem) => Next[F, A](elem, F.delay(repeat(elem)), F.unit)
+      case _ => NextBatch[F, A](Batch(elems: _*), F.delay(repeat(elems: _*)), F.unit)
+    } catch {
+      case e if NonFatal(e) => Halt(Some(e))
+    }
+  }
 
   /** Returns an empty stream. */
   def empty[F[_], A]: Iterant[F, A] =
--- a/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
@@ -2136,14 +2136,10 @@ object Iterant extends IterantInstances {
     *
     * It terminates either on error or if the source is empty.
     */
-  def repeat[F[_], A](elems: A*)(implicit F: Sync[F]): Iterant[F, A] = {
-    try elems match {
-      case Seq() => Iterant.empty
-      case Seq(elem) => Next[F, A](elem, F.delay(repeat(elem)), F.unit)
-      case _ => NextBatch[F, A](Batch(elems: _*), F.delay(repeat(elems: _*)), F.unit)
-    } catch {
-      case e if NonFatal(e) => Halt(Some(e))
-    }
+  def repeat[F[_], A](elems: A*)(implicit F: Sync[F]): Iterant[F, A] = elems match {
+    case Seq() => Iterant.empty
+    case Seq(elem) => Next[F, A](elem, F.delay(repeat(elem)), F.unit)
+    case _ => NextBatch[F, A](Batch(elems: _*), F.delay(repeat(elems: _*)), F.unit)
   }
 
   /** Returns an empty stream. */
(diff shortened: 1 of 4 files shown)


monix-tail/shared/src/main/scala/monix/tail/internal/IterantRepeat.scala
```diff
@@ -0,0 +1,57 @@
...
+import monix.tail.Iterant.{Halt, Last, Next, NextBatch, NextCursor, Suspend}
+import monix.tail.internal.IterantUtils.signalError
+import cats.syntax.all._
+
+private[tail] object IterantRepeat {
+  /** Implementation for `Iterant.repeat`. */
+  def apply[F[_], A, B](source: Iterant[F, A])(implicit F: Sync[F]): Iterant[F, A] = {
+
+    def loop(self: Iterant[F, A]): Iterant[F, A] =
+      try self match {
```
Avasil: I'm not sure if this is needed

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -2125,6 +2132,16 @@ object Iterant extends IterantInstances {
...
     NextBatch(Batch.range(from, until, step), F.pure(empty[F, Int]), F.unit)
 
+  /** Builds a stream that repeats the items provided in argument.
+    *
+    * It terminates either on error or if the source is empty.
+    */
+  def repeat[F[_], A](elems: A*)(implicit F: Sync[F]): Iterant[F, A] = elems match {
+    case Seq() => Iterant.empty
+    case Seq(elem) => Next[F, A](elem, F.delay(repeat(elem)), F.unit)
+    case _ => NextBatch[F, A](Batch(elems: _*), F.delay(repeat(elems: _*)), F.unit)
```
oleg-py: It should be possible to optimize away the recursion / allocation on each step like so:
```scala
var result: Iterant[F, A] = null
result = NextBatch(Batch(elems: _*), F.delay(result), F.unit)
result
```
(the trick is not mine, [here's](https://japgolly.blogspot.com.by/2017/12/practical-awesome-recursion-ch-02.html) where I found it)

monix-tail/shared/src/main/scala/monix/tail/internal/IterantRepeat.scala
```diff
@@ -0,0 +1,57 @@
...
+        case NextBatch(gen, rest, stop) =>
+          NextBatch(gen, rest.map(loop), stop)
+        case Suspend(rest, stop) =>
+          Suspend[F, A](rest.map(loop), stop)
+        case Last(item) =>
+          Next[F, A](item, F.delay(loop(Iterant.empty)), F.unit)
+        case Halt(Some(ex)) =>
+          signalError(self, ex)
+        case Halt(None) =>
+          Suspend(F.delay(loop(source)), self.earlyStop)
```
oleg-py: If `self` is `Halt(_)`, then `self.earlyStop` is going to be `F.unit` anyway

monix-tail/shared/src/main/scala/monix/tail/internal/IterantRepeat.scala
```diff
@@ -0,0 +1,57 @@
...
+        case Next(head, tail, stop) =>
+          Next[F, A](head, tail.map(loop), stop)
+        case NextCursor(cursor, rest, stop) =>
+          NextCursor[F, A](cursor, rest.map(loop), stop)
+        case NextBatch(gen, rest, stop) =>
+          NextBatch(gen, rest.map(loop), stop)
+        case Suspend(rest, stop) =>
+          Suspend[F, A](rest.map(loop), stop)
+        case Last(item) =>
+          Next[F, A](item, F.delay(loop(Iterant.empty)), F.unit)
```
oleg-py: Can it be `loop(source)` directly instead of `loop(Iterant.empty)`?

Avasil: :+1

monix-tail/shared/src/main/scala/monix/tail/internal/IterantRepeat.scala
Avasil: :+1:

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
oleg-py: It will be desugared to something along the lines of
```scala
val result: ObjectRef[Iterant[F, A]] = ObjectRef.create(null)
// delay accepts by-name parameter, which is Function0 at runtime
result.elem = Next[F, A](elem, F.delay(() => result.elem), F.unit)
result.elem
```

The null will be observable if, and only if by-name parameter passed to delay evaluates *before* the assignment happens on the same line. It will happen if you're using `pure`, which is strict in `cats`, but for `delay` this would be a violation of `unsequencedDelayIsNoop` law from `Sync`, so it cannot happen (anything goes if the instance is not lawful, no need to support this).

In the article, the assignment happens to function, for which body, too, will never be evaluated until the function called, and the call only happens after the variable was assigned.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantRepeat.scala
```diff
@@ -0,0 +1,57 @@
...
+          signalError(self, ex)
+        case Halt(None) =>
+          Suspend(F.delay(loop(source)), F.unit)
+      } catch {
+        case ex if NonFatal(ex) => signalError(source, ex)
+      }
+
+    source match {
+      // terminate if the source is empty
+      case empty @ Halt(_) => empty
```
alexandru: So this is good, however there's a catch.

If you'll take a look at [RepeatSourceObservable](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/RepeatSourceObservable.scala), it has an `isEmpty: Boolean` used to track if the source emitted anything or not.

And this check right here is not sufficient. For one because you could be dealing with a `Suspend(Halt(None))`.
`NextCursor` is problematic because you could have a `NextCursor().repeat`. Now if the user does that, then fine, he probably knows what he's doing, but with no more elements to repeat, we need to stop the loop.

So I'm thinking about this

1. keep an `isEmpty` state in that loop
2.  only as long as `isEmpty == true`, when receiving a `NextCursor` we need to check if it `hasNext`, such that we set `isEmpty` accordingly, because we don't want to repeat a stream made of one empty `NextCursor`
3. only as long as `isEmpty == true`, we need to unwrap `NextBatch` into `NextCursor` such that we can check if it `hasNext` and set `isEmpty` accordingly
4. because we end up evaluating cursors, this means that in this match right here we need to `Suspend` in case we have `NextBatch` or `NextCursor`

Avasil: @alexandru 
Ok, I've added handling cases like ```Suspend(Halt(None))``` and ```NextCursor(BatchCursor(), F(BatchCursor(1, 2, 3)))```. There is ugly nested `if` in ```case NextBatch(_, _, _)```

monix-tail/shared/src/main/scala/monix/tail/internal/IterantRepeat.scala
```diff
@@ -0,0 +1,75 @@
...
+          else
+            Suspend(rest.map(loop(isEmpty)), stop)
+
+        case NextBatch(gen, rest, stop) =>
+          if(isEmpty){
+            val cursor = gen.cursor()
+            if(cursor.hasNext()) NextCursor[F, A](cursor, rest.map(loop(isEmpty = false)), stop)
+            else Suspend(rest.map(loop(isEmpty = true)), stop)
+          } else
+            NextBatch[F, A](gen, rest.map(loop(isEmpty = false)), stop)
```
alexandru: Sorry to be annoying about formatting 😀 but also wrap this `else` in curly brackets and add spaces in `if(` here and in the `NextCursor` branch too.

I think we already have the scalafmt configuration on master, install the IntelliJ IDEA plugin and work with it for this file. Hopefully that configuration can be part of the build process soon.

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -2125,6 +2132,24 @@ object Iterant extends IterantInstances {
...
+    case Seq(elem) =>
+      // trick to optimize recursion, see Optimisation section at:
+      // https://japgolly.blogspot.com.by/2017/12/practical-awesome-recursion-ch-02.html
+      var result: Iterant[F, A] = null
+      result = Next[F, A](elem, F.delay(result), F.unit)
+      result
+    case _ =>
+      var result: Iterant[F, A] = null
+      result = NextBatch(Batch(elems: _*), F.delay(result), F.unit)
+      result
```
alexandru: This is a really nice trick @oleg-py, thanks for pointing it out, it's the first time I'm seeing it.

PR #576, opened 2018-01-28
Clarify comment copied from Akka: this is Monix

ms-ati:
Replace "Akka" with "Monix" in the InternalApi comment. I'm under the
presumption that this was intended to be changed, but was missed.


alexandru:
Good catch, sorry — I don't usually copy/paste like that, I imported `@InternalApi` after this [discussion on Scala contributors](https://contributors.scala-lang.org/t/make-a-java-class-package-private-for-scala/1431).

PR #579, opened 2018-01-29
Iterant.reduce and headOptionL error handling

### LINKED ISSUE #577: Not all Iterant left folds handle errors from F[_] context yet
Some unfinished business is left from #563:
* `Iterant.reduceL`
* `Iterant.headOptionL`

(maybe I missed something)

These functions too need to handle error from `F[_]` context

@Avasil maybe you could take a look at those?

Avasil:
Fixes #577.

That leaves us with:
```scala
final def skipSuspendL(implicit F: Sync[F]): F[Iterant[F, A]]
final def foldRightL[B](b: F[B])(f: (A, F[B], F[Unit]) => F[B])(implicit F: Sync[F]): F[B]
```

We leave `foldRightL` as it is, right?


alexandru:
From what I remember for `foldRightL` the user is in control, including calling `stop` when the recursion stops, `foldRightL` describing a loop with that function parameter.

There's nothing we can do because we are not in charge of evaluating `earlyStop`.

PR #581, opened 2018-01-31
Handle broken nodes in Iterant.skipSuspendL

### LINKED ISSUE #559: Iterant error handling is leaky

oleg-py:
Closes #559.

Per my testing, `skipSuspendL` too, requires error handling as it has been done for other left folds. Otherwise you could end up with leaked resources due to `earlyStop` not being called on broken nodes (the unit test I added was failing before).

Unlike other left folds, `skipSuspendL` returns an `Iterant`, and so is free to report errors both inside `F` and `Iterant` context. Existing code was doing the latter on broken batches / cursors, so I went for consistency and made it handle broken nodes in this way as well (and I don't think anybody would appreciate some errors being inside `F` and others being reported as `Halt`).

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantSkipSuspend.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantSkipSuspend.scala
@@ -23,37 +23,43 @@ import monix.execution.misc.NonFatal
 import monix.tail.Iterant
 import monix.tail.Iterant.{Halt, Last, Next, NextBatch, NextCursor, Suspend}
 
+import scala.runtime.ObjectRef
+
 private[tail] object IterantSkipSuspend {
   /**
     * Implementation for `Iterant#skipSuspendL`
     */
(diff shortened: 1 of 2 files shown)


alexandru:
Looks good.

Issue #438, opened 2017-09-22
Java 9 Support


alexandru:
We need to provide support for Java 9 and this means:

1. conversions from `org.reactivestreams` to the new `Flow` interfaces, which are actually 1:1 translations
2. see if changes are required for our `Atomic` references, as we are using `Unsafe` access and it might have been deprecated (TBD)

alexandru:
Not doing anything else for Java 9 / Java 11 at this point.

We'll have to see about VarHandles, but we actually have bigger problems:
[JCTools](https://github.com/JCTools/JCTools) has to be upgraded first.

PR #584, opened 2018-02-01
Add conversions module for Java 8 CompletableFuture

### LINKED ISSUE #221: Java 8 compatibility: Add conversions for Completable Future
I'm thinking `Task.fromCompletableFuture` and `def runAsyncAsJava: CompletableFuture`, both being pretty similar to the existing `Task.fromFuture` and `def runAsyncForJava: Future`. Not sure what the best way to add new methods is without cluttering up `Task` and maintaining Scala.js compatibility, though.

oleg-py:
Closes #221 

I could not build `CancelableFuture -> CompletableFuture` because cancelled `CancelableFuture` is non-terminating, which I cannot react on without something like `Task.onCancelRaiseError`. I went with `Future -> CompletableFuture` conversion instead.

There are also APIs I don't think we can support, like `obtrudeValue` which lets you mutate the result of `CompletableFuture` even when it is already completed. Luckily, even the doc for `obtrudeValue` says it guarantees nothing.

Also, a lot of `CompletableFuture` API allows supplying `Executor` to run on (and if you don't, it runs on synchronously or on a common `ForkJoinPool`) - and since `Scheduler` is an `Executor` too, I added `deferCompletableFutureAction` to `Task` companion.

Btw, although @alexandru said:
> Given that Scala 2.12 requires Java 8, this should be added for Scala 2.12 only.

It should be possible to cross-build it for Scala 2.11.x too, since it is a separate module and there might be somebody running 2.11 on Java 8.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -270,6 +270,12 @@ lazy val testSettings = Seq(
   )
 )
 
+lazy val java8Settings = sharedSettings ++ testSettings ++ Seq(
+  name := "monix-java8",
+  // Support only Scala 2.12+
+  crossScalaVersions ~= { _.filterNot(_ startsWith "2.11") }
+)
+
 lazy val scalaJSSettings = Seq(
   coverageExcludedFiles := ".*"
 )
@@ -393,6 +399,11 @@ lazy val reactiveJS = project.in(file("monix-reactive/js"))
   .dependsOn(executionJS, evalJS % "compile->compile; test->test")
   .settings(reactiveCommon)
   .settings(scalaJSSettings)
+
+lazy val java8JVM = project.in(file("monix-java8"))
+  .configure(profile)
+  .dependsOn(evalJVM % "compile->compile; test->test")
+  .settings(java8Settings)
 
 lazy val reactiveTests = project.in(file("reactiveTests"))
   .configure(profile)
--- a/build.sbt
+++ b/build.sbt
@@ -270,8 +270,8 @@ lazy val testSettings = Seq(
   )
 )
 
-lazy val java8Settings = sharedSettings ++ testSettings ++ Seq(
-  name := "monix-java8",
+lazy val javaExtensionsSettings = sharedSettings ++ testSettings ++ Seq(
+  name := "monix-java",
   // Support only Scala 2.12+
   crossScalaVersions ~= { _.filterNot(_ startsWith "2.11") }
 )
@@ -303,8 +303,8 @@ lazy val monix = project.in(file("."))
 
 lazy val coreJVM = project.in(file("monix/jvm"))
   .configure(profile)
-  .dependsOn(executionJVM, evalJVM, tailJVM, reactiveJVM)
-  .aggregate(executionJVM, evalJVM, tailJVM, reactiveJVM)
+  .dependsOn(executionJVM, evalJVM, tailJVM, reactiveJVM, javaJVM)
+  .aggregate(executionJVM, evalJVM, tailJVM, reactiveJVM, javaJVM)
   .settings(crossSettings)
   .settings(name := "monix")
 
@@ -400,10 +400,11 @@ lazy val reactiveJS = project.in(file("monix-reactive/js"))
   .settings(reactiveCommon)
   .settings(scalaJSSettings)
 
-lazy val java8JVM = project.in(file("monix-java8"))
+lazy val javaJVM = project.in(file("monix-java"))
   .configure(profile)
-  .dependsOn(evalJVM % "compile->compile; test->test")
-  .settings(java8Settings)
+  .dependsOn(executionJVM % "provided->compile; test->test")
+  .dependsOn(evalJVM % "provided->compile; test->test")
+  .settings(javaExtensionsSettings)
 
 lazy val reactiveTests = project.in(file("reactiveTests"))
   .configure(profile)
--- a/build.sbt
+++ b/build.sbt
@@ -271,9 +271,7 @@ lazy val testSettings = Seq(
 )
 
 lazy val javaExtensionsSettings = sharedSettings ++ testSettings ++ Seq(
-  name := "monix-java",
-  // Support only Scala 2.12+
-  crossScalaVersions ~= { _.filterNot(_ startsWith "2.11") }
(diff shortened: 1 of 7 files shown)


alexandru:
We can describe `onCancelRaiseError` for `CancelableFuture` as well and we should probably add it:

```scala
import monix.execution.atomic.Atomic
import monix.execution.schedulers.TrampolineExecutionContext.immediate

def onCancelRaiseError[A](fa: CancelableFuture[A], e: Throwable)
  (implicit ec: ExecutionContext): CancelableFuture[A] = {

  if (fa.isComplete) fa else
    CancelableFuture.async { cb =>
      val active = Atomic(true)
      val cb2 = new RaiseCancelable(active, cb)
      // Register listener
      fa.underlying.onComplete(cb2)(immediate)
      // Return cancelable
      new RaiseCancelable(active, cb)
    }
}

final class RaiseCallback[A](active: AtomicBoolean, cb: Try[A] => Unit)
  (implicit ec: ExecutionContext) extends (Try[A] => Unit) {
  
  def apply(value: Try[A]) =
    if (active.getAndSet(false)) {
      cb(value)
    } else value match {
      case Failure(e) => ec.reportFailure(e)
      case _ => ()
    }
}

final class RaiseCancelable[A](active: AtomicBoolean, cb: Try[A] => Unit) 
  extends Cancelable {

  def cancel(): Unit = {
    if (active.getAndSet(false)) 
      cb(Failure(new CancelledException()))
  }
}
```

The nice thing about `onCancelRaiseError` is that this kind of synchronization is precisely what would have been needed if the contract would have been for `CancelableError` to always terminate in error on cancel, but this one is optional, on a by-need basis.

alexandru:
So nice of you to pick this up 😀

Some ideas:

1) I'd rather have this project named `monix-java`  with the `monix.java` namespace, because the current Java is Java 9 and who knows, we might end up with Java 9 conversions when Java 10 is released and I'd rather not have multiple Java-related projects. We'll be conservative and mindful in what we accept too.

2) I'd rather have its dependencies optional, as we might want to add conversions for `monix.reactive` as well - so each sub-project gets its own object / module under `monix.java`:

```scala
// For CancelableFuture
import monix.java.execution._

// For Task
import monix.java.eval._
```

And then in this sub-project's dependencies we can specify them as `"provided"`.  Not entirely sure if this is a good idea, since it can blow up in people's faces with a `ClassNotFoundException`, but we should include it in the `monix` package that brings in everything, anyway and `"provided"` is useful for people that want an a la carte experience.

oleg-py:
> `onCancelRaiseError`

That will only work if you call `cancel` on a result of `CancelableFuture.onCancelRaiseError`. But since Futures represent a computation that is already running, there's not much utility in this unless it propagates cancelled status downstream. Cancelled monix futures become non-terminating, so they propagate nothing. I cannot turn cancelled monix future into cancelled `j.u.c.CompletableFuture`.

So it's might be better to leave that as `Future -> j.u.c.CompletableFuture` that turns non-terminating `Future`s into non-terminating `j.u.c.CompletableFuture`s - with no support for cancelling.

> `monix.java` namespace

I'm slightly :-1: on that. First, if somebody does `import monix._`, that will shadow `java` package. Second, if we do add Java 9 conversions (e.g. to `Flow` interfaces, #438), it will be harder for users to understand why that `ClassNotFoundException` happens. I believe different namespaces (`monix.java8`, `monix.java9`) would be more convenient, even if it lives in the same module.

oleg-py:
Hm, I cannot find a way to conditionally aggregate the subproject if `scalaVersion` used in build is not included in its `crossScalaVersions`. Is there a way?

alexandru:
@oleg-py unfortunately you can't pick and choose the Scala versions per sub-project.

When I first wrote that description on the issue, I was thinking of a module inside `monix-eval` to only be included with Scala 2.12, not in a separate sub-project, but by using a source trick. Like we have for example here: https://github.com/monix/monix/tree/master/monix-execution/shared/src/main/scala_2.12/

This would be another way of approaching the problem. We could make `object CancelableFuture` extend a trait that's empty for Scala 2.11, but that contains extra utilities for Scala 2.12+.

On the `monix.java` namespace I agree with you, bad idea.

oleg-py:
@alexandru I made it work for Scala 2.11 / Java 8. Seemed easier 😅 (esp. since CI uses that combination for builds too). WDYT?

monix-java/src/main/scala/monix/java8/execution/package.scala
```diff
@@ -0,0 +1,80 @@
...
+  * Provides methods to convert between `scala.concurrent.Future`
+  * and `java.util.concurrent.CompletableFuture`
+  */
+package object execution {
+
+  implicit class JavaCompletableFutureUtils[A](val source: CompletableFuture[A]) extends AnyVal {
+
+    /** Convert `CompletableFuture` to [[monix.execution.CancelableFuture]]
+      *
+      * If the source is cancelled, returned `Future` will never terminate
```
alexandru: Note a `.onCancelRaiseError` op for `CancellableFuture` would work for non-terminating futures, even if the source is already cancelled 😉

build.sbt
```diff
@@ -394,6 +398,12 @@ lazy val reactiveJS = project.in(file("monix-reactive/js"))
   .settings(reactiveCommon)
   .settings(scalaJSSettings)
 
+lazy val javaJVM = project.in(file("monix-java"))
+  .configure(profile)
+  .dependsOn(executionJVM % "provided->compile; test->test")
+  .dependsOn(evalJVM % "provided->compile; test->test")
```
alexandru: Hope this works 😀

alexandru:
Merged. Hope it works well.

Thanks.

fancywriter:
@alexandru @oleg-py 
Hi guys! It may be silly question, but what dependency do I have to add to my project to make `monix.java.*` package available? I didn't find it in README.md...

*monix-java* of last version 2.3.3 can't be resolved...

alexandru:
Sorry guys for procrastinating on releasing `3.0.0-M4`. The release process is really annoying 😑

It's ready for release though, will do so sometime today.

PR #585, opened 2018-02-01
make onErrorFallbackTo take fallback by-name

jozic:
this way task is evaluated only on error

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -1284,7 +1284,7 @@ sealed abstract class Task[+A] extends Serializable {
   /** Creates a new task that in case of error will fallback to the
     * given backup task.
     */
-  final def onErrorFallbackTo[B >: A](that: Task[B]): Task[B] =
+  final def onErrorFallbackTo[B >: A](that: => Task[B]): Task[B] =
     onErrorHandleWith(_ => that)
 
   /** Given a predicate function, keep retrying the
--- a/monix-eval/shared/src/test/scala/monix/eval/TaskErrorSuite.scala
+++ b/monix-eval/shared/src/test/scala/monix/eval/TaskErrorSuite.scala
@@ -225,6 +225,15 @@ object TaskErrorSuite extends BaseTestSuite {
     assertEquals(f.value, Some(Failure(err)))
   }
 
+  test("Task.onErrorFallbackTo should take fallback by-name") { implicit s =>
+    var effect = 0
+    val task = Task(2).onErrorFallbackTo(Task.now{ effect = 5; effect })
+    val f = task.runAsync
+    s.tick()
+    assertEquals(f.value, Some(Success(2)))
+    assertEquals(effect, 0)
+  }
+
   test("Task.onErrorFallbackTo should be cancelable if the ExecutionModel allows it") { implicit s =>
     def recursive(): Task[Int] = {
       Task[Int](throw DummyException("dummy")).onErrorFallbackTo(Task.defer(recursive()))


Avasil:
Hi @jozic,
Thanks but I don't think it should be by-name. If you change your test to:
``` scala
Task(2).onErrorFallbackTo(Task.eval { effect = 5; effect })
```
or
``` scala
Task(2).onErrorFallbackTo(Task { effect = 5; effect })
```
It passes without your changes. That's because `Task` suspends side effects by design and it shouldn't matter whether you take it as a strict value or by-name because there won't happen any evaluation until you run it.

One of the exceptions is `Task.now` (and `Task.pure` ) which takes strict value, it is equivalent of `Future.successful` so you shouldn't pass any side-effects there if you care about referential transparency because everything will happen straight away.

If you write:
```scala
val fallbackTask = Task.now { effect = 5; effect }
val task = Task(2).onErrorFallbackTo(fallbackTask)
```
The side-effect will happen regardless of the change because that's how `Task.now` works.

I hope my explanation is clear!

jozic:
Hi @Avasil , i'm aware of `eval` and `apply`, that's why my test explicitly uses `now`. 
Could you please explain why you think by-name is bad here? Thanks

Avasil:
@jozic 
Sure, hopefully it will be clearer now!

I just think it doesn't really do anything. 
This evaluation happens not because of `onErrorFallbackTo` but because of `Task.now` and that's expected behavior.
Do you have any use case where you pass side effects in `Task.now`? If so, that's not correct. 
Think what it means that it takes strict value, evaluation of `Task.now` will happen like that:
```scala
val missiles = launchMissiles() // missiles launched
val task = Task(missiles) // task created with already evaluated missiles
task.runAsync // doesn't really do anything interesting because missiles were already launched at first line
```
other than
```scala
val task = Task { launchMissiles() } // task created, nothing happens
task.runAsync // starts execution, launching missiles
```
Additionaly, consider this:
```scala
  test("Task.onErrorFallbackTo should take fallback by-name") { implicit s =>
    var effect = 0
    val fallbackTask = Task.now { effect = 5; effect }
    val task = Task(2).onErrorFallbackTo(fallbackTask)
    val f = task.runAsync
    s.tick()
    assertEquals(f.value, Some(Success(2)))
    assertEquals(effect, 0)
  }
```
Can any reasonable implementation of `onErrorFallbackTo` pass the test above?

alexandru:
@jozic thanks for your pull request.

Here's a more high-level take on the problem - data types like `Task` and `Coeval` are basically a replacement for by-name parameters and so usually you pick between one or the other, but not both.

`Task` allows you to pick the evaluation model — you can have it be strict or lazy and that's a feature.

By introducing a by-name parameter in `onErrorFallbackTo`, it's not only unnecessary, but fairly error prone, because users are expected to know the difference between `Task.now` and `Task.eval` and when you shove a `Task.now` in `onErrorFallbackTo`, then it has to have strict behavior otherwise users can be caught off-guard.

In the API there are instances of by-name parameters used. For example in `Task.suspend`, but that's an instance where laziness is an explicit purpose of the function and one of Task's builders, part of the FFI so to speak.

alexandru:
Btw, a by-name parameter in this function would have made sense for Scala's `Future`, because `Future` isn't lazy.

jozic:
@Avasil thanks for your explanation, but I guess I wasn't clear enough. I have no issues understanding lazy nature of `eval` and eager one for `now`. You say 
> I just think it doesn't really do anything.

but it does *something* as the test proves. Whether this is desired effect or not that's another question. I though this won't harm, according to @alexandru this is not desired and error-prone. 
My question was  `why is it not desired`. 

anyway, i can definitely live without it, thanks for discussion

PR #587, opened 2018-02-08
POC: Make Observable.scan behave like stdlib scanLeft

### LINKED ISSUE #522: scan of Observable and Iterant are not consistent with stdlib
Standard library `scan` yields initial element at the beginning of resulting collection:

```scala
// List(0, 1, 2, 3)
List(1, 1, 1).scan(0)(_ + _)
```

While in monix, neither `Observable` nor `Iterant` do:

```scala
import monix.eval.Coeval
import monix.tail.Iterant

// List(1, 2, 3)
Iterant[Coeval].of(1, 1, 1)
  .scan(0)(_ + _)
  .toListL
  .value
```
and
```scala

import scala.concurrent.Await
import scala.concurrent.duration.Duration

val r = Observable(1, 1, 1)
  .scan(0)(_ + _)
  .toListL

// List(1, 2, 3)
Await.result(r.runAsync, Duration.Inf)
```

I'm not sure if it is a bug or intentional, but IMO it is at least worth being explicitly documented.

oleg-py:
My attempt to fix #522. Currently lacks `scanTask` and `Iterant` methods, but I want to get some feedback on the approach.

With `AsyncSemaphore` it turned out to be pretty straightforward conversion (although I only discovered it after 2 somewhat messy attempts :). Some places might benefit from `TrampolineEC`, but I'm not too sure where it's safe to put it.

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ScanObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ScanObservable.scala
@@ -17,8 +17,8 @@
 
 package monix.reactive.internal.operators
 
-import monix.execution.Ack.Stop
-import monix.execution.misc.NonFatal
+import monix.execution.Ack._
+import monix.execution.misc.{AsyncSemaphore, NonFatal}
 import monix.execution.{Ack, Cancelable}
 import monix.reactive.Observable
 import monix.reactive.observers.Subscriber
@@ -40,24 +40,33 @@ class ScanObservable[A,R](
       source.unsafeSubscribeFn(
         new Subscriber[A] {
           implicit val scheduler = out.scheduler
-          private[this] var isDone = false
+          @volatile private[this] var isDone = false
           private[this] var state = initialState
+          private[this] val semaphore = AsyncSemaphore(1)
+          // Signal initial state immediately
+          private[this] var lastAck = out.onNext(state)
 
-          def onNext(elem: A): Future[Ack] = {
-            // Protects calls to user code from within the operator and
-            // stream the error downstream if it happens, but if the
-            // error happens because of calls to `onNext` or other
-            // protocol calls, then the behavior should be undefined.
-            var streamError = true
-            try {
-              state = f(state, elem)
-              streamError = false
-              out.onNext(state)
-            }
-            catch {
-              case NonFatal(ex) if streamError =>
-                onError(ex)
+          def onNext(elem: A): Future[Ack] = semaphore.greenLight { () =>
+            lastAck.syncFlatMap {
+              case Stop =>
                 Stop
+
+              case Continue =>
+                // Protects calls to user code from within the operator and
+                // stream the error downstream if it happens, but if the
+                // error happens because of calls to `onNext` or other
+                // protocol calls, then the behavior should be undefined.
+                var streamError = true
+                try {
+                  state = f(state, elem)
+                  streamError = false
+                  lastAck = out.onNext(state)
+                  lastAck
+                } catch {
+                  case NonFatal(ex) if streamError =>
+                    onError(ex)
+                    Stop
+                }
             }
           }
 
@@ -68,9 +77,13 @@ class ScanObservable[A,R](
             }
 
           def onComplete(): Unit =
-            if (!isDone) {
-              isDone = true
-              out.onComplete()
+            // `onNext` might require error reporting, so we need for it
+            // to actually finish before we can try to forward `onComplete`
+            semaphore.awaitAllReleased().foreach { _ =>
+              if (!isDone) {
+                isDone = true
+                out.onComplete()
+              }
             }
         })
     }
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ScanObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ScanObservable.scala
@@ -23,7 +23,7 @@ import monix.execution.{Ack, Cancelable}
 import monix.reactive.Observable
 import monix.reactive.observers.Subscriber
 
-import scala.concurrent.{Future, Promise}
+import scala.concurrent.Future
 
 private[reactive] final
 class ScanObservable[A,R](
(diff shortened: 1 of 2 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ScanObservable.scala
```diff
@@ -40,24 +40,33 @@ class ScanObservable[A,R](
...
-            var streamError = true
-            try {
-              state = f(state, elem)
-              streamError = false
-              out.onNext(state)
-            }
-            catch {
-              case NonFatal(ex) if streamError =>
-                onError(ex)
+          def onNext(elem: A): Future[Ack] = semaphore.greenLight { () =>
```
alexandru: You don't need a semaphore, because by the time `source.unsafeSubscribeFn` actually happens, the first `onNext(state)` was already sent, so you only need to take care of back-pressuring, therefore `lastAck.syncFlatMap` should be all that's necessary.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ScanObservable.scala
```diff
@@ -68,9 +77,13 @@ class ScanObservable[A,R](
             }
 
           def onComplete(): Unit =
-            if (!isDone) {
-              isDone = true
-              out.onComplete()
+            // `onNext` might require error reporting, so we need for it
+            // to actually finish before we can try to forward `onComplete`
+            semaphore.awaitAllReleased().foreach { _ =>
+              if (!isDone) {
```
alexandru: Here you actually need to back-pressure on `lastAck` unfortunately. It is true that `onComplete` does not need to wait for the final `onNext`, but it does need to wait for the second to last `onNext`.

So picture this situation:

```scala
Observable.now(1).scan(0)(_ + _)
```

What will happen is this:

```scala
lastAck = out.onNext(0)
lastAck.syncFlatMap { case Continue => out.onNext(1) }
out.onComplete()
```

Now if you really think about it this way, notice that `onNext(1)` can be concurrent with `onComplete` (if the downstream `Observer` returns a delayed `Future`) and that should never happen. 

Also you don't need the semaphore here either.

oleg-py:
@alexandru

I refactored the implementation. It passes tests if both `onNext` and `onComplete` use `sync...` methods (i.e. it's not possible to replace `syncFlatMap` with `flatMap` in `onNext`). So I'm not sure if it's good or just an artifact of how test scheduler is implemented.

> notice how this complicates the implementation

Yes. Original `scan` is 1-to-1 transformation of the original stream, just like `map`. Emitting additional elements is what makes it more complicated. But it's not too complex, I would say.

> lets introduce a scan0 alternative that also streams the first element.

Personally, I don't like `scan0` as a name. I think having it as `scan` (like RxScala) or `scanLeft` (like standard collections) would improve discoverability. For the latter route, it is possible to deprecate current `scan` and add an overload that is consistent with collections' and RxScala's one (`def scan[U >: T](accumulator: (U, U) ⇒ U): Observable[U]`). I still remember having learned Scala collections and RxScala, and then, after switching to Monix, I spent two hours debugging before I realized that initial element is never emitted.

I, too, am not fond of having code broken silently. As a partial solution, it's possible to introduce Scalafix rewrite to help users automatically migrate to another version. This also might make sense for other breaking changes (e.g. add unsafe constructor to `TaskSemaphore` and rewrite `TaskSemaphore()` to `TaskSemaphore.unsafeCreate()`, rewrite all deprecated `fork` methods, etc.). But there's no good way to avoid breaking things and get what I propose.

I'm totally fine if you decide to leave `scan` as it is. It's harder for my use case, however, than to *just prepend the initial value* if `initialValue` is allowed to have side effects, or for `mapEval` and `mapTask` (since initial values for those have to be evaluated once, too). For other way it's always `.drop(1)` and that's it. Would be nice if we could ask users directly.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ScanObservable.scala
```diff
@@ -42,22 +42,31 @@ class ScanObservable[A,R](
...
-              out.onNext(state)
-            }
-            catch {
-              case NonFatal(ex) if streamError =>
-                onError(ex)
+            lastAck.syncFlatMap {
+              case Stop =>
                 Stop
+
+              case Continue =>
```
alexandru: Please fix the indentation here.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ScanObservable.scala
```diff
@@ -42,22 +42,31 @@ class ScanObservable[A,R](
           implicit val scheduler = out.scheduler
           private[this] var isDone = false
           private[this] var state = initialState
+          // Signal initial state immediately
+          private[this] var lastAck = out.onNext(state)
```
alexandru: You don't need to implement back-pressure on each `onNext`. This shouldn't be a `var`, because it's only the first `onNext` that's out of place:

```scala
private[this] val firstAck = out.onNext(state)
```

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ScanObservable.scala
```diff
@@ -68,9 +77,13 @@ class ScanObservable[A,R](
             }
 
           def onComplete(): Unit =
-            if (!isDone) {
-              isDone = true
-              out.onComplete()
+            // `onNext` might require error reporting, so we need for it
+            // to actually finish before we can try to forward `onComplete`
+            lastAck.syncOnComplete { _ =>
```
alexandru: Here you need to do `firstAck.syncOnContinue`.

You also need the same logic in `onError`, because otherwise the last `onNext` streamed can become concurrent with the `onError` signal and that's against the protocol.

alexandru:
@oleg-py Can we add this as a `scan0`?

I know you don't like it, but I fear breaking people's code silently and we can't do anything to prevent it by keeping the current name; Scalafix is cool, but it's for power users that read release announcements carefully, which I think are a minority; plus I'm pretty sure `scan` usage is pervasive; I abused `scan` a lot in my own code at least and given a lot of advice for its usage.

alexandru:
Plus `scan0` really isn't terrible. I find it very descriptive, while being still short 🙂

oleg-py:
@alexandru it's possible to implement `scan0` in terms of existing `scan` and a few other operators:

    Observable.evalOnce(seed).flatMap(s => s +: Observable.scan(s)(op))

Similar implementation is also possible for `scanEval0` and `scanTask0`. If we're not going to remove existing implementations, it probably makes sense to piggy-back on them. WDYT?

alexandru:
👍  sounds good

oleg-py:
Cool. I'll put up a new PR soon, then

PR #588, opened 2018-02-08
Update headers in 2.11 sources + add myself to AUTHORS

oleg-py:
Few things I forgot:
* When doing #531, I ran `headerCreate` without cross-building, so sources specific to 2.11 didn't get updated
* I also never added myself to AUTHORS file :)


alexandru:
Thanks @oleg-py

Issue #580, opened 2018-01-31
Questionable Iterant ScalaCheck checks w.r.t. exceptions


oleg-py:
Exhibit A:
```scala
  test("All exceptions are born equal") { implicit s =>
    check1 { (iter: Iterant[Coeval, Int]) =>
      val dummy = DummyException("dummy")
      val tail = Iterant[Coeval].raiseError[Int](dummy)
      (iter ++ tail) <-> tail
    }
  }
```
This is only true for Coeval. Unlike its other effectful comrades, `IsEqIterantCoeval` does not use `equalityThrowable` instance to compare errors, it just checks if the both results are a failure.

---

Exhibit B:
```scala
  test("Prefixes matter, but there's an Exception") { implicit s =>
    check1 { (iter: Iterant[IO, Int]) =>
      val dummy = DummyException("dummy")
      val tail = Iterant[IO].raiseError[Int](dummy)
      (iter.onErrorIgnore ++ tail) <-> tail
    }
  }
```

For iterants that end with an error, only error is checked. This is true for `IO`, `Task` and `Coeval`.

This means that you can have iterants `a, b` and a pure function `f` such that
* `a <-> b` holds, and
* `f(a) <-> f(b)` does not

For example:
```scala
val dummy = DummyException("dummy")
val prefix = Iterant[Coeval].of(1, 2, 3, 4, 5)
val tail = Iterant[Coeval].raiseError[Int](dummy)
(prefix ++ tail) <-> tail // holds, as shown above
// (prefix ++ tail).take(3) <-> tail.take(3)  // does not hold: left stream is pure-ish, right one throws
```

alexandru:
I think our problems here come from the fact that we are using `.toListL` to compare the streams in the `Eq` instance, so any two streams ending in error will be equal regardless of prefix, plus we have the issue with `Task` vs `Coeval`.

We should probably expose the errors and compare via `.attempt.toListL`.

Also, on comparing errors, I think we should stop comparing `Throwable` instances, because they can be easily wrapped — if both Coevals or Tasks or IOs fail, then they should be considered equivalent, regardless of the actual error thrown.

PR #589, opened 2018-02-14
Improve handling of broken batches/cursors in Iterant.attempt

oleg-py:
I was looking into #580 and I discovered few bits missing from my `Iterant.attempt` fix. In particular, it didn't handle broken batches and cursors correctly. All unit tests added here for `attempt` were failing.


alexandru:
Ah, nice, thanks @oleg-py

PR #596, opened 2018-02-21
Add Fiber interface, refactor `memoize`

alexandru:
Changes:

- adds `Fiber` interface, to be consistent with [cats-effect](https://github.com/typelevel/cats-effect/pull/121)
- refactors `Coeval` and `Task` to no longer treat `memoize` as a special state in their ADT definition

`Fiber` gets used whenever a `Task` is attached to a result that will be available on the completion of an already running process. Resembles `CancellableFuture` in semantics, but gets used within the `Task` context, currently for:

1. `start`
2. `pairBoth`

This PR also refactors the `memoize` operation on both `Coeval` and `Task`. We are simplifying the ADTs, `memoize` no longer being represented as a special state. This simplifies the implementation of the run-loop, reducing the area for potential bugs.  Thus, internally, `Coeval.Once` and `Task.MemoizeSuspend` are gone.


monix-eval/shared/src/test/scala/monix/eval/TaskStartSuite.scala
```diff
@@ -26,12 +26,12 @@ import scala.util.Success
 object TaskStartSuite extends BaseTestSuite {
   test("task.start.flatMap(id) <-> task") { implicit sc =>
```
oleg-py: I think test descriptions need to be updated too 😏

monix-eval/shared/src/main/scala/monix/eval/internal/TaskMemoize.scala
```diff
@@ -0,0 +1,184 @@
...
+  private final class MemoFiber[A](
+    source: Task[A],
+    val cacheErrors: Boolean)
+    extends Fiber[A] {
+
+    var thunk: Task[A] = source
+    val state = Atomic(null : AnyRef)
+
+    def isCachingAll: Boolean =
+      cacheErrors
```
oleg-py: Isn't this getter redundant, as `cacheErrors` is already a public field?

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -545,6 +573,44 @@ sealed abstract class Task[+A] extends Serializable {
...
+    * $unsafeMemoize
+    *
+    * @see [[memoizeOnSuccess]] for a version that only caches
+    *     successful results
+    *
+    * @return a `Task` that can be used to wait for the memoized
+    *         value or to cancel the background process that calculates
+    *         the memoized result (as long as it can be cancelled)
+    */
+  final def memoize: Fiber[A] =
```
oleg-py: Excuse me if I'm totally wrong here

I don't really see how `Fiber` is a way to represent memoization. My intuition tells me that when I get a `Fiber[A]`, the evaluation process is already running, so my options are either waiting for its completion or telling it that it's no longer necessary (which might still continue running, but I'll never get a result). But implementation of `TaskMemoize` seems to imply the evaluation will only happen when I'll try to perform a `join`, so it doesn't really represent a _running background process_.

alexandru: You're totally right, haven't thought this through. I'm reverting the change.

alexandru:
Thanks for the review @oleg-py. Not modifying the `.memoize` signature.

alexandru:
@Avasil I forgot to mention.

Because of this PR I ended up reverting https://github.com/monix/monix/pull/555 unfortunately.

When we fork a `Task` now we get a `Fiber` back, which basically allows for cancellation.
This kind of makes it incompatible with `CoflatMap`, at least the signature.

We might get back to forking as the behavior for `CoflatMap`, but we'll have to wait to see what happens with the type class hierarchy in `cats-effect`, because we might end up with some `MonadFork` type class.

PR #598, opened 2018-02-24
Integrates cats.effect.Timer and IO.cancelable

alexandru:
In `cats-effect` I managed to push a new `Timer` data type, as a pure alternative to `Scheduler`: https://github.com/typelevel/cats-effect/pull/132

This is my last annoyance with Monix's API, hopefully the last breaking change before a final 3.0 release (in our own back yard so to speak, `cats-effect` being another story).

Summary of this PR:

1. change `Iterant.intervalAtFixedRate` and `Iterant.intervalWithFixedDelay` to be polymorphic via `Timer` (instead of specialized for `Task`)
2. change the `Iterant` builders, getting rid of `Coeval` or `Task` specializations
3. change `Scheduler` to also provide `clockRealTime` and `clockMonotonic` operations, instead of `currentTimeMillis`, which should address an older complaint; an extension method with `@deprecated` flag is still provided for source compatibility
4. add a `Task.timer: Timer[Task]` implicit and a `Task.timer(sc: Scheduler)` builder
5. add a `Scheduler#timer` extension method
6. add a `Cancelable#cancelIO` extension method and `Cancelable.fromIO` builder
7. `Task#toIO` to yield cancelable `IO` values

---

## 1. Iterant.intervalAtFixedRate and intervalWithFixedDelay

`Iterant.intervalAtFixedRate` and `Interant.intervalWithFixedDelay` have been defined specifically for `Task`, however this dependency on `Task` annoys me because the point of `Iterant` was to abstract over `F[_]` data types.

This changes the interval builders like so:

```scala
def intervalAtFixedRate[F[_]](period: FiniteDuration)
  (implicit F: Async[F], timer: Timer[F]): Iterant[F, Long]

def intervalAtFixedRate[F[_]](initialDelay: FiniteDuration, period: FiniteDuration)
  (implicit F: Async[F], timer: Timer[F]): Iterant[F, Long]

def intervalWithFixedDelay[F[_]](delay: FiniteDuration)
  (implicit F: Async[F], timer: Timer[F]): Iterant[F, Long]

def intervalWithFixedDelay[F[_]](initialDelay: FiniteDuration, delay: FiniteDuration)
  (implicit F: Async[F], timer: Timer[F]): Iterant[F, Long]
```

This means that we can now work with `cats.effect.IO` as well:

```scala
Iterant[IO].intervalAtFixedRate(3.seconds)
```

## 2. Iterant builders

As part of this ticket the `IterantBuilders` is expanded into a hierarchy that allows for an à la carte approach, as we now have to differentiate between `Sync[F]` and `Async[F]` data types, plus we allow `Applicative[F]` and simple builders to be curried as well. No more hard-coding of `Task` or `Coeval` builders, although we still have `From[Task]`, `From[Coeval]` and `From[IO]` for performance reasons.

Plus the `Iterant.CatsInstances` got simplified. We no longer allow the possibility of providing alternative `Task` or `Coeval` instances for Cats type-classes because type-classes are meant to be coherent.

## 3. Scheduler to provide clockRealTime and clockMonotonic

So first the `Scheduler` interface gets rid of `currentTimeMillis` and changes to this:

```scala
trait Scheduler {
  // ...

  def clockRealTime(unit: TimeUnit): Long

  def clockMonotonic(unit: TimeUnit): Long

}
```

These are the equivalents of `System.currentTimeMillis` and of `System.nanoTime` respectively. They also allow picking the precision of the output (even if the actual resolution will be implementation dependent).

These reflect the signature of their pure counterparts in the `Timer` interface that I pushed in `cats-effect`.

## 4. Task.timer

We can provide a `Task[Timer]` that's globally available out of the box.

```scala
import cats.effect.Timer
import monix.eval.Task

val timer = Timer[Task]
// Equivalent to:
val timer = Task.timer
```

Given a `Scheduler`, we can also compile a `Task[Timer]` that uses it, instead of the `Task`'s default `Scheduler`:

```scala
val scheduler = Scheduler.io()

val timer = Task.timer(scheduler)
```

This is an overload, because the default `Task[Timer]` instance does not take an implicit `Scheduler` 😉

## 5. Scheduler#timer extension method

`Scheduler` has a new extension method for your convenience for building `Timer` instances out of `IO`:

```scala
import cats.effect._

val scheduler = Scheduler.io()

val timer: Timer[IO] = scheduler.timer[IO]
```

This is not a method of `Scheduler`, but an extension method defined in `Scheduler.Extensions`.

## 6. Cancelable to and from IO[Unit] conversion

The new `IO.cancelable` needs an `IO[Unit]` in the passed function describing the cancellation logic.
Therefore it becomes useful to turn `Cancelable` values in `IO[Unit]` values and vice versa.

```scala
// Conversion to IO
val c = Cancelable(() => println("Cancelling!"))
val io: IO[Unit] = c.cancelIO

// Conversion from IO
val io = IO(println("Cancelling!"))
val c: Cancelable = Cancelable.fromIO(io)
```

Note that the execution of an `IO` value is not idempotent, but the `Cancelable.fromIO` conversion guarantees idempotency.

## 7. Task#toIO to yield cancelable IO values

Such `IO` values are now cancelable:

```scala
val io: IO[Unit] = Task.sleep(10.seconds).toIO
```


alexandru:
Decided that this PR will get merged after https://github.com/typelevel/cats-effect/pull/121 gets merged in `cats-effect`.

This is because I'm hoping we get a `Timer` in `cats-effect`, as painful as this wait may be, because I'm striving for stability here.

oleg-py:
@alexandru I'm +1 on waiting for `cats-effect`. If somebody needs to use these builders with `IO` as of now, they can do `.liftMap(_.to[IO], _.to[IO])`.

alexandru:
@oleg-py and now the PR is official: https://github.com/typelevel/cats-effect/pull/132

😏

alexandru:
This PR is unfortunately a little big now due to the refactoring of `Scheduler`.

But I'm personally happy with it because it fixes a long time complaint of `Scheduler` using `System.currentTimeMillis`.

@oleg-py @Avasil would like to know your thoughts.

N.B. this PR makes Monix depend on a hash version. Will have to wait on a final `0.10` release of `cats-effect` for this to be merged.

monix-execution/shared/src/main/scala/monix/execution/schedulers/TestScheduler.scala
```diff
@@ -15,53 +15,151 @@
...
-// Prevents conflict with the deprecated symbol
-import monix.execution.{ExecutionModel => ExecModel}
 
 import scala.annotation.tailrec
 import scala.collection.immutable.SortedSet
 import scala.concurrent.duration.{Duration, FiniteDuration, TimeUnit}
 import scala.util.Random
 
-/** A scheduler meant for testing purposes. */
+/** [[Scheduler]] implementation meant and a provider of
```
Avasil: I think it could be worded a bit better, maybe without "implementation meant" phrase?

monix-execution/shared/src/main/scala/monix/execution/schedulers/TestScheduler.scala
```diff
@@ -116,6 +217,45 @@ final class TestScheduler private (
...
+    * executing them all, until no tasks remain in the queue to execute.
+    *
+    * Order of execution isn't guaranteed, the queued `Runnable`s are
+    * being shuffled in order to simulate the needed non-determinism
+    * that happens with multi-threading.
+    *
+    * {{{
+    *   implicit val ec = TestScheduler()
+    *
+    *   val f = Future(1 + 1).flatMap(_ + 1)
```
Avasil:
```scala
val f = Future(1 + 1).flatMap(x => Future(x + 1))
```

monix-execution/shared/src/test/scala/monix/execution/CancelableSuite.scala
```diff
@@ -36,4 +41,88 @@ object CancelableSuite extends SimpleTestSuite {
...
+    val dummy = DummyException("dummy")
+
+    val c = Cancelable.collection((0 until 100).map(_ => Cancelable { () =>
+      effect += 1
+      throw dummy
+    }))
+
+    try {
+      c.cancel()
+      fail("c.cance() should throw")
```
Avasil: cancel

alexandru:
This is getting merged, because the next release of Monix is now synced with 0.10.

alexandru:
N.B. not everything that needs to happen for cats-effect 0.10 is in this PR.

We need to implement the new `Concurrent` type class for example. Will create another PR.
W00t, exciting!

alexandru:
Oh, the good news is that I fixed the `Sync` / `Async` tests for `Iterant` by making that param for testing stack safety configurable in cats-effect.

PR #600, opened 2018-02-25
Add Sync & Async instances for Iterant

### LINKED ISSUE #599: monix.tail.Iterant should have a cats.effect.Sync instance
`cats-effect` [version 0.9](https://github.com/typelevel/cats-effect/releases/tag/v0.9) is released and @oleg-py's changes to `SyncLaws` is in (issue [#118](https://github.com/typelevel/cats-effect/issues/118)).

This means that we should be able to implement a `Sync` instance for `monix.tail.Iterant`. And it would be cool to do so because we get extra validations for free.

oleg-py:
Closes #599 

I went ahead and implemented `Async` instance too (under the assumption that callback will only be invoked once, so it just delegates to underlying `F`). Couldn't do `Effect` though in a similar way, because signature of `runAsync` sort-of implies cardinality of 1.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,7 +10,7 @@ addCommandAlias("ci-js",      ";clean ;coreJS/test:compile  ;coreJS/test")
 addCommandAlias("release",    ";project monix ;+clean ;+package ;+publishSigned ;sonatypeReleaseAll")
 
 val catsVersion = "1.0.1"
-val catsEffectVersion = "0.8"
+val catsEffectVersion = "0.9"
 val jcToolsVersion = "2.1.1"
 val reactiveStreamsVersion = "1.0.2"
 val scalaTestVersion = "3.0.4"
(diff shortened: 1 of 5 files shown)


oleg-py:
@alexandru not sure why build fails, could you please take a look?

alexandru:
The build is timing out and it can do so in a piece of logic that wasn't timing out before, possibly due to the CPU being very busy (although I don't understand how, because I remember that I disabled testing in parallel, but those settings never worked properly anyway).

The tests for `AsyncLaws` and `SyncLaws` might break Travis. We might need to adjust the ScalaCheck parameters in those tests, to generate shorter `Iterant` values.

alexandru:
One problem are all the stack-safety checks in `SyncLaws`, which on Travis are a problem due to the limited CPU available.

alexandru:
The attempted fix was to disable them when running on top of Travis.

PR #602, opened 2018-02-25
TaskLocal should expose Local

### LINKED ISSUE #595: TaskLocal should take Local
Under the hood, a TaskLocal creates a Local instance to pass context around, which is great- because hopefully the average use case is wrapped inside of a Task anyway so TaskLocal is a sane API on top of an unsafe mutable one. Unfortunately my use case is to trace operations in APIs that are inherently synchronous (JDBC) and it would be nice if TaskLocal could share an existing Local so that tracing use cases don't have to do gymnastics to synchronize state between the two.

Avasil:
Fixes #595

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
@@ -95,7 +95,7 @@ import monix.execution.misc.Local
 final class TaskLocal[A] private (default: => A) {
   private[this] val ref = new Local(default)
 
-  /** Returns Local instance used in this [[TaskLocal]].
+  /** Returns [[monix.execution.misc.Local]] instance used in this [[TaskLocal]].
     *
     * Note that `TaskLocal.bind` will restore the original local value
     * on the thread where the `Task's` run-loop ends up so it might lead
--- a/monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
@@ -95,6 +95,14 @@ import monix.execution.misc.Local
 final class TaskLocal[A] private (default: => A) {
   private[this] val ref = new Local(default)
 
+  /** Returns Local instance used in this [[TaskLocal]].
+    *
+    * Note that `TaskLocal.bind` will restore the original local value
+    * on the thread where the `Task's` run-loop ends up so it might lead
+    * to leaving local modified in other thread.
+    */
+  def local: Task[Local[A]] = Task.eval(ref)
+
   /** Returns the current local value (in the `Task` context). */
   def read: Task[A] =
     Task.eval(ref.get)
@@ -116,7 +124,7 @@ final class TaskLocal[A] private (default: => A) {
     *   // Should yield 200 on execution, regardless of what value
     *   // we have in `local` at the time of evaluation
     *   val task = local.bind(100)(Task {
-    *     local.get * 2
+    *     local.read * 2
     *   })
     * }}}
     *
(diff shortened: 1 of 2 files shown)


monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
```diff
@@ -116,7 +124,7 @@ final class TaskLocal[A] private (default: => A) {
     *   // Should yield 200 on execution, regardless of what value
     *   // we have in `local` at the time of evaluation
     *   val task = local.bind(100)(Task {
-    *     local.get * 2
+    *     local.read * 2
```
oleg-py: Actually, `local` here is `Task[TaskLocal[Int]]` and `local.read` returns a `Task[Int]` too, so it cannot be multiplied directly. So the code sample should be something along the lines of

```scala
for {
  local <- TaskLocal(0)
  value <- local.bind(100)(local.read.map(_ * 2))
} yield value
```

monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
```diff
@@ -95,6 +95,15 @@ import monix.execution.misc.Local
...
   private[this] val ref = new Local(default)
 
+  /** Returns [[monix.execution.misc.Local]] instance used in this [[TaskLocal]].
+    *
+    * Note that `TaskLocal.bind` will restore the original local value
+    * on the thread where the `Task's` run-loop ends up so it might lead
+    * to leaving local modified in other thread.
+    */
+  def local: Task[Local[A]] =
+    Task.eval(ref)
```
leandrob13: If this is going to be exposed, shouldn't we define read/write/clean with `local`?

```scala
def read: Task[A] =
  local.map(_.get)
```

monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
```diff
@@ -142,13 +151,13 @@ final class TaskLocal[A] private (default: => A) {
...
-    *   val local = TaskLocal(0)
-    *
     *   // Should yield 200 on execution, regardless of what value
     *   // we have in `local` at the time of evaluation
-    *   val task = local.bindL(Task.eval(100))(Task {
-    *     local.get * 2
-    *   })
+    *   val task: Task[Int] =
+    *     for {
+    *       local <- TaskLocal(0)
```
leandrob13: I am worried that the constructor yields a `Task[TaskLocal[T]]`. The thing is that the `TaskLocal` constructor was used for initializing the local, the read and write operations were the ones that lifted the access of the underlying local into a `Task`.  It means that even if I want to expose the local from the creation, I would get a `Task[Task[Local]]` which will always require a flatmap operation. Why did it change?

leandrob13:
Would be also useful if we could obtain a Coeval from the local?

```scala
def maybeLocal(implicit sc: Scheduler):  Coeval[Either[CancelableFuture[Local], Local]] = 
  local.coeval
```

monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
Avasil: That's related to #539

leandrob13: @Avasil I missed that one! Understood!

monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
alexandru: Not necessarily. It's more elegant, but internals are allowed to be ugly for performance reasons 😀

monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
alexandru: @leandrob13 yeah, it's related to purity. With `Task` you can do `.memoize` btw and `flatMap` isn't that expensive.

Issue #601, opened 2018-02-25
TaskLocal.bind might not restore value on cancelled tasks


oleg-py:
If task passed to `bind` is cancelled, it is possible for `TaskLocal` to never be restored to its original state:
```scala
import monix.eval._
import scala.concurrent.duration._

object Sample extends TaskApp {
  override def runc: Task[Unit] = for {
    local <- TaskLocal[String]("Good")
    forked <- Task.sleep(1.second).fork
    _ <- forked.cancel
    _ <- local.bind("Bad!")(forked).fork
    _ <- Task.sleep(1.second)
    s <- local.read
    _ <- Task.eval(println(s)) // prints Bad!
  } yield ()
}
```

This can probably be solved by using `bracket` to implement `TaskLocal.bind`

alexandru:
Indeed, we might need to implement it in terms of `bracket`.

leandrob13:
@oleg-py  I can work on this if you haven't started it already

leandrob13:
@oleg-py I couldn't reproduce the error in your example. I changed it to this:

```scala
implicit val ec: Scheduler = monix.execution.Scheduler.Implicits.global
implicit val opts = Task.defaultOptions.enableLocalContextPropagation

val test: Task[Unit] = for {
      local <- TaskLocal[String]("Good")
      forked <- Task.sleep(1.second).fork
      //_ <- forked.cancel
      _ <- local.bind("Bad!")(forked.cancel).fork // needed a Task and forked is a Fiber
      _ <- Task.sleep(1.second)
      s <- local.read
      _ <- Task.now(println(s)) // prints Good
    } yield ()

test.runAsyncOpt
```

And I am getting `Good` printed.

I also changed it to this:

```scala
val test: Task[Unit] = for {
      local <- TaskLocal[String]("Good")
      forked <- Task.sleep(1.second).fork
      //_ <- forked.cancel
      _ <- local.bind("Bad!")(forked.cancel) // needed a Task and forked is a Fiber
      _ <- Task.sleep(1.second)
      s <- local.read
      _ <- Task.now(println(s)) // prints Good
    } yield ()
```

So it would stay on the same `Thread` after the bind and still got `Good` printed.

Am I missing something?

How can we assert that in fact a `Task` was cancelled this way?

oleg-py:
@leandrob13 I might have gotten this completely wrong if it works properly with `runAsyncOpt` on all scheduler types. Could be that it is not a problem.

[...]

PR #603, opened 2018-02-26
Changed implementation TaskLocal#bind in terms of Task#bracket

leandrob13:
Addresses #601 

Changes the `TaskLocal#bind ` implementation in terms of `Task#bracket` which handles any cleanup of resources upon success, failure or cancellation.

```scala
def bind[R](value: A)(task: Task[R]): Task[R] =
    Task.suspend {
      val saved = ref.value
      ref.update(value)
      task.bracket(r => Task.now(r))(_ => restore(saved))
    }
```


monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
```diff
@@ -134,7 +134,7 @@ final class TaskLocal[A] private (default: => A) {
     Task.suspend {
       val saved = ref.value
       ref.update(value)
-      task.doOnFinish(_ => restore(saved))
+      task.bracket(r => Task.now(r))(_ => restore(saved))
```
oleg-py: If you want to handle cancellation, `task` has to be used inside bracket. E.g.

```scala
Task.unit.bracket(_ => task)(_ => restore(saved))
```

Also `bindL` needs to have similar change, so maybe it's easier to implement `bind` as `bindL(Task.now(value))`

leandrob13:
@oleg-py  So I redefined bindL like this:

```scala
def bindL[R](value: Task[A])(task: Task[R]): Task[R] = {
    val saved = ref.value
    value.bracket { v =>
      ref.update(v)
      task
    }(_ => restore(saved))
  }
```

Also, took care of `bindClear` too.

monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
```diff
@@ -170,12 +166,13 @@ final class TaskLocal[A] private (default: => A) {
...
     *        reset to the previous value
     */
-  def bindL[R](value: Task[A])(task: Task[R]): Task[R] =
-    value.flatMap { value =>
-      val saved = ref.value
-      ref.update(value)
-      task.doOnFinish(_ => restore(saved))
-    }
+  def bindL[R](value: Task[A])(task: Task[R]): Task[R] = {
+    val saved = ref.value
```
alexandru: This `ref.value` should be suspended as well ...

```scala
Task.eval(ref.value).flatMap { saved =>
  value.bracket {
    ref.update(v)
    task
  } { _ =>
    restore(saved)
  }
}
```

monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
```diff
@@ -140,11 +140,7 @@ final class TaskLocal[A] private (default: => A) {
     *        reset to the previous value
     */
   def bind[R](value: A)(task: Task[R]): Task[R] =
-    Task.suspend {
-      val saved = ref.value
-      ref.update(value)
-      task.doOnFinish(_ => restore(saved))
-    }
+    Task.suspend(bindL(Task.now(value))(task))
```
alexandru: If you suspend that value as I mentioned below, you don't need to suspend here.

leandrob13: @alexandru  Understood, will take care of it later today.

leandrob13:
@alexandru should I trigger another build?

alexandru:
Please merge with master.

alexandru:
Not 100% sure, but I think this is fine. Will merge for now.

PR #604, opened 2018-02-28
Update Scheduler.scala

fancywriter:


alexandru:
Thanks @fancywriter, but I'm not sure I want to depend on `ExecutionContextExecutor`. Seems to me like it's an implementation detail of Scala's library.

Any reason you'd want `Scheduler` to inherit from `ExecutionContextExecutor`?

fancywriter:
`ExecutionContextExecutor` is nothing else but `ExecutionContext` with mixed in `Executor`. I thought it reduces some of the code duplication.

alexandru:
Right, but it's an implementation detail of Scala's library that you can't count on in the long term.
The reason is a little subtle ...

`ExecutionContext` is a clear subset of Java's `Executor`. That `ExecutionContext` itself does not inherit directly from `Executor`, that should be a question mark.

So I don't want us to depend on it simply because I understand `ExecutionContext`, I understand `Executor`, but I don't understand why `ExecutionContextExecutor` exists.

alexandru:
I don't want to discourage you from contributing btw. Please keep them coming 😀

fancywriter:
OK, I agree, it's more **scala** problem than **monix**, let's keep it.

satorg:
@alexandru, @fancywriter ,
I'd also vote for this change, because inheriting `Scheduler` from `ExecutionContextExecutor` helps with interoperability with some other libraries/frameworks.

For example, Akka HTTP uses `ExecutionContextExecutor` in its public API, see the `withExecutionContext` directive:
```scala
def withExecutionContext(ec: ExecutionContextExecutor): Directive0 = ...
```
So it seems that Akka HTTP's authors don't think that `ExecutionContextExecutor` is an internal scala library stuff.

And if someone wants to have Akka HTTP server backed by Monix streams (which is not something unusual), they might need to inject Monix `Scheduler` into Akka HTTP pipeline with the aforementioned directive (for example – to leverage a traced `Scheduler`). It isn't possibly out-of-the-box now, though, and requires to write some boilerplate wrapper around `Scheduler`. It is easy indeed, but why reinvent the wheel?

alexandru:
My problem with `ExecutionContextExecutor` is that it sounds like a band aid and it might get deprecated in the future.

@viktorklang what do you think of `ExecutionContextExecutor`. Will it be deprecated, or can we depend on it?

viktorklang:
@alexandru It's not going to go away any soon. The limitation of using it is that you'll be more heavily depending on Java features, which may or may not be an issue for something like Scala.js or Scala Native.

PR #607, opened 2018-03-08
Implement ConcurrentEffect[Task]

alexandru:
In `cats-effect` I worked on introducing `Task` instances for:

1. `cats.effect.Concurrent`
2. `cats.effect.ConcurrentEffect`

The new [Concurrent](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Concurrent.scala) type class in `cats-effect` introduces concurrent, cancelable data types that are capable of being forked (via `start`) and that can participate in race conditions (via `race` and `racePair`). We can now describe polymorphic code making use of `Task`'s cancelability features.

In this process I also changed the `Task` conversions available to use `Concurrent` in case it is available.

```scala
import cats.effect._
import cats.syntax.all._

val io1 = IO.sleep(5.seconds) *> IO(println("Hello!"))

val task1 = Task.fromEffect(io)
val io2 = task1.to[IO]
```

In this example the `io1` value is cancelable, the `task1` value resulting from it is also cancelable and the conversion back in `io2` also preserves its cancelability.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,7 +10,7 @@ addCommandAlias("ci-js",      ";clean ;coreJS/test:compile  ;coreJS/test")
 addCommandAlias("release",    ";project monix ;+clean ;+package ;+publishSigned ;sonatypeReleaseAll")
 
 val catsVersion = "1.0.1"
-val catsEffectVersion = "0.10-c64e7c9"
+val catsEffectVersion = "0.10-4e98208"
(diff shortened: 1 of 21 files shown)


alexandru:
Pushed PR in cats-effect: https://github.com/typelevel/cats-effect/pull/142

PR #610, opened 2018-03-18
Observable.scan0 & friends

oleg-py:
Alternative to #587

I went ahead and made versions that emit seed element for all scan operators that exist for `Observable`.

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -1352,9 +1352,20 @@ abstract class Observable[+A] extends Serializable { self =>
     * producing and concatenating observables along the way.
     *
     * It's the combination between [[scan]] and [[flatMap]].
+    *
+    * @see [[flatScan0]] for the version that emits seed element at the beginning
     */
   final def flatScan[R](seed: => R)(op: (R, A) => Observable[R]): Observable[R] =
     new FlatScanObservable[A, R](self, seed _, op, delayErrors = false)
+
+  /** Applies a binary operator to a start value and to elements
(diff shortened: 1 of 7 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -1352,10 +1352,21 @@ abstract class Observable[+A] extends Serializable { self =>
...
     new FlatScanObservable[A, R](self, seed _, op, delayErrors = false)
 
+  /** Applies a binary operator to a start value and to elements
+    * produced by the source observable, going from left to right,
+    * producing and concatenating observables along the way.
+    *
+    * It's the combination between [[scan0]] and [[flatMap]].
+    */
+  final def flatScan0[R](seed: => R)(op: (R, A) => Observable[R]): Observable[R] =
+    Observable.eval(seed).flatMap(s => s +: flatScan(s)(op))
```
alexandru: This is fine, but I would use `++` because it is more efficient and stack/memory safe for that matter.

oleg-py: Since `seed` is by-name parameter, I am assuming it can be effectful, in which case it should be evaluated once, as in regular `scan`. I don't think it can be done without `flatMap`.

alexandru: Sorry, I just skimmed the code without thinking. You actually need `flatMap` here. So it's fine.

alexandru:
@oleg-py I think `Iterant` also needs the same treatment. Would you like to do a PR for it as well?

oleg-py:
@alexandru will do

PR #613, opened 2018-03-19
Fix execute on task local

### LINKED ISSUE #612: TaskLocal doesn't preserve locals for Task.executeOn
See code samples below.

`asyncBoundary` works fine:
```scala
implicit val ec: Scheduler = monix.execution.Scheduler.Implicits.global
val ec2: Scheduler = monix.execution.Scheduler.io()

for {
  local <- TaskLocal(0).asyncBoundary(ec2)
  _ <- local.write(100).asyncBoundary(ec)
  v1 <- local.read.asyncBoundary(ec2) // 100
  v2 <- local.read // 100
  _ <- Task { println(s"$v1 : $v2")}
} yield ()
```

`executeOn`:
```scala
for {
  local <- TaskLocal(0)
  _ <- local.write(100).executeOn(ec2)
  v1 <- local.read.executeOn(ec)  // 0
  v2 <- local.read.executeOn(ec2) // 0
  _ <- Task { println(s"$v1 : $v2")}
} yield ()
```

[Link to the discussion on Gitter](https://gitter.im/monix/monix?at=5aafaafb6f8b4b99464384fb)

leandrob13:
fixes #612 

When executing the following test:

```scala
implicit val ec: Scheduler = monix.execution.Scheduler.Implicits.global
implicit val opts = monix.execution.Scheduler.io()

testAsync("Local.apply with different schedulers with onExecute") {
val ec2: Scheduler = TracingScheduler(Scheduler.computation(4, "ec2"))
    val test =
      for {
        local <- TaskLocal(0)
        _ <- local.write(1000).executeOn(ec2)
        v1 <- local.read.executeOn(ec)
        v2 <- local.read.executeOn(ec2)
        _ <- Task.now(assertEquals(v1, 1000))
        _ <- Task.now(assertEquals(v2, 1000))
      } yield ()

    test.runAsyncOpt
  }
```
The `local.write(1000).executeOn(ec2)` does not set the local var before the `executeAsyncTask` method gets to [prepare the locals](https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/internal/TaskRunLoop.scala#L420) in the `RestartCallback`. Which means, by the time it does and it calls the `onSuccess` function it will have the `savedLocals` internal var as an empty map.

The solution proposed now is to let the `TracingScheduler` handle the propagation in async boundaries. After removing the local bindings in `TaskRunLoop` we can successfully execute the test as follows:

```scala
implicit val opts: Task.Options = Task.defaultOptions.enableLocalContextPropagation
implicit val ec: Scheduler = monix.execution.Scheduler.Implicits.global

testAsync("Local.apply with different schedulers with onExecute") {
val ec2: Scheduler = Scheduler.computation(4, "ec2")
    val test =
      for {
        local <- TaskLocal(0)
        _ <- local.write(1000).executeOn(ec2)
        v1 <- local.read.executeOn(ec)
        v2 <- local.read.executeOn(ec2)
        _ <- Task.now(assertEquals(v1, 1000))
        _ <- Task.now(assertEquals(v2, 1000))
      } yield ()

    test.runAsyncOpt
  }
```


monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
```diff
@@ -110,7 +110,7 @@ final class TaskLocal[A] private (default: => A) {
 
   /** Updates the local value. */
   def write(value: A): Task[Unit] =
-    Task.eval(ref.update(value))
+    Task.now(ref.update(value))
```
oleg-py: I think it makes `write` non-referentially transparent, no?

alexandru: Hey — so this might fix this particular issue, but we cannot do it. 

If it has eager behavior, then it is fundamentally incompatible with `Task` unfortunately.

leandrob13:
@alexandru @oleg-py understood guys. Reverting and digging deeper.

leandrob13:
@alexandru @oleg-py Updated description with the proposal of the solution. Javadocs and code polish will be completed once I get the thumbs up.

I think this might solve the issue for every `Task` creation that relies on `Task.Async` with a `Callback.async` instance.

Avasil:
@leandrob13 
I think the solution is pretty good, as you mentioned it should fix `TaskExecuteOn`, `Task.uncancelable`, `TaskExecuteWithOptions` and I'm not sure if some `MVar` functions weren't affected. Would be cool to check before and after the fix. :)

I wonder if the same problem isn't applicable to methods like `Callback.fromPromise`? We should check and if my understanding is correct then all it needs is the same fix so that's cool.

leandrob13:
@Avasil understood, I could check as soon as I can on `Callback.fromPromise`.

alexandru:
Hey, I just added a comment at: https://github.com/monix/monix/issues/612#issuecomment-379433943

leandrob13:
@alexandru I did changes based on your comments in #612. I do think it is a better solution to use `TracingScheduler` when propagating locals.

monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
```diff
@@ -193,9 +194,9 @@ final class TaskLocal[A] private (default: => A) {
     *        the local gets reset to the previous value
     */
   def bindClear[R](task: Task[R]): Task[R] =
-    Task.suspend {
-      val saved = ref.value
-      ref.clear()
+    local.flatMap { r =>
```
alexandru: Does this call to `Local`? It breaks referential transparency.

monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
```diff
@@ -19,10 +19,11 @@ package monix.eval
 
 import minitest.SimpleTestSuite
 import monix.execution.Scheduler
+import monix.execution.schedulers.TracingScheduler
```
alexandru: Having users import TracingScheduler by themselves might not be OK.

It’s usage should probably be automatic when that option is on.

monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
leandrob13: @alexandru not the `Local` object but the function in `TaskLocal`:

```scala
def local: Task[Local[A]] =
    Task.eval(ref)
```

Since we added it I thought it was better to use it when necessary.

monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
leandrob13: @alexandru Nice idea. So if the option is on, basically wrap the current scheduler into the `TracingScheduler`. Will work on it.

oleg-py: BTW I would like to have `.tracing` method on a scheduler that returns a `TracingScheduler`. For those that are already tracing, it could simply return `this`.

leandrob13: @oleg-py that could be an extension method, will try it out.

monix-eval/shared/src/main/scala/monix/eval/TaskLocal.scala
alexandru: Sorry, I’m on my phone and the mobile GitHub interface is pretty bad.

leandrob13: @alexandru no problem, it is indeed terrible.

leandrob13:
Closing this PR because the primary work is being done in #647 

I will open a new one with a minor fix on TaskLocal#bindL implementation.

PR #621, opened 2018-03-25
Fix #620: change behavior of Batch#fromArray, fromAnyArray

### LINKED ISSUE #620: Batch.fromAnyArray(Array[Int]) throws
```scala
val array: Array[Int] = (0 until 1000).toArray

Batch.fromAnyArray(array)
// => java.lang.ClassCastException: [I cannot be cast to [Ljava.lang.Object;
// =>   at monix.tail.batches.Batch$.fromAnyArray(Batch.scala:176)
// =>   at monix.tail.batches.Batch$.fromAnyArray(Batch.scala:185)
// =>   at monix.tail.Playground$.<init>(Playground.scala:63)
```

The signature currently is:

```scala
def fromAnyArray[A](array: Array[_]): ArrayBatch[A]
```

Arrays on the JVM are reified. Thus a conversion is required in case `array.isInstanceOf[Array[AnyRef]] == false`.

alexandru:
In the `ArrayBatch` and `ArrayCursor` we rely on having a `ClassTag[A]` reference for being able to build new arrays. This is because arrays on the JVM are reified.

The detail that I missed is that we can build a `ClassTag` reference out of `array.getClass.getComponentType`. 

Interestingly this works for JavaScript too. I have no idea what Scala.js is doing to maintain compatibility for arrays.


monix-tail/shared/src/main/scala/monix/tail/batches/Batch.scala
```diff
@@ -163,19 +163,19 @@ object Batch {
...
+  }
 
   /** $fromAnyArrayDesc
     *
     * @param array $paramArray
     * @param offset $paramArrayOffset
     * @param length $paramArrayLength
     */
-  def fromAnyArray[A](array: Array[_], offset: Int, length: Int): ArrayBatch[A] = {
-    val ref = new ArrayBatch[Any](array.asInstanceOf[Array[Any]], offset, length, arrayAnyBuilder)
```
Avasil: We can now remove `arrayAnyBuilder` in `monix.tail.batches` package object, right?

alexandru: Yes. Will do.

monix-tail/shared/src/main/scala/monix/tail/batches/Batch.scala
```diff
@@ -163,19 +163,19 @@ object Batch {
...
-    ref.asInstanceOf[ArrayBatch[A]]
-  }
+  def fromAnyArray[A](array: Array[_], offset: Int, length: Int): ArrayBatch[A] =
+    fromArray(array, offset, length).asInstanceOf[ArrayBatch[A]]
```
oleg-py: Is this method even needed anymore? Judging by the doc it has been introduced exactly for the cases where you don't have a `ClassTag`, and is not really type-safe, i.e. this compiles:

    Batch.fromAnyArray[Int](Array("A string!"), 0, 1)

oleg-py:
Nice trick! I too totally missed the fact you can build a `ClassTag` out of java `Class`

monix-tail/shared/src/main/scala/monix/tail/batches/Batch.scala
alexandru: Due to Array's reification it might be needed. This is because Java allows boxing of primitives into `AnyRef`, so you can do this:

```scala
val array = new Array[Any](10)
array(0) = 1001
```

And we have to do this unfortunately when describing a `map[B](f: A => B)` operation in `ArrayCursor` for example. What happens is that we build a new `Array[Any]`, iterate the source and initialize it, but at the end you have an `Array[Any]` on your hands and it's not something you can workaround unless you require a `ClassTag[B]`.

And then manual casting doesn't work either, due to reification:

```
scala> array.asInstanceOf[Array[Int]]
java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [I
  ... 28 elided
```

So unfortunately I think it's still needed in order to work with `Array[Any]`, due to reification.

I hope generics reification never makes it in Java 🙂

oleg-py: Sure, you cannot cast an array. But you can cast the returned `ArrayCursor` itself.

alexandru: Ah, right 🙂 Will try it out and if so, we'll get rid of it.

monix-tail/shared/src/main/scala/monix/tail/batches/Batch.scala
alexandru: I removed the package object completely. Thanks for pointing it out.

alexandru:
I removed `Batch.fromAnyArray` and `BatchCursor.fromAnyArray`, relying on casting when dealing with `fromArray`.

Also removed unused `ClassTag` parameters and the `monix.tail.batches` package object.

Spring cleanup 🙂

PR #622, opened 2018-03-25
Add mapBatch for Iterant

### LINKED ISSUE #617: Iterant#mapBatch
We need an operation like this:

```scala
  // ...
  def mapBatch[B](f: A => Batch[B]): Iterant[F, B] = ???
}
```

The yielded `Iterant` then be made of `NextBatch` nodes. This is useful for optimizing a data stream, for example for converting an `Iterant[F, Array[A]]` into an optimized `Iterant[F, A]`.

Avasil:
Closes #617

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantMapBatch.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantMapBatch.scala
@@ -32,15 +32,18 @@ private[tail] object IterantMapBatch {
   def apply[F[_], A, B](source: Iterant[F, A], f: A => Batch[B])
     (implicit F: Sync[F]): Iterant[F, B] = {
 
-    def processBatch(ref: NextCursor[F, A]): NextBatch[F, B] = {
+    def processBatch(ref: NextCursor[F, A]): Iterant[F, B] = {
       val NextCursor(cursor, rest, stop) = ref
 
-      val batch = if(cursor.hasNext()) f(cursor.next()) else Batch[B]()
-      val next: F[Iterant[F, A]] =
-        if (cursor.hasNext()) F.pure(ref)
-        else rest
+      if(cursor.hasNext()) {
+        val next: F[Iterant[F, A]] =
+          if (cursor.hasNext()) F.pure(ref)
+          else rest
 
-      NextBatch(batch, next.map(loop), stop)
+        NextBatch(f(cursor.next()), next.map(loop), stop)
+      } else {
+        Suspend(rest.map(loop), stop)
+      }
     }
 
     def loop(source: Iterant[F, A]): Iterant[F, B] =
--- a/monix-tail/shared/src/main/scala/monix/tail/internal/IterantMapBatch.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/internal/IterantMapBatch.scala
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) 2014-2018 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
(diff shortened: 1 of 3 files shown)


Avasil:
btw I've noticed bugs with following:
```scala
val stream = Iterant[Task].nextBatchS(Batch.fromSeq(List(elem, elem), 2), Task.delay(Iterant[Task].lastS(elem)), Task.unit)
val f: Int => List[Int] = List.fill(3)(_)
```
so I'm trying to fix it right now

monix-tail/shared/src/main/scala/monix/tail/internal/IterantMapBatch.scala
```diff
@@ -0,0 +1,95 @@
...
+    def processSeq(ref: NextCursor[F, A]): NextBatch[F, B] = {
+      val NextCursor(cursor, rest, stop) = ref
+      val buffer = ArrayBuffer.empty[B]
+      var toProcess = cursor.recommendedBatchSize
+      var nextRef: F[Iterant[F, B]] = null.asInstanceOf[F[Iterant[F, B]]]
+
+      // protects against infinite cursors
+      while (toProcess > 0 && cursor.hasNext()) {
+        val cursorB = f(cursor.next()).cursor()
+        while (toProcess > 0 && cursorB.hasNext()) {
```
alexandru: When that mapping function gives you a `Batch`, there's no point in splitting it any further, because the `Batch` is already built for you and you can just serve it as is in a `NextBatch` node. All operations need to handle fairness concerns via `recommendedBatchSize`, so even if you serve bigger batches via this function, it doesn't matter.

The point of `mapBatch` is to be used in certain cases for optimizing a stream, e.g. for turning an `Iterant[F, Array[A]]` into an `Iterant[F, A]`. If we end up splitting those batches, then it becomes too heavy. At that point we might as well work with `flatMap` directly:

```scala
def mapBatch[F[_] : Sync, A](fa: Iterant[F, A])(f: A => Batch[B]): Iterant[F, B] =
  fa.flatMap { a =>
    Iterant.nextBatchS(f(a), Iterant.empty, F.unit)
  }
```

So here what I see happening is no batched processing at all, e.g..

```scala
val batch = f(cursor.next()).cursor()
val next = 
  if (cursor.hasNext()) NextCursor(cursor, rest.map(loop), stop)
  else rest.map(loop)

NextBatch(batch, next, stop)
```

monix-tail/shared/src/main/scala/monix/tail/internal/IterantMapBatch.scala
```diff
@@ -0,0 +1,74 @@
...
+  /**
+    * Implementation for `Iterant#mapBatch`
+    */
+  def apply[F[_], A, B](source: Iterant[F, A], f: A => Batch[B])
+    (implicit F: Sync[F]): Iterant[F, B] = {
+
+    def processBatch(ref: NextCursor[F, A]): NextBatch[F, B] = {
+      val NextCursor(cursor, rest, stop) = ref
+
+      val batch = if(cursor.hasNext()) f(cursor.next()) else Batch[B]()
```
Avasil: @alexandru 
I've changed it to what you suggested but I'm not sure about this case. Is it ok to handle it like this?
Alternatively I could return `Suspend` when I get empty cursor

alexandru: Sorry for not replying, I'm busy these days at work.

A `Suspend` is better than an empty `NextBatch`. Although it's not a major issue, since they should be equivalent.

The rest looks pretty good. Do this small change and it's ready to go.

PR #626, opened 2018-03-28
Add forever for Task

RaasAhsan:
Didn't see a method similar to this, correct me if I'm wrong. Basically, given a source Task, return another Task that continuously executes the source Task forever, until an error happens or cancelled. Will add better documentation and necessary tests if needed

My use case for this was in a producer/consumer system where a consumer takes something off a (blocking) queue, does some processing, and loops back.

What I had was something recursive like this:

```scala
def loop: Task[Unit] = {
  for {
    item <- Task.eval(queue.take())
    _       <- process(item)
    _       <- loop       
  } yield ()
}
```

which could be expressed as:

```scala
val task = for {
  item <- Task.eval(queue.take())
  _       <- process(item)
} yield ()

val loop = task.forever
```

Is this a good idea or is there something I'm missing about why it shouldn't be done or should be done another way?

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -1235,10 +1235,10 @@ sealed abstract class Task[+A] extends Serializable {
     foreachL(f).runAsync(s)
 
   /** Returns a new `Task` that executes the source repetitively
-    * and never produces a value.
+    * as long as it succeeds and never produces a value.
     */
-  final def forever: Task[Nothing] =
-    flatMap(_ => this.forever)
+  final def loopForever: Task[Nothing] =
+    flatMap(_ => this.loopForever)
 
   /** Start asynchronous execution of the source suspended in the `Task` context.
     *
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -1234,8 +1234,17 @@ sealed abstract class Task[+A] extends Serializable {
   final def foreach(f: A => Unit)(implicit s: Scheduler): CancelableFuture[Unit] =
     foreachL(f).runAsync(s)
 
-  /** Returns a new `Task` that executes the source repetitively
-    * as long as it succeeds and never produces a value.
+  /** Returns a new `Task` that executes the source repeatedly,
+    * as long as it succeeds. It never produces a terminal value.
+    *
+    * Example:
+    *
+    * {{{
+    *   Task.eval(println("Tick!"))
+    *     .delayExecution(1.second)
+    *     .loopForever
+    * }}}
+    *
     */
   final def loopForever: Task[Nothing] =
     flatMap(_ => this.loopForever)
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -1234,12 +1234,14 @@ sealed abstract class Task[+A] extends Serializable {
   final def foreach(f: A => Unit)(implicit s: Scheduler): CancelableFuture[Unit] =
     foreachL(f).runAsync(s)
 
-  /** Returns a new `Task` that executes the source repeatedly,
-    * as long as it succeeds. It never produces a terminal value.
+  /** Returns a new `Task` that repeatedly executes the source as long
+    * as it continues to succeed. It never produces a terminal value.
     *
     * Example:
     *
     * {{{
+    *   import scala.concurrent.duration._
+    *
     *   Task.eval(println("Tick!"))
     *     .delayExecution(1.second)
     *     .loopForever
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -1234,6 +1234,12 @@ sealed abstract class Task[+A] extends Serializable {
   final def foreach(f: A => Unit)(implicit s: Scheduler): CancelableFuture[Unit] =
     foreachL(f).runAsync(s)
 
+  /** Returns a new `Task` that executes the source repetitively.
+    *
+    */
+  final def forever: Task[A] =
+    flatMap(_ => this.forever)
+
   /** Start asynchronous execution of the source suspended in the `Task` context.
     *
     * This can be used for non-deterministic / concurrent execution.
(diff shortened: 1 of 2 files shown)


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1234,6 +1234,12 @@ sealed abstract class Task[+A] extends Serializable {
   final def foreach(f: A => Unit)(implicit s: Scheduler): CancelableFuture[Unit] =
     foreachL(f).runAsync(s)
 
+  /** Returns a new `Task` that executes the source repetitively.
+    *
+    */
+  final def forever: Task[A] =
+    flatMap(_ => this.forever)
```
oleg-py: I think `loopForever` will read better when I see it in code. It should also probably return `Task[Nothing]` to indicate non-termination.

RaasAhsan: Regarding `Task[Nothing]`, would have to use the async creator to get that type signature, since we can't produce a value of type `Nothing`, yeah? Could also use `Task[Unit]`, but I suppose that indicates termination, just with a value of type `Unit`.

oleg-py: @RaasAhsan you can't produce a value, but you can have that in type if you never return anything.

E.g. this typechecks:
```
def loop: Nothing = loop
```

But if I *call* `loop`, I will never get a value back.

RaasAhsan: Got it, that makes sense, just need to change the type signature then.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1234,6 +1234,12 @@ sealed abstract class Task[+A] extends Serializable {
   final def foreach(f: A => Unit)(implicit s: Scheduler): CancelableFuture[Unit] =
     foreachL(f).runAsync(s)
 
+  /** Returns a new `Task` that executes the source repetitively
+    * and never produces a value.
+    */
+  final def forever: Task[Nothing] =
+    flatMap(_ => this.forever)
```
martin-g: Is this stack safe ?

oleg-py: Yes. Flat-mapping `Task`s is always safe, even in this case. This is also specified in `cats.effect.Sync`

alexandru:
Sorry for the late reply @RaasAhsan, I'm busy at work.

I'm reluctant on adding your `forever` because we already have `restartUntil`, so ...

```scala
task.forever <-> task.restartUntil(_ => true)
```

Note how `restartUntil` is more flexible in that it allows to specify an exit condition for your loop.
Speaking of a queue, you can imagine processing stuff until the queue is empty:

```scala
Task.eval(queue.take())
  .map(process)
  .restartUntil(_ => queue.isEmpty)
```

Having utils is good, I don't mind it, but too many overloads of approximately the same thing hurts usability.

@oleg-py @martin-g given that you took a look at it, what do you think? Is `forever` useful in addition to `restartUntil`?

RaasAhsan:
Ahh, I missed that method through my look, seems to fit the bill! No strong feelings about this one then, if you guys decide to keep it I can make necessary changes.

Thanks!

oleg-py:
@alexandru I'd vouch for having it (but named `loopForever`) and implemented exactly as `restartUntil`. I never found this as a way to make a `Task` loop forever. This would have helped me to discover `restartUntil` itself (which *is* neat, although not pure).

martin-g:
I personally do not find it that useful.

1. It can be implemented via `restartUntil`
2. I find it dangerous and confusing to start something that has no way to stop it.

oleg-py:
@martin-g it is still as cancelable as original task is. You can do `task.forever.cancelable` and will be able to stop it on any flatMap.

alexandru:
@oleg-py `restartUntil` can be used in a pure way, as that function's parameter receives the task's result as input. So you can do stuff like ...

```scala
Task.eval(Random.randInt()).restartUntil(_ % 2 == 0)
```

oleg-py:
@alexandru sure. But I cannot do something like

```
def secondsPassed: Task[Int] = ???
def myJob: Task[Unit] = ???
myJob.restartUntil(_ => secondsPassed.map(_ > 60))
```

alexandru:
Well, you can always suspend that and trigger whatever side effects you want 🙂

This is actually a pet peeve btw. I'd like to express `Scheduler#scheduleAtFixedRate` and `Observable#intervalAtFixedRate`. Can't do that with `restartUntil`.

Have you seen `onErrorRestartLoop` btw? Maybe we need something like that for the happy path as well. Although to be honest it might not be easily understandable by beginners.

oleg-py:
@alexandru 
> Well, you can always suspend that and trigger whatever side effects you want

I have to `.run...` the `secondsPassed` task then. I *don't* want to 😆 

> Have you seen onErrorRestartLoop btw? Maybe we need something like that for the happy path as well.

Isn't `forever` exactly `onErrorRestartLoop` for the happy path? next `flatMap` won't execute if the Task fails. Maybe for consistency sake it could be named `onSuccessRestartLoop`.

> This is actually a pet peeve btw

Sure. I have a few more operations I would like `Task` to have but they are all one-liners I don't mind adding to project myself, so no problem.

alexandru:
@RaasAhsan please rename to `loopForever`

RaasAhsan:
@alexandru done, let me know if any documentation needs to be added or anything else

RaasAhsan:
Tests seemed to fail after that commit (just renamed), not sure if I can restart it on my own without pushing another commit

alexandru:
I restarted that build job, we'll see.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1234,6 +1234,12 @@ sealed abstract class Task[+A] extends Serializable {
   final def foreach(f: A => Unit)(implicit s: Scheduler): CancelableFuture[Unit] =
     foreachL(f).runAsync(s)
 
+  /** Returns a new `Task` that executes the source repetitively
```
alexandru: Change `repetitively` to `repeatedly`.

Also add this example:

```scala
Task.eval(println("Tick!"))
  .delayExecution(1.second)
  .loopForever
```

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1234,6 +1234,12 @@ sealed abstract class Task[+A] extends Serializable {
...
+    * as long as it succeeds and never produces a value.
+    */
+  final def loopForever: Task[Nothing] =
+    flatMap(_ => this.loopForever)
```
alexandru: So we need a test for it. Do something that triggers an effect, e.g...

```scala
Task.suspend {
  var idx = 0
  var sum = 0

  val err = new CancellationException
  Task.eval { if (idx < 10) { sum += idx; idx += 1 } else throw err }
    .loopForever
    .onErrorRecover { case `err` => sum }
}
```

RaasAhsan:
@alexandru fixed documentation and added a test

alexandru:
Thanks.

Issue #624, opened 2018-03-27
TaskLocal not always restoring context after execution


oleg-py:
I finally managed to reproduce the issue I was talking about on gitter.

The app below crashes with `MatchError` because variable becomes `true` on 2nd iteration (on my machine).

Happens on both RC1 and #613 (commit # 63044eb)

```scala
import monix.execution.Scheduler.Implicits.global
import monix.eval._
import scala.concurrent.duration._

object Fakes extends App {
  implicit val opts = Task.Options(false, true)
  val local = TaskLocal[Boolean](false).runSyncUnsafeOpt(Duration.Inf)

  def attempt = local.read flatMap {
    case false => local.bind(true) {
      Task.unit
    }
  }

  for (_ <- 1 to 1000) {
      attempt.runSyncUnsafeOpt(Duration.Inf)
  }
}
```
Ironically, this works when LCP is *disabled*

/cc @leandrob13

leandrob13:
@oleg-py got it, will test and debug asap

alexandru:
So this sample should probably work because it's on the same thread, don't know what is wrong here; but it got me thinking ... your usage of `runSyncUnsafeOpt` in that loop is incompatible with the chosen model and this can't work for all tasks.

The problem gets more complicated when you introduce forks:

```scala
  def attempt = local.read flatMap {
    case false => local.bind(true) {
      Task.shift // <-- fork
    }
  }
```

In such a case the restore will happen on another thread entirely, captured in a `ThreadLocal`, which will never be seen from `main`. Basically the restore happens for the bind continuation of that `Task`, so the loop that should work is this:

```scala
def loop(n: Int = 1000): Task[Unit] =
  if (n > 0) attempt.flatMap(_ => loop(n - 1))
  else Task.unit
```

Is this a use case we care about?

If yes, then we cannot deliver on `TaskLocal`. If no, then we need to warn people in the docs, but if somebody as capable as you @oleg-py couldn't see the problem in this, then I'm afraid that `TaskLocal` is too error prone.

alexandru:
Btw, our implementation is inspired by the implementation of Twitter's `Future` and `Local`.

But it has one big difference. Back when we implemented this, @leandrob13 used an immutable `Map` in that `ThreadLocal` which I liked very much, however Twitter's implementation uses a mutable `Array`.

I'm now thinking that we might have made the wrong decision and that we also need a mutable `Array`.

oleg-py:
[...]

PR #635, opened 2018-04-01
Fix local context not clearing

oleg-py:
Possible fix for #624

The sync part of Task run-loop is implicitly side effecting on Local context. Those side effects were not wrapped in `bind`, so all the values written by evaluating tasks were persisted after asynchronous boundary or completion.

By wrapping everything in `Local.bind(Local.getContext())` we ensure that whatever values were written, would be restored back. This is o.k. since the run-loop is in charge of transferring the context further.

Opening it for discussion/suggestions. I also have yet to write tests.


monix-eval/jvm/src/main/scala/monix/eval/internal/TaskRunSyncUnsafe.scala
```diff
@@ -19,20 +19,22 @@ package monix.eval.internal
...
+
 import scala.concurrent.blocking
 import scala.concurrent.duration.{Duration, FiniteDuration}
 
 private[eval] object TaskRunSyncUnsafe {
   /** Run-loop specialization that evaluates the given task and blocks for the result
     * if the given task is asynchronous.
     */
-  def apply[A](source: Task[A], timeout: Duration, scheduler: Scheduler, opts: Task.Options): A = {
+  def apply[A](source: Task[A], timeout: Duration, scheduler: Scheduler, opts: Task.Options): A = Local.bind(Local.getContext()){
```
alexandru: I'm still digesting this POC, however one thing that really needs to happen is that this saving and restoring of the context should be a no-op in case it's not active via `Task.Options`.

oleg-py: @alexandru yep, I expected that :)

It's just I suddenly had a realization of why things might be misbehaving the way they are and went for the quickest possible way to fix 'em, and put it for you and @leandrob13 to see.

oleg-py:
I changed things to only affect the methods that start the run-loop, because in other cases it should already be handled by async boundaries processing. Will add a test case soon

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -146,6 +152,16 @@ object Local {
...
        """)
     }
+
+    def localLetCurrent(f: Tree): Tree = {
+      val saved = util.name("saved")
+      val Local = symbolOf[Local[_]].companion
+      resetTree(
+        q"""
+       val $saved = $Local.getContext()
+       try { $f } finally { $Local.setContext($saved) }
```
oleg-py: I can add another parameter to the macro, so that if...else happens here. This would let me make things more DRY but the `bindCurrent` signature will seem to be too special-cased, so I'll probably have to make it `private[monix]`

alexandru:
@oleg-py do you still want this PR?

I don't really want this feature. Can you work around it?

alexandru:
If you still want it, you'll have to merge with master and fix the PR.

oleg-py:
@alexandru I don't have strong feelings anymore, and it doesn't cause me any bugs so far, but this seems to be a surprising and unintuitive behavior, and I cannot think of a plausible workaround that we could recommend.

oleg-py:
Didn't occur to me 😅 Yeah, that's possible (ofc, `bindCurrent` is from this PR, so we can tell people to write its code inline).

alexandru:
Do a PR for `bindCurrent` 😛

oleg-py:
@alexandru why not take care of it on our side, then? :)

oleg-py:
Ok, this seems to bring in more trouble now than it's worth. Let's close this PR for now and see if any problems arise later on

Issue #619, opened 2018-03-22
What to do with Task#foreach ?


alexandru:
`Task#foreach` triggers side effects and doesn't behave like `Future#foreach`, in that it doesn't log errors.

We need to decide what to do with it. Solutions:

1. we make it behave like `Future#foreach`, to report errors via `Scheduler#reportFailure`
2. we make it return a `Task` instead, but this helps with nothing imo
3. we deprecated it in 3.0.0-RC2 and then remove it in 3.0.0

Opinions?

oleg-py:
**2.** is a no-go:
- There's already `Task#foreachL` that returns a `Task[Unit]`
- for-syntax might be silently broken:
```scala
for (a <- taskA; b <- taskB) println(a + b)

//desugars
taskA.foreach {a =>
  taskB.foreach(b => println(a + b)) // discarded value, so loop is no-op!
}
```

I think it's ok to deprecate it and leave for a while with a message to use `runAsync` or `runSyncUnsafe` instead. That means losing for-loop support, however.

---

Also, `foreach` is essentially implemented as `map { a => f(a); () }.runAsync`

So the issue you mention means that there is a more general bug: exceptions thrown inside `map` might not get reported. That has to be fixed.

Avasil:
Unfortunately I don't have time now to play around with it too much but I agree with @oleg-py that the problem might be with `map`. Shouldn't it `reportFailure` somewhere?

alexandru:
It should, but it doesn't and I got a confused user on Gitter complaining about it.

The current behavior is wrong, along with its signature.

Returning `CancelableFuture[Unit]` is also a bad idea for the same reason that @oleg-py mentioned above ... the result will not wait on an inner-for, this isn't a `flatMap` that we are describing, so returning a `Future` in here gives a false sense of what the thing does.

We need to get rid of it. 
And first we'll fix the implementation to report errors and `@deprecate` it for 3.0.0-RC.

oleg-py:
I have problems reproducing the behavior. Seems like `map` is catching things, like expected.

PR #638, opened 2018-04-04
Fix Task.foreach waiting / error reporting

oleg-py:
Partial fix for #619

By piggy-backing on stdlib Future methods, it's easy to get error reporting AND correct order of operations. It required the change in signature to return `Unit`, however.

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -1230,9 +1230,11 @@ sealed abstract class Task[+A] extends Serializable {
     *
     * The application of this function has strict behavior, as the
     * task is immediately executed.
+    *
+    * Exceptions in `f` are reported using provided (implicit) Scheduler
     */
-  final def foreach(f: A => Unit)(implicit s: Scheduler): CancelableFuture[Unit] =
-    foreachL(f).runAsync(s)
+  final def foreach(f: A => Unit)(implicit s: Scheduler): Unit =
+    runAsync.foreach(f)
 
   /** Returns a new `Task` that repeatedly executes the source as long
     * as it continues to succeed. It never produces a terminal value.
--- a/monix-eval/shared/src/test/scala/monix/eval/TaskCoevalForeachSuite.scala
+++ b/monix-eval/shared/src/test/scala/monix/eval/TaskCoevalForeachSuite.scala
@@ -18,6 +18,7 @@
 package monix.eval
 
 import minitest.TestSuite
+import monix.execution.exceptions.DummyException
 import monix.execution.schedulers.TestScheduler
 
 object TaskCoevalForeachSuite extends TestSuite[TestScheduler] {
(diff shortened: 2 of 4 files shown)


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1230,9 +1230,11 @@ sealed abstract class Task[+A] extends Serializable {
     *
     * The application of this function has strict behavior, as the
     * task is immediately executed.
+    *
+    * Exceptions in `f` are reported using provided (implicit) Scheduler
     */
-  final def foreach(f: A => Unit)(implicit s: Scheduler): CancelableFuture[Unit] =
-    foreachL(f).runAsync(s)
+  final def foreach(f: A => Unit)(implicit s: Scheduler): Unit =
+    runAsync.foreach(f)
```
alexandru: We should also add a `@deprecated` tag.

oleg-py: @alexandru I actually wanted your opinion on this. While I myself don't need `foreach` at all, this should fix all problems we discussed at #619, so there might not be a big problem if it was to stay.

alexandru: I don't know, maybe we should leave it.

I also don't like that we are changing the signature, because it creates yet another incompatibility. An alternative implementation would be something like ...

```scala
task.transform(_ => (), s.reportFailure).runAsync
```

And it keeps the `CancelableFuture[Unit]`. We don't necessarily have to `@deprecate` it now.

oleg-py: @alexandru but that doesn't fix the for-comprehension, does it?

alexandru: What do you mean? 

The problem was that it doesn't report errors via `Scheduler#reportFailure` like `Future` does, a behavior that apparently people expect.

Have you noticed other problems?

oleg-py: @alexandru 
> ... the result will not wait on an inner-for, this isn't a flatMap that we are describing, so returning a Future in here gives a false sense of what the thing does.
https://github.com/monix/monix/issues/619#issuecomment-375573445

oleg-py: ^ I thought from that perspective it makes sense to return `Unit`.

alexandru: Ah, right. I guess we can leave it with `Unit`.

alexandru:
@oleg-py tests are failing in monix-reactive due to the behavior change:

- `monix.reactive.internal.operators.EvalOnCompleteSuite`
- `monix.reactive.internal.operators.EvalOnErrorSuite`

You need to replace all usage of `Task#foreach` from `monix-reactive`.

PR #639, opened 2018-04-06
Make Coeval.value empty-parens to indicate side effects

oleg-py:
We discussed this on gitter some time ago. `Coeval#value` is now empty-parens, aligning with conventional way to indicate side-effects in Scala.

This does not trigger any warnings or errors. Only if using some kind of linter (e.g. Intellij does that) you will get something shown.

I went ahead and changed pretty much every usage of `value` to `value()` in monix codebase too.


alexandru:
Looks good.

PR #640, opened 2018-04-06
add sbt-doctest, to verify code examples in scaladoc

jozic:
this is a POC for adding sbt-doctest to monix, to make code examples in scaladoc combilable/runnable. This is just a small portion for discussion.
The motivation here is hopefully clear, API evolves quickly and things that are not checked by compiler and tests become obsolete and/or misleading

here are some potential drawbacks
- there's no integration with `minitest`, so some other testing framework has to be used here, in this case it's `utest`
- some examples will require an import
- it looks a bit more verbose in the doc
- things like `???` no longer work, as they probably will fail the test

there is only one benefit, but it think it's huge - it compiles and runs as a test => documentation is in unison with the code

this is how it looks when you run tests

image
 

What do you guys think of it? Should I continue with it?


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -103,16 +104,17 @@ import scala.util.{Failure, Success, Try}
...
-  *       if (times <= 0) Task.raiseError(err) else {
-  *         // Recursive call, yes we can!
-  *         retryOnFailure(times - 1, source)
-  *           // Adding 500 ms delay for good measure
-  *           .delayExecution(500)
-  *       }
-  *     }
+  *   scala> import scala.concurrent.duration._
+  *   scala> def retryOnFailure[A](times: Int, source: Task[A]): Task[A] =
+  *        | source.onErrorHandleWith { err =>
```
jozic: here `onErrorRecoverWith` is `replaced with `onErrorHandleWith` as it's total and `onErrorRecoverWith` is partial

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -75,18 +75,19 @@ import scala.util.{Failure, Success, Try}
...
   * }}}
   *
   * This `Task` reference will trigger a side effect on evaluation, but
   * not yet. To make the above print its message:
   *
   * {{{
-  *   import monix.execution.CancelableFuture
-  *
-  *   val f: CancelableFuture[Unit] = sayHello.run()
+  *   scala> import monix.execution.Scheduler.Implicits.global
```
jozic: extra import here

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -153,20 +155,20 @@ import scala.util.{Failure, Success, Try}
   * One way of doing that is to process your list in batches:
   *
   * {{{
-  *   // Some array of tasks, you come up with something good :-)
-  *   val list: Seq[Task[Int]] = ???
+  *   // Some array of tasks
+  *   scala> val list: Seq[Task[Int]] = (0 to 100).map(Task.eval(_))
```
jozic: `???` fails the test, so need to be replaced

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -153,20 +155,20 @@ import scala.util.{Failure, Success, Try}
...
   *   // Specify that each batch should process stuff in parallel
-  *   val batchedTasks = chunks.map(chunk => Task.gather(chunk))
+  *   scala> val batchedTasks = chunks.map(chunk => Task.gather(chunk))
   *   // Sequence the batches
-  *   val allBatches = Task.sequence(batchedTasks)
+  *   scala> val allBatches = Task.sequence(batchedTasks)
   *
   *   // Flatten the result, within the context of Task
-  *   val all: Task[Seq[Int]] = allBatches.map(_.flatten)
+  *   scala> val all: Task[Iterator[Int]] = allBatches.map(_.flatten)
```
jozic: `chunks` is iterator, so type need to be fixed here

oleg-py:
This is a very good effort, thanks for getting the ball rolling.

It seems that sbt-doctest supports custom test frameworks:
https://github.com/tkawachi/sbt-doctest#using-%C2%B5test

So maybe it can be done without adding utest?

jozic:
@oleg-py I believe it's possible, but not without integrating minitest with sbt-doctest. Something like https://github.com/tkawachi/sbt-doctest/blob/master/src/main/scala/com/github/tkawachi/doctest/MicroTestGen.scala need to be implemented for minitest

i can take a look and see how hard it's to implement MinitestGen if sbt-doctest is a way to go

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -63,8 +63,8 @@ import scala.util.{Failure, Success, Try}
   * [[monix.eval.Task.apply Task.apply]]:
   *
   * {{{
-  *   val hello = Task.eval("Hello ")
-  *   val world = Task("World!")
+  *   scala> val hello = Task.eval("Hello ")
```
alexandru: I don’t really like having a Scala prompt in code samples. I know it has become a more common practice since “tut”, but code samples need to be copy/paste-able.

I also hope this doesn’t yield any extra output. I’m guessing that it doesn’t.

Can we not do the `scala>` thing?

build.sbt
```diff
@@ -339,8 +339,16 @@ lazy val executionJS = project.in(file("monix-execution/js"))
   .settings(requiredMacroDeps)
   .settings(executionCommon)
 
+lazy val doctestTestSettings = Seq(
+  testFrameworks += new TestFramework("utest.runner.Framework"),
```
alexandru: We can’t depend on utest, we’ll need Minitest integration.

I don’t think that’s too hard.

jozic: I will see what it cost to integrate minitest with sbt-doctest

monix-eval/shared/src/main/scala/monix/eval/Task.scala
jozic: I'm also not a big fan of it, but unfortunately, this is how sbt-doctest works. 
to my understanding, it's can be either

```
 * >>> Test.f(
 * ...   10
 * ... )
 * 20
 ```
or 
```
 * scala> Test.f(
 *      |   20
 *      | )
``` 

I'm not sure how much of an effort would be to change it to work without `scala>`, so I will check it and get back

Comment (alexandru) [reply monix-eval/shared/src/main/scala/monix/eval/Task.scala]: Maybe we should open an issue on the sbt-doctest repository.

File: build.sbt (hunk 6 above)
Comment (jozic) [reply build.sbt]: first attempt to integrate minitest with sbt-doctest https://github.com/tkawachi/sbt-doctest/pull/108

**Comment by jozic** [comment]
So Minitest if now supported as of sbt-doctes 0.8.0 (https://github.com/tkawachi/sbt-doctest/pull/108)
we have two ways now
- go with what we have now, i.e. `scala>` prompt in tut style 
- change sbt-doctest to take all the code inside {{{}}} and put it in a test. Only thing those "tests" will test is that it compiles (which is already big)

@alexandru @oleg-py what do you guys thinkg?

@tkawachi would you be open to extend sbt-doctest to treat all code inside {{{}}} as Verbatim (in sbt-docs terms)? I can try it if it sounds acceptable to you

**Comment by tkawachi** [comment]
@jozic First of all, thank you for trying sbt-doctest. I'm open to the idea if it doesn't break backward compatibility.

**Comment by jozic** [comment]
so I gave it a try and with approach where code from scaladocs inside {{{}}} is directly injected in minitest's `test()` then some of the examples fail to compile because of errors like `Problem with test spec, expecting `Unit`, but received: Coeval.FlatMap$1118651723  (CoevalDoctest.scala:44)`

So one solution I can think of is to modify sbt-doctest to add following code to generated specs 

```scala
implicit def toVoid[A](ref: A)(implicit location: _root_.minitest.api.SourceLocation): _root_.minitest.api.Void = _root_.minitest.api.Void.UnitRef
```

Do you guys have better ideas?

oleg-py:
@jozic I think @alexandru is on a vacation now. I'll review the updated PR in the next two days.

monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
```diff
@@ -74,35 +74,35 @@ import scala.util.{Failure, Success, Try}
   * instances out of strict values:
   *
   * {{{
-  *   val fa = Coeval.now(1)
-  *   fa.value //=> 1
+  *   val faNow = Coeval.now(1)
+  *   faNow.value //=> 1
```
oleg-py: Nitpick 1: I'd prefer `// =>` style, if that's possible
Nitpick 2: `value` now uses parens, so it should be `value()`.

monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
```diff
@@ -74,35 +74,35 @@ import scala.util.{Failure, Success, Try}
...
+  *   feThrowNow.failed //=> has RuntimeException
   * }}}
   *
   * The "always" strategy is equivalent with a plain function:
   *
   * {{{
   *   // For didactic purposes, don't use shared vars at home :-)
   *   var i = 0
-  *   val fa = Coeval.eval { i += 1; i }
+  *   val faEval = Coeval.eval { i += 1; i }
```
oleg-py: Why rename?

monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
```diff
@@ -74,35 +74,35 @@ import scala.util.{Failure, Success, Try}
...
   * }}}
   *
   * The "once" strategy is equivalent with Scala's `lazy val`
   * (along with thread-safe idempotency guarantees):
   *
   * {{{
-  *   var i = 0
-  *   val fa = Coeval.evalOnce { i += 1; i }
+  *   var j = 0
+  *   val faEvalOnce = Coeval.evalOnce { j += 1; j }
```
oleg-py: Same question: why rename? `fa` is a typical name for "some effect with some value". I'd either keep it as is, or name it `coevalOnce` or sth like that.

monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
```diff
@@ -771,15 +775,15 @@ sealed abstract class Coeval[+A] extends (() => A) with Serializable { self =>
     * of 10 times before giving up:
     *
     * {{{
-    *   import scala.concurrent.duration._
+    *   val fa = Coeval(78)
```
oleg-py: This could use a better example, e.g. with `Random.nextInt(20)` failures and a `var`, so it is actually visible what's going on.

monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
jozic: that should be easy

monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
jozic: `fa` is already defined in other code block in same Scaladoc, they appear in same test and compiler will not let it go
name can be something else, but we can't use same names in same Scaladoc

monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
jozic: `coevalOnce` will do as long as it's not repeated in same Scaladoc

jozic:
@oleg-py does it look better now? I think i addressed addressable comments, renaming has to be done for the same Scaladoc though

@alexandru we can keep ScalaDocs without `scala>` prompts now

I bumped sbt-scalafmt as the old one depends on conflicting version of scalameta

jozic:
cool, thanks, i will work on covering Scaladocs for more classes

alexandru:
@jozic @tkawachi, folks, this is really cool ... after merging this in `master`, I got a Travis build failure on another PR because indeed I had a broken example.

OMG, this is so awesome!

alexandru:
And this integration is a match made in heaven, as this project has always prided itself with great ScalaDocs.

alexandru:
Can you integrate it in Cats-Effect as well?

https://github.com/typelevel/cats-effect

jozic:
sure, it's possible
cats uses it for a long time, but they use it in original style with `scala>` prompt
not sure if maintainers will want to change that

jozic:
i'm working on fixing rest of the Scaladocs for monix-eval project, after that i can try cats-effect

alexandru:
Thanks. For sure, focus on Monix first 😛

PR #646, opened 2018-04-12
Fix ack.future called when ack is null

oleg-py:
A bug from gitter with .delayOnNextBySelector. See unit test for previously-failing case.


Avasil:
For historical purposes:
https://gitter.im/monix/monix?at=5acf26201130fe3d36cbe2dc

alexandru:
Will get released in RC2 — hopefully on the 16-th.

Issue #644, opened 2018-04-10
Optimize Observable.concat and Observable.concatDelayError


alexandru:
Currently we've got these functions on the companion object:

```scala
  def concat[A](sources: Observable[A]*): Observable[A] =
    Observable.fromIterable(sources).concatMap[A](identity)

  def concatDelayError[A](sources: Observable[A]*): Observable[A] =
    Observable.fromIterable(sources).concatMapDelayErrors[A](identity)
```

They are defined in terms of `concatMap`, however we can define them in terms of `++`, the strict implementation (not the one with the by-name param, after #643 happens).

So the challenge, should you choose to accept it:

1. provide an implementation based on `ConcatObservable` instead of `fromIterable` and `concatMap`
2. prove that it is better by a benchmark, you can dump a quick benchmark in the `benchmarks` sub-project ... but don't commit it in the PR

agajek:
I've tried to provide some implementation by stacking `sources` into `ConcatObservable`. It ended up with sth like that:
```
def concat[A](sources: Observable[A]*): Observable[A] =
    sources.toList.foldLeft(Observable.empty[A])((acc, a) => new ConcatObservable[A](acc, a))
```
In benchmark I've generated 3k `Observable` containing pure `Int` and then applied my implementation of concat to them. But the performance was terrible.
Current implementation is:
```
ObservableConcatBenchmark.concatObservable    3000  thrpt   20  139531619.493 ± 6637768.282  ops/s
```
and mine:
```
ObservableConcatBenchmark.concatObservable    3000  thrpt   20  12198.799 ± 1458.540  ops/s
```
Is the sth I need to put attention while implementing this?

alexandru:
No longer relevant.

PR #654, opened 2018-04-15
Make Observable#++'s argument lazy

### LINKED ISSUE #643: Make Observable's ++ take a lazy (by-name) parameter
Similar to #642, on `Observable` we've got the `++` operator currently defined like this:

```scala
abstract class Observable[+A] {
  //...
  def ++[B >: A](other: Observable[B]): Observable[B]
}
```

We need to change that parameter to be by-name (lazy):

```scala
def ++[B >: A](other: => Observable[B]): Observable[B]
```

kubukoz:
Fixes #643.

Done as part of the Scala Spree at https://scala.sphere.it/ :D

### DIFF
--- a/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
+++ b/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
@@ -52,9 +52,11 @@ object RecursiveConcatSuite extends BaseOperatorSuite {
   }
 
   val nats: Observable[Long] = {
-    Observable.now(1L) ++ nats.map(_ + 1)
+    def loop(acc: Long): Observable[Long] =
+      Observable.now(acc) ++ loop(acc + 1)
+    loop(1)
   }
-
+  
   test("laziness of ++'s param") { implicit s =>
     val count = 1000000L
 
--- a/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
+++ b/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
@@ -56,7 +56,7 @@ object RecursiveConcatSuite extends BaseOperatorSuite {
   test("laziness of ++'s param") { implicit s =>
     val count = 10000
 
-    import concurrent.duration._
-    assertEquals(nats.take(count).sumL.runSyncUnsafe(1.second), (1 to count).sum)
+    val f = nats.take(count).sumL.runAsync; s.tick()
+    assertEquals(f.value, Some(Success((1 to count).sum)))
   }
 }
--- a/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
+++ b/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
@@ -50,4 +50,13 @@ object RecursiveConcatSuite extends BaseOperatorSuite {
     val f = range(0, count).sumL.runAsync; s.tick()
     assertEquals(f.value, Some(Success(count.toLong * (count - 1) / 2)))
   }
+
+  val nats: Observable[Long] = Observable.now(1L) ++ nats.map(_ + 1)
+
+  test("laziness of ++'s param") { implicit s =>
+    val count = 10000
+
+    import concurrent.duration._
+    assertEquals(nats.take(count).sumL.runSyncUnsafe(1.second), (1 to count).sum)
+  }
 }
--- a/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
+++ b/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
@@ -51,7 +51,9 @@ object RecursiveConcatSuite extends BaseOperatorSuite {
     assertEquals(f.value, Some(Success(count.toLong * (count - 1) / 2)))
   }
 
-  val nats: Observable[Long] = Observable.now(1L) ++ nats.map(_ + 1)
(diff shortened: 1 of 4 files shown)


kubukoz:
I see that the build fails when compiling for JS - will take a look

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
```diff
@@ -50,4 +50,13 @@ object RecursiveConcatSuite extends BaseOperatorSuite {
...
     assertEquals(f.value, Some(Success(count.toLong * (count - 1) / 2)))
   }
+
+  val nats: Observable[Long] = Observable.now(1L) ++ nats.map(_ + 1)
+
+  test("laziness of ++'s param") { implicit s =>
+    val count = 10000
+
+    import concurrent.duration._
+    assertEquals(nats.take(count).sumL.runSyncUnsafe(1.second), (1 to count).sum)
```
Avasil: Using `TestScheduler` we can manipulate time and instead of `runSyncUnsafe` use just `runAsync` and then `s.tick(1.second)` (test above) - it might also help with passing Scala.JS test :)

kubukoz: I'm now trying with `tick()` following the test above's way of doing it :)

kubukoz:
@Avasil gotcha.

The semantics for `startWith` wouldn't have changed much (`self` was passed as the rhs, so it would already have been evaluated), but perhaps it'll be more efficient to use `concat` in it anyways, so I did that

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
```diff
@@ -50,4 +50,15 @@ object RecursiveConcatSuite extends BaseOperatorSuite {
     val f = range(0, count).sumL.runAsync; s.tick()
     assertEquals(f.value, Some(Success(count.toLong * (count - 1) / 2)))
   }
+
+  val nats: Observable[Long] = Observable.defer {
```
Avasil: Ideally we could just do `Observable.now(1L) ++ nats.map(_ + 1)` - if `Observable.defer` is needed then maybe we could just include it in the implementation? So we could even leave old `ConcatObservable` but pass suspended `Observable`

kubukoz: Yeah, I was following the convention from `def range` above - but I think with `++` having a lazy rhs it doesn't make sense to defer either of these observables. Unless we make `range` use `concat` instead too...

Avasil: I mean that we could leave `ConcatObservable` strict (and use it in `concat` implementation if possible - #644) implementing `++` like:
```scala
new ConcatObservable(self, Observable.suspend(other))
```
What do you think?

kubukoz: it could work - I'll see

kubukoz: It works, but I'm having some random SOs from time to time - looking for a reason

kubukoz: Apparently it has something to do with `defer` not trampolining the function passed to it - but I'm not sure about the details

Avasil:
@alexandru There seems to be problem with `Observable.++` stack-safety, we've tried different implementations and all of them(using by-name, deferring `Observable` etc.) were failing due to `StackOverflow` after running test many times.

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/RecursiveConcatSuite.scala
```diff
@@ -48,6 +48,17 @@ object RecursiveConcatSuite extends BaseOperatorSuite {
   test("stack safety") { implicit s =>
     val count = 10000
     val f = range(0, count).sumL.runAsync; s.tick()
-    assertEquals(f.value, Some(Success(count.toLong * (count - 1) / 2)))
+    assertEquals(f.value, Some(Success(sum(count))))
+  }
+
+  val nats: Observable[Long] = {
+    Observable.now(1L) ++ nats.map(_ + 1)
```
alexandru: This PR is really old, sorry for coming back so late. I'm not a good PR manager 😐

Anyway, so one problem is that this `nats` definition is not _tail recursive_, therefore the application of `++` cannot be. Actually in this case that `Observable` throws `StackOverflowException` is a feature, because otherwise you have a memory leak on your hands.

A tail-recursive `nats` definition would go like this:

```scala
val nats: Observable[Long] = {
  def loop(acc: Long): Observable[Long] =
    Observable.now(acc) ++ loop(acc + 1)
  loop(1)
}
```

alexandru:
I changed `nats` to this, will see if the build passes:

```scala
val nats: Observable[Long] = {
  def loop(acc: Long): Observable[Long] =
    Observable.now(acc) ++ loop(acc + 1)
  loop(1)
}
```

alexandru:
A merge with `master` might be needed.

kubukoz:
Rebased to master :) yeah, now I see that mistake with the tail recursion.

kubukoz:
Thanks for the commit, btw!

Avasil:
Seems like it's just missing 
```
ProblemFilters.exclude[IncompatibleMethTypeProblem]("monix.reactive.Observable.++")
```

kubukoz:
Hopefully that'll be enough ;) thanks @avasil

Avasil:
Thanks and we can finally merge it! @kubukoz

PR #655, opened 2018-04-15
Handle InterruptedException in NonFatal

### LINKED ISSUE #645: Properly deal with InterruptedException on the JVM
Copying the issue from https://github.com/typelevel/cats-effect/issues/181 because Monix suffers from the same issue.

We have our own [NonFatal](https://github.com/monix/monix/blob/master/monix-execution/shared/src/main/scala/monix/execution/misc/NonFatal.scala) implementation because the standard one also filters things like `ControlThrowable`, which is pretty dumb.

However this means we are no longer catching `InterruptedException`, which is a problem.

The problem is that on top of the JVM, when you catch an `InterruptedException`, you can't just swallow it like any other exception. This is a problem:

```scala
try {
  operation()
} catch {
  case ignored: InterruptedException => /* BIG PROBLEM! */
}
```

What needs to happen is that we need to "restore the interrupted status". This can be done like this:

```scala
try {
  operation()
} catch {
  case ignored: InterruptedException =>
    // Restores interrupted status
    Thread.currentThread().interrupt()
}
```

So we need a `NonFatal` specific to the JVM that, upon catching `InterruptedException`, it restores its interrupted status.

kubukoz:
Fixes #645, so far it's using a shared implementation, but could be changed to have a separate one for JS and JVM worlds.

### DIFF
--- a/monix-execution/shared/src/main/scala/monix/execution/misc/NonFatal.scala
+++ b/monix-execution/shared/src/main/scala/monix/execution/misc/NonFatal.scala
@@ -29,6 +29,9 @@ object NonFatal {
   def apply(t: Throwable): Boolean = t match {
     // VirtualMachineError includes OutOfMemoryError and StackOverflowError
     case _: VirtualMachineError => false
+    case _: InterruptedException =>
+      Thread.currentThread().interrupt()
+      true
     case _ => true
   }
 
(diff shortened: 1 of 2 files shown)


Avasil:
Thanks @kubukoz !

@alexandru I'm merging this, if we decide to go different route with `NonFatal` (like triggering cancelation) we can always do another change

Issue #642, opened 2018-04-10
Make Iterant's ++ take a lazy (by-name) parameter


alexandru:
Currently the signature for `++` for `Iterant` is this:

```scala
sealed abstract class Iterant[F[_], A] {
  //...

  def ++[B >: A](rhs: Iterant[F, B])(implicit F: Applicative[F]): Iterant[F, B]
}
```

We need to change it to take a by-name parameter:

```scala
def ++[B >: A](rhs: => Iterant[F, B])(implicit F: Applicative[F]): Iterant[F, B]
```

Note that via this change we might slow down operators that depend on `++`, so we need to provide a strict alternative. Might as well provide a function on the companion object:

```scala
object Iterant {
  // ...

  def concat[F[_], A](xs: Iterant[F, A]*): Iterant[F, A] = ???
}
```

This needs to be implemented in the most efficient way possible, so it should not be in terms of `flatMap`, but in terms of the current `++` implementation.

Also search and replace in the code-base any usage of `++` with `concat` (I don't think there are any, but should be double checked).

Avasil:
Just commenting that one of the Scala Spree participants is working on this issue. :)

Avasil:
Resolved in #656

PR #656, opened 2018-04-15
Make Iterant's ++ take a lazy (by-name) parameter #642

kamilkloch:
Please note of the following lazy ++ performance issue:
```scala
val n = 1000
lazy val nats: Iterant[Coeval, Long] = Iterant[Coeval].of(1L) ++ nats.map(_ + 1L) take(n)
nats.sumL.value()
```
The following code has running time O(n^2) (as opposed to O(n) for the analogous `Stream` example, due to Stream memoization).

### DIFF
--- a/monix-tail/shared/src/test/scala/monix/tail/IterantConcatSuite.scala
+++ b/monix-tail/shared/src/test/scala/monix/tail/IterantConcatSuite.scala
@@ -62,4 +62,20 @@ object IterantConcatSuite extends BaseTestSuite {
     assertEquals(nats.toListL.value(), List(1, 2, 3, 4))
   }
 
+  test("Iterant.concat(Iterant*)") { implicit s =>
+    check1 { (ll: List[List[Int]]) =>
+      val li = ll.map(Iterant[Coeval].fromList)
+      val concat = Iterant.concat(li: _*)
+      val expected = Iterant[Coeval].fromList(ll.flatten)
+      concat <-> expected
+    }
+  }
+
+  test("Iterant.concat is eager") { implicit s =>
+    lazy val i: Iterant[Coeval, Int] = Iterant.concat(Iterant[Coeval].now(1), i)
+    intercept[StackOverflowError] {
+      i.take(1).sumL.value()
+    }
+  }
+
 }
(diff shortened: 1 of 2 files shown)


Avasil:
Fixes #642

Thanks @kamilkloch - could you also add `concat` in companion object?
```scala
object Iterant {
  // ...

  def concat[F[_], A](xs: Iterant[F, A]*): Iterant[F, A] = ???
}
```

kamilkloch:
@Avasil Added Iterant.concat.

monix-tail/shared/src/test/scala/monix/tail/IterantConcatSuite.scala
```diff
@@ -56,4 +56,26 @@ object IterantConcatSuite extends BaseTestSuite {
...
+      val li = ll.map(Iterant[Coeval].fromList)
+      val concat = Iterant.concat(li: _*)
+      val expected = Iterant[Coeval].fromList(ll.flatten)
+      concat <-> expected
+    }
+  }
+
+  test("Iterant.concat is eager") { implicit s =>
+    lazy val i: Iterant[Coeval, Int] = Iterant.concat(Iterant[Coeval].now(1), i)
+    intercept[StackOverflowError] {
```
Avasil: I don't think we need this test

monix-tail/shared/src/test/scala/monix/tail/IterantConcatSuite.scala
```diff
@@ -56,4 +56,26 @@ object IterantConcatSuite extends BaseTestSuite {
       (i :+ e) <-> (i ++ Iterant[Coeval].pure(e))
     }
   }
+
+  test("Iterant ++ Iterant rhs by name") { implicit s =>
```
Avasil: What do you think about renaming it to something that mentions that this is test for stack safety?

monix-tail/shared/src/test/scala/monix/tail/IterantConcatSuite.scala
kamilkloch: OK, removed.

monix-tail/shared/src/test/scala/monix/tail/IterantConcatSuite.scala
kamilkloch: Any propositions?

Avasil: "(...) is stack safe" :D

kamilkloch: Done.

Avasil:
@kamilkloch Looks good, I'll merge when tests pass :)

PR #660, opened 2018-04-22
Make TaskApp scheduler and options defs

oleg-py:
We discussed this at gitter a while ago

Benefits of using `def`s:

* You can override them with an `implicit val`, which is quite relevant for options after #634. Doing so will make `ConcurrentEffect[Task]` instance materializable with expected configuration.
* You can do simple alterations by doing `super`, e.g.
```scala
override val scheduler = super.scheduler.withExecutionModel(AlwaysAsyncExecution)
override val options = super.options.enableLocalContextPropagation
```


alexandru:
👍

PR #661, opened 2018-04-22
Proposal: make mapBoth always execute tasks in parallel

oleg-py:
Got caught up by surprise by the following behavior:

```scala
import cats._, implicits._, effect._, monix.eval._
import java.util.concurrent.TimeUnit

// Imagine this is some blocking operation
val blockingOp = Task.eval { Thread.sleep(1000) }

val nowL = Timer[Task].clockMonotonic(TimeUnit.SECONDS)

val task = for {
  start <- nowL
  _ <- Task.zip4(blockingOp, blockingOp, blockingOp, blockingOp)
  end <- nowL
  _ <- Task.eval(println(s"${end - start}s passed")) // 4s passed, expected: 1s
} yield ()

task.runSyncUnsafe(Duration.Inf)
```

This happens if Task is constructed by `.eval` as opposed to `.apply` (which I do all the time since this way I can control behavior later). In real example I have a flatMap-chain of `Task`s built on `Task.eval`.

This affects `Task.zipX`, `Task.parMapX` and also `Parallel` instance. I think it makes sense to always fork inside `mapBoth` because it's more likely to be what user expects.

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/internal/TaskMapBoth.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/internal/TaskMapBoth.scala
@@ -84,9 +84,8 @@ private[eval] object TaskMapBoth {
         val context2 = context.copy(connection = task2)
         mainConn push Cancelable.collection(Array(task1, task2))
 
-        // Light asynchronous boundary; with most scheduler implementations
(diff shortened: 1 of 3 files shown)


alexandru:
I agree with this change.

Had to force merge it because Travis is malfunctioning.

PR #662, opened 2018-04-22
Add Iterant.fromReactivePublisher

### LINKED ISSUE #618: Iterant.fromReactivePublisher
The Reactive Streams integration isn't complete.

We can convert from `Iterant` to `org.reactivestreams.Publisher`, but not vice-versa. To convert from `Publisher` to `Iterant` we can do it in two ways:

1. behavior similar with `Observable.bufferIntrospective`, as in we could use a buffer that collects events from the source up to a limit, then when the `Iterant`'s consumer requests the next batch we drain the buffer — this approach is pretty cool because it will serve whatever elements are buffered at any moment, but it's harder to implement, since it requires synchronization for the case when the `Iterant`'s consumer demands elements but there are no elements buffered, so it would have to wait on some kind of signal for the buffer has elements in it
2. a much simpler implementation that does a `publisher.request(n)`  and only serves a `NextBatch` when all `n` requested events have been buffered — this would be very simple to implement and deterministic, however it affects the API (e.g. the `n` buffer size is now mandatory as a parameter) and it's not very responsive if the source is slow, although it could be argued that in case the source is slow we can just work with `n = 1`

At this point I'm actually leaning towards solution 2, because it's easier to implement and I'm a little disgusted at the thought of working again with concurrent buffers, as we have to basically implement something very much like the `BackpressuredBuffer` in `monix-reactive`.

I wouldn't mind solution nr 1 though. Maybe we should provide both, first nr 2, then leave room for nr 1. Maybe the `fromReactivePublisher` should provide some sort of "pull strategy" and I'm thinking right now of:

1. `Single`
2. `Batched`
3. `Introspective`

The behavior of `Single` would be that of `Batched(size = 1)`, but without having to allocate an actual internal buffer. Not sure if it's worth it to single out, or just assume that behavior on `Batched(size = 1)`.

Interestingly if we are converting from `Observable`, then we don't actually need `Introspective`.
Assuming we also have `mapBatch` from https://github.com/monix/monix/issues/617 we could then do:

```scala
val arrays = Iterant[Task].fromReactivePublisher(observable.bufferIntrospective(n))

arrays.mapBatch(Batch.fromArray(_))
```

So this highlights that the priority is the simple version, with the more complicated version being a nice to have.

oleg-py:
Closes #618.

This uses `AsyncVar` from Monix itself which is very nice for providing a necessary backpressure / waiting for that sort of conversion, as compared to messing with Promises / callbacks directly.


monix-tail/shared/src/main/scala/monix/tail/internal/IterantFromReactivePublisher.scala
```diff
@@ -0,0 +1,112 @@
...
+    private[this] val result = AsyncVar.empty[Iterant[F, A]]
+    private[this] var subscription: Subscription = _
+    private[this] var offset = 0
+    private[this] var buffer: Array[Any] =
+      if (bufferSize == 1) null else new Array(bufferSize)
+
+    val generate: F[Iterant[F, A]] = F.async[Iterant[F, A]] { cb =>
+      result.take
+        .onComplete(try_ => cb(try_.toEither))(immediate)
+    } <* timer.shift
```
oleg-py: This is the only thing I use `Timer` for. This way I can be sure that continuation goes on on a user-defined pool (Timer.shift does not require to pass a EC, unlike `Async.shift`), but it might not be strictly required.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantFromReactivePublisher.scala
```diff
@@ -0,0 +1,112 @@
...
+    F: Async[F],
+    timer: Timer[F]
+  )
+    extends Subscriber[A]
+  {
+    private[this] val result = AsyncVar.empty[Iterant[F, A]]
+    private[this] var subscription: Subscription = _
+    private[this] var offset = 0
+    private[this] var buffer: Array[Any] =
+      if (bufferSize == 1) null else new Array(bufferSize)
```
oleg-py: I'm using array here for memory efficiency, but it might not be the best idea if user passes in a buffer size of `Int.MaxValue`. Maybe `ArrayBuffer` is a better solution, at least if `bufferSize` is greater than some reasonably small number.

monix-tail/shared/src/test/scala/monix/tail/ArbitraryInstances.scala
```diff
@@ -96,18 +96,16 @@ trait ArbitraryInstances extends monix.eval.ArbitraryInstances {
       }
     }
 
-  implicit def isEqIterantTask[A](implicit A: Eq[A]): Eq[Iterant[Task, A]] =
+  implicit def isEqIterantTask[A](implicit A: Eq[A], sc: TestScheduler): Eq[Iterant[Task, A]] =
     new Eq[Iterant[Task, A]] {
       def eqv(lh: Iterant[Task,  A], rh: Iterant[Task,  A]): Boolean = {
-        implicit val s = TestScheduler()
```
oleg-py: I had to remove these local `TestScheduler`s and require a test suite to provide one.

The reason was that, say I wanted to use `Iterant.toReactivePublisher`, which takes a `Scheduler`, and I have one in test suite. To actually run it, I need to call a `.tick()` on a test scheduler. In other methods, `equalityFuture` will internally do it, but here it calls `.tick()` on a different scheduler, and the `Future` I got from `toReactivePublisher` would not be completed.

In particular, this intuitive test:
```scala
Iterant[IO].fromReactivePublisher(it.toReactivePublisher) <-> it
```
would fail, because LHS would not be completed.

monix-tail/shared/src/main/scala/monix/tail/PullStrategy.scala
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.tail
+
+sealed abstract class PullStrategy extends Product with Serializable
```
alexandru: I would also name it `ReactivePullStrategy` or something like that, because it's very much related to the `reactivestreams.org` protocol. If we'll add another `Introspective` strategy or something, it will only make sense via this protocol integration.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantFromReactivePublisher.scala
```diff
@@ -0,0 +1,113 @@
...
+
+    val generate: F[Iterant[F, A]] = F.async[Iterant[F, A]] { cb =>
+      result.take
+        .onComplete(Callback.fromAttempt(cb))(immediate)
+    } <* timer.shift
```
alexandru: I think this could use an optimization, this being effectively the implementation of `MVar#take`. Avoids going through `Promise`:

```scala
      F.async { cb =>
        val cb2 = Callback.fromAttempt(cb)
        av.unsafeTake(cb2) match {
          case null => () // do nothing
          case a => async(Right(a))
        }
      }
```

Issue #594, opened 2018-02-16
Async Tasks Submitted to Full Run Queues are Orphaned


pettyjamesm:
Just hit this today and can submit more details later, but thought I'd submit the issue anyway in case the fix is easy on my side. Tasks that asyncBoundary and executeOn (eg: IO schedulers) stop executing entirely if the IO scheduler throws a rejected execution exception.

pettyjamesm:
I should say, in our case this is a database IO pool, thread-sized against the connection pool size, and where we actually want the rejected execution behavior to prevent unbounded growth of the thread pool beyond a certain queue limit. This is especially painful during auto scaling where fresh instances have yet to hit the server side prepared statement threshold and are still warming up the JIT, but are coming online under situations of heavy load.

alexandru:
@pettyjamesm sorry for the delay in response.

It's pretty bad if the execution context throws an error and the task will probably become non-terminating, because it's currently a protocol violation — consider that the callback given in `Task.create` needs to be called at most once, so once you pass that callback to another piece of logic, if any of that logic throws, then in absence of further synchronization (which is slow), all you can safely do is to log the error.

I'm not a fan of using errors like that for limiting growth. It's better if you have some back-pressure mechanism.

But please describe the wanted behavior. I'm guessing that you want the task to end in error. I guess that if there's a specific error that the thread pool throws, we could recognize it and treat it as a case where we can signal the error via the callback.

So if the solution is straightforward, I'm willing to accept a PR.

alexandru:
Note that we might be able to treat just `Task.shift` / `Task.delay*` / `Task#executeAsync` and leave `Task.create` alone — because there it's the user's responsibility to catch and treat errors.

pettyjamesm:
Closing issue as likely resolved by https://github.com/monix/monix/pull/666 . As mentioned in that PR, our code base already works around the issue by using Task.create directly, but the test suite seems reasonable to guarantee the implementation handles this case now. Thanks for the great open source project.

PR #666, opened 2018-04-29
Make Task propagate RejectedExecutionException

### LINKED ISSUE #651: Task should deal with RejectedExecutionException
In issue #594 we are described a situation in which thread-pools that can throw
[RejectedExecutionException](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/RejectedExecutionException.html) are used.

But if that happens in methods such as `executeAsync`, `executeOn`, `shift`, or `asyncBoundary`, we aren't catching and streaming the exception and the tasks will become non-terminating.

In these functions dealing with `Scheduler`, we need to catch exceptions that are triggered by `Scheduler#execute` and stream it as an error via the provided callback. It's safe doing so because execute is supposed to be async, so if `execute` throws then we can be sure that the callback wasn't called.

oleg-py:
Fix #651, #594.

Getting `RejectedExecutionException` requires a manually created Executor with a `BlockingQueue` of finite size. It cannot be done with monix Scheduler factory methods nor with Executors ones.

Scala.js has `REE` too: https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/concurrent/Throwables.scala#L22, so no platform-specific code is required.


oleg-py:
It's a bit complicated, because it only works for shift-based methods and executeOn. It's not enough if you try to use a bounded queue-based executor as your main one.

oleg-py:
@alexandru seems to not run correctly on Travis.

I can create a fake ExecutorService that always throws RejectedExecutionException, tho it might make the test less... legitimate.

alexandru:
@oleg-py I think creating a fake `ExecutionContext` is totally fine. We just need to catch and handle that particular exception, no complicated logic there.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -2174,12 +2176,17 @@ object Task extends TaskInstancesLevel1 {
...
+      try {
+        ec2.execute(new Runnable {
+          def run(): Unit = {
+            context.frameRef.reset()
+            cb.onSuccess(())
+          }
+        })
+      } catch {
+        case ex: RejectedExecutionException =>
+          cb.onError(ex)
```
alexandru: Works, but use `cb.asyncOnError(ex)` as it's safer — I make it a habit of doing that in `Async` states in order to ensure stack safety, as you never know how it ends up being used. It's not needed for `onSuccess` here because you have `ec2.execute` first.

alexandru:
This PR diverged from master unfortunately.

pettyjamesm:
I should say, since I reported the issue that in our case we choose to stop making the database ThreadPool implement Scheduler and just use Task.create with a surrounding try / catch to handle this case.

The next surprise was discovering that the callback would start running the subsequent logic on the database ThreadPool threads, but I know that master already changed the behavior to shift back to the default scheduler. Any idea when a new RC will be released? I'd like to make sure that our tracing implementation (using Local) and workaround still work after all the internal changes I've seen merged lately.

PR #667, opened 2018-05-05
WIP: Observable.bracket

### LINKED ISSUE #649: Add Observable.bracket, similar with Iterant.bracket
Currently `Iterant` has this function described in its companion object:

```scala
def bracket[F[_], A, B](acquire: F[A])(use: A => Iterant[F, B], release: A => F[Unit])
  (implicit F: Sync[F]): Iterant[F, B] = ???
```

So given a resource created by `acquire`, it feeds that into `use` to create an `Iterant` and then at the end executes `release` no matter what — i.e. even if the consumer stops the streaming early, or the stream ends in success or error.

We need to describe the same for `Observable`:

```scala
def bracketTask[A, B](acquire: Task[A])
  (use: A => Observable[B], release: A => Task[Unit]): Observable[A]
```

Implementation notes:

- the returned `Observable` is cancelable and cancelation needs to take into account at least `acquire`, which yields a task that could also be cancelable
- it needs to trigger `release` on early stop (by noticing `Stop` events in `onNext`)
- it needs to trigger `release` on `onComplete` and `onError`
- it needs to trigger `release` the observable gets cancelled
- `release` can be executed only once and cancelation is concurrent with whatever happens in onNext/onComplete/onError, so some synchronization is required

To use for inspiration:

- [doOnTerminate](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DoOnTerminateOperator.scala)
- [doOnSubscriptionCancel](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DoOnSubscriptionCancelObservable.scala)

Avasil:
Fixes #649.

I might need some help to finish it - I've based it on current `Iterant` implementation but so far it doesn't play well with cancelation in case of `Observable`. When I fix this I also need to consider if using hot `Observable` doesn't mess up any logic. And I'm pretty sure I lose some errors. :D 

On the bright side all `Bracket` laws are passing without any issue. :)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -482,6 +482,20 @@ abstract class Observable[+A] extends Serializable { self =>
   final def bufferIntrospective(maxSize: Int): Observable[List[A]] =
     new BufferIntrospectiveObservable[A](self, maxSize)
 
+  /** Implementation of `bracket` from `cats.effect.Bracket`.
+    *
+    * See [[https://typelevel.org/cats-effect/typeclasses/bracket.html documentation]].
```
Avasil: I've followed `Iterant`'s doc but I was thinking about copying scaladoc from `cats.effect.Bracket` - is it ok/legal? I think it might be easier for users just browsing it in IDE and wondering what it does

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/DoOnExitCaseObservable.scala
```diff
@@ -0,0 +1,109 @@
...
+              } catch {
+                case err if NonFatal(err) =>
+                  scheduler.reportFailure(err)
+                  out.onError(err)
+              }
+        }
+      }
+    })
+
+    Cancelable { () =>
```
Avasil: This is never called with current `resource` implementation

monix-reactive/shared/src/test/scala/monix/reactive/internal/builders/ResourceObservableSuite.scala
```diff
@@ -0,0 +1,295 @@
...
+      val f = method(broken).runAsync
+      s.tick()
+      assertEquals(f.value, Some(Failure(dummy)))
+    }
+
+    assertEquals(rs.acquired, completes.length * 3)
+    assertEquals(rs.released, completes.length * 3)
+  }
+
+  test("Observable.resourceTask does not require non-strict use") { implicit s =>
```
Avasil: I took this from tests for `Iterant` but in case of `Observable` it's non-deterministic and I believe it's not something we expect to hold? Builders like `Observable.pure` call `onComplete` straight away after subscription leading to releasing resources. Or I might be understanding it wrong and it's something wrong with my implementation

alexandru:
@Avasil I finally got around to implementing `bracket` for `Observable`, because we badly need it... I discovered for example that at least 2 `Observable` implementations are broken in combination with a `Task` that auto-cancels (and I'm talking about the `Observable` implementations that wrap `InputStream` and `Reader`).

I'm have many other changes for `Observable`, to upgrade its interface to make it safer (less side-effecty ops).  I will integrate your PR in it because this is good work. I will change your `DoOnExitCaseObservable.scala` because I've got some improvement ideas. And will see what else, but it's pretty close to what we need.

Avasil:
@alexandru 
Sure, feel free to either go from this branch or merge it in different one. Probably tests can be reused

alexandru:
Merged in https://github.com/monix/monix/pull/729

alexandru:
I merged your changes on master along with mine. Am closing this.

PR #668, opened 2018-05-07
Task.mapBoth scaladoc doesn't mention parallel evaluation for sync Task

Avasil:
After #661 `Task.mapBoth` will force parallel evaluation even for synchronous `Task` but the change is missing from Scaladoc.


alexandru:
Thanks for the correction @Avasil, but I already added a mention in #670, as I did so in bulk for all the operations modified.

PR #673, opened 2018-05-16
Proposal: Fix auto-cancelable loophole

alexandru:
Changes:

1. remove ability to set custom `Task.Options` in `Effect`; unfortunately due to auto-cancelable run-loops, the `Concurrent` laws are failing and we have to pick a side — as an example of what can happen, `Iterant` is basically broken, see discussion at https://github.com/typelevel/cats-effect/issues/192
2. remove ability to set auto-cancelable run-loops as part of the Java environment
3. fix `Task.bracket` to work when auto-cancelable run-loops are active
4. fix `TaskLocal.bind` to use `bracket` properly
5. make `TaskLocal.bind` to activate `localContextPropagation` when executing the given task (within its context), thus making `TaskLocal` work when used with binds without having to explicitly enable `localContextPropagation`
6. renamed `Task#cancelable` to `Task#autoCancelable`, to differentiate it from the `Task.cancelable` builder (and provide the old symbol `@deprecated`)
7. deprecate `executeWithOptions` and `runAsyncOpt` methods, basically everything that allows specifying `Options`

/cc @oleg-py 

Unfortunately I had to revert the changes in #634. If we make `Effect[Task]` auto-cancelable, it violates cats-effect's laws and we cannot have that.


alexandru:
@oleg-py I'll let this PR open for a while, doesn't need to get shipped in RC2.

But first of all, what if I told you that I found a way to enable `localContextPropagation` whenever a `TaskLocal` is used? Which would make it effectively enabled all the time.

I just pushed a new Task state in this PR called `ContextSwitch` that should make changing these settings internally really cheap. And so we can enable the local context propagation option when we create a `TaskLocal`, or on each `write` or `bind`.

Along with this change, I also deprecated `executeWithOptions` and `runAsyncOpt` 🙂

> A consensus is still that laws need to be changed to accommodate this and Scalaz 8 IO, if I remember correctly. If we can fix other tests, that might be essential to figuring out which laws need adjustment.

I will want to talk about that. Did not have the energy, plus I've been sleeping on it.

However, the issue is that auto-cancelable run-loops are breaking `Iterant`'s encoding. I described the issue in https://github.com/typelevel/cats-effect/issues/192 and in absence of a reasonable solution to describe a safe `Iterant[F, A]` that is terminated properly, by evaluating `stop` whenever a folded `F[_]` gets cancelled, then my vote will be strongly against it.

Even more important, if the laws in Cats-Effect accommodate to allow for auto-cancelable flatMaps, we now have the problem that code making use of cancelation needs to be described differently. You cannot treat auto-cancelable flatMaps as something optional. You either have them, or you don't. Just to give an example of the design that's already in Cats-Effect, with auto-cancelable flatMaps the `onCancelRaiseError` operation in `Concurrent` becomes effectively useless.

> I'd rather have completeL take a Priority[Concurrent[F], Sync[F]] (from here), so you can make terminal operators create uncancelable things, or just advise against such usage in docs.

Well see, here's the issue. With auto-cancelable flatMaps that no longer works. Our `Async`-powered data types are only non-cancelable because the laws of `Concurrent` say so 😉

This is actually an implicit requirement for `Async` even if `Concurrent` never existed. Consider things like `Deferred` or `MVar` and how they work with `Async`, in which case auto-cancellation can actually yield memory leaks, because the user is no longer back-pressuring on their result and cancellation wouldn't clean up after itself. The only way to counteract that is via `bracket`. But now you no longer have versions that don't need to care about cancelation.

Auto-cancellation basically makes everything more complicated.

> Auto-cancelable run loops are just quite a cool feature to have, and deserve a better treatment 😃

I agree, which is why I fixed the `bracket` operation.

If Scalaz 8 will ever have type classes describing their flavor of interruptible tasks, `Task` will be able to implement them.

alexandru:
Renamed to Proposal, as I want us to sleep on it. 

Maybe I should have fixed only `bracket` in this PR 😆

alexandru:
Master is diverging, I'm including some improvements in other PRs.

Plus a decision is on hold.

Closing for now.

PR #674, opened 2018-05-16
Fix Task.bracket for auto-cancelable run-loops

alexandru:
This test fails currently:

```scala
    var effect = 0
    val task = Task(1)
      .bracket(_ => Task.sleep(1.second))(_ => Task(effect += 1))
      .autoCancelable

    val f = task.runAsync
    sc.tick()
```

PR is fixing it for auto-cancelable loops.

Also renamed `Task#cancelable` to `Task#autoCancelable`, to differentiate it from the `Task.cancelable` builder.


alexandru:
Superseded by #679.

PR #677, opened 2018-05-21
Proposal: Add Starting Point section to CONTRIBUTING.md

Avasil:
What do you think about updating CONTRIBUTING section to better emphasize the fact that contributors are welcome? I think it's a small thing but might actually help attracting new people.

Just an idea in this PR any suggestions are welcome. :)


alexandru:
Great idea 👍

PR #678, opened 2018-05-21
WIP: introduce Scope to Iterant encoding

### LINKED ISSUE #676: Iterant: investigate alternative encoding
Suggested at [May 21, 2018 11:04 AM](https://gitter.im/typelevel/cats-effect?at=5b027da81cfca775e125e363) by @oleg-py ...

Instead of the `earlyStop` mechanism for `Iterant`, we need to investigate if this is feasible:

```scala
case class Bracket[F[_], A](acquire: F[A], use: A => Iterant[F, B], release: A => F[Unit])
  extends Iterant[F, B]
```

And the nodes themselves will no longer have a `stop: F[Unit]` reference:

```scala
case class Next[F[_], A](value: A, rest: F[Iterant[F, A]]) 
  extends Iterant[F, A]
```

Open questions:

1. does it work with auto-cancelable binds?
  - need to investigate the implementation of folds, e.g. `completeL`
2. does it work for the current operations, the most important ones being `flatMap` and `concat` (e.g. what happens with memory safety?)

This issue is related to https://github.com/typelevel/cats-effect/issues/192

oleg-py:
Fixed #676

Pass 1: monix-tail subproject compiles with new encoding

### Good stuff:
* all the stopRef machinery and a significant chunk of error handling are gone, and writing folds is nice and easy
* foldRight is now resource-safe, since `Scope`, like `Suspend` before, is completely free of any significant data.
* With function-less encoding, stuff like `liftMap` can be translated easily.

### Bad stuff:
* Can't have concat and flatMap with that encoding don't keep resources for longer than necessary. Like all problems in this world, this one might be solved by introducing a primitive: `Concat(lhs: Iterant[F, A], rhs: F[Iterant[F, A]])`
* Probably would need an `uninterruptible` combinator, like the one I mentioned [https://github.com/typelevel/cats-effect/issues/192#issuecomment-389766143](here)
* Yet to figure out how to handle errors from `close`


alexandru:
This is significant effort @oleg-py 👍

Random idea — what if you introduce an `after: F[Iterant[F, A]]` in that `Scope`?

```scala
case class Scope[F[_], A]( 
  open: F[Unit],
  rest: F[Iterant[F, A]],
  close: ExitCase[Throwable] => F[Unit],
  after: F[Iterant[F, A]])
```

This would be effectively the `Concat` that you're proposing, except that we only need `Concat` when dealing with `Scope`, so you know, in case `Concat` isn't useful for anything else, might as well unify them.

Random question — I don't understand how this encoding works for `Iterant.resource`:

```scala
def resource[F[_], A](acquire: F[A])
    (release: A => F[Unit])
    (implicit F: Sync[F]): Iterant[F, A]
```

How do you describe this via `Scope`?

alexandru:
Other questions — why would it need an `interruptible` combinator and how would that be any different from `uncancelable`?

oleg-py:
@alexandru I've found a simpler solution to concat problem: we can always `close` manually, if close is idempotent. See [here](https://github.com/monix/monix/pull/678/commits/77a13e9cdc54b369b1468f0e8b0752fdf3623bfd#diff-2a1da9fc5d41ca2c178ea44ecfc5cb07R117) for change. It's a bit unnatural, but seems to work exactly like I wanted in simple.

Given that with `Ref.unsafe` we can implement `Task.memoize` for any `F[_]: Sync`, I think that's fine.

With `Concat` as a primitive, or merged into `Scope`, loops that have any state (e.g. `drop(n)`) become problematic, as we pass the current state to the left branch, but also need to pass the *updated* state into the right branch, which we currently have no way of getting back.

If there are no objections, I'll proceed with fixing tests and see where that would get us.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantConcat.scala
```diff
@@ -106,31 +107,34 @@ private[tail] object IterantConcat {
...
     */
-  def concat[F[_], A](lhs: Iterant[F, A], rhs: Iterant[F, A])
+  def concat[F[_], A](lhs: Iterant[F, A], rhs: Iterant[F, A], close: ExitCase[Throwable] => F[Unit])
     (implicit F: Sync[F]): Iterant[F, A] = {
 
     lhs match {
-      case s @ Scope(_, _, _) =>
-        s.runMap(concat(_, rhs))
+      case s @ Scope(_, _, newClose) =>
+        s.runMap(concat(_, rhs, ec => close(ec) >> newClose(ec)))
```
alexandru: Btw, is this stack safe?

Also, I'd rather not have `Scope` the first in the match statement. Normally Scala should generate a  switch for case classes in bytecode, so it doesn't matter much, but the Scala compiler is moody, so I order them by frequency.

I would add `Scope` branches after `Suspend`, but before Last / Halt.

alexandru:
I just had a closer look and to be honest, I don't think the new encoding is going to cut it.

In `concat` you started from this:

```scala
case Scope(acquire, lhs, release) =>
  Scope(acquire, lhs.map(concat(_, rhs)), release)
```

This obviously doesn't work because it delays `release` until _after_ processing both streams.
So you fixed it with this:

```scala
case s @ Scope(_, _, newClose) =>
  Scope(acquire, concat(_, rhs, ec => close(ec) >> newClose(ec)), newClose)
```

But this relies on ensuring idempotency for `close`, kind of increasing complexity. Now in a fold like `completeL` you're doing this:

```scala
case Scope(open, use, close) =>
  F.bracketCase(open)(_ => F.flatMap(use)(f))((_, exitCase) => close(exitCase))
```

Which is fine, takes care of errors thrown and cancelations. Takes care of everything, except for _early stop_.

In the [Iterant.take](https://github.com/monix/monix/blob/v3.0.0-RC1/monix-tail/shared/src/main/scala/monix/tail/internal/IterantTake.scala#L83) we've got a usage sample of early stop, the whole raison d'être for the `stop` reference being placed on each node.

So we have this use case:

- take 100 elements from the stream
- then release any resources

How will you handle that with `Scope`? I don't see how, I don't think it is possible without throwing an exception maybe and that's terrible.

alexandru:
Also, unless I'm missing something, this is less than ideal and is not going to cut it:

```scala
case s @ Scope(open, use, newClose) =>
  Scope(open, use.map(concat(_, rhs, ec => close(ec) >> newClose(ec))), newClose)
```

The problem is that the `newClose` reference should be gone, garbage collected, as soon as the left hand parameter is finished and before the right hand parameter starts being processed by the fold.

By keeping it around until both streams are processed, it's most likely a memory leak.

oleg-py:
> How will you handle that with Scope? I don't see how, I don't think it is possible without throwing an exception maybe and that's terrible.

This use case is solved by `Scope` basically for free and is already working with current changes.

```
              acquire ------------------ release
                 v                         v
Batch(50 elems)  *   Iterant(100 elems)    * 

```

When I do a `take(55)`, I get the following Iterant back, since I transform the inner node:

```
              acquire ------------------ release
                 v                         v
Batch(50 elems)  *   Iterant(5 elems)      * 

```

Had I done something like `take(5)`, the Scope would be dropped together with acqusition and release logic.

And if I do an early-stoppable fold like `exists`, the resulting `F[Boolean]` is processed like follows:

* Inspect 50 elems
* Bracket: acquire
* Inspect 1 elem, complete in F of `true`
* Bracket: release

---

> The problem is that the newClose reference should be gone, garbage collected, as soon as the left hand parameter is finished and before the right hand parameter starts being processed by the fold.

That's a good catch. I think this (and idempotency concern) can be solved with a dirty optimization trick like we did with ObjectRef and folds, by basically having a mutable queue of handlers modified as needed while list is traversed. With that and `uncancelable` operator, we can remove inner `Scope` nodes on `concat`.

More natural and simpler encoding would be to move away from linked list model and have `Concat(..., ...)` state as I proposed above or have that concern in `Scope`. The problem with it is that all transforms (like `take`) require rewrite under that, since processing two branches is not monadic tail-recursion anymore, a state often needs to be shared.

monix-tail/shared/src/main/scala/monix/tail/internal/IterantConcat.scala
oleg-py: Sorry for going off-topic, but how can Scala generate a switch for case classes? 😮 I mean, there's nothing to switch on.

alexandru:
Looking at the implementation of `onErrorHandleWith`, which judging by the TODOs I guess you've seen the issues:

```scala
        // TODO: open uninterruptibly
        // TODO: lift errors from close
        case Scope(open, rest, close) => Suspend {
          open.attempt.map {
            case Right(_) => Scope(F.unit, rest.map(loop), close)
            case Left(ex) => f(ex)
          }
        }
```

The only solution I can think of is to store state into a shared ref, something like this:

```scala
val open2 = open.handleErrorWith(e => ref.set(e) *> F.raiseError(e))
val close2 = close.handleErrorWith(e => ref.set(e) *> F.raiseError(e))

Suspend(Ref[F].of(null : Throwable).map { ref =>
  Scope(open2, use, close2) ++ ref.get.map {
    case null => Iterant.empty
    case error => recover(error)
  }
})
```

It's kind of horrible 😐 Well, TBH the current `onErrorHandleWith` isn't any better.

alexandru:
Closing in favor of #683

PR #680, opened 2018-05-24
WIP / Proposal: Maybe drop monix-tail and Iterant

alexandru:
**WARN:** the decision to merge this has not been made yet. At this point this is only speculation of one possible outcome — the purpose of this PR being to put the possibility on the table.

The TL;DR is that Cats-Effect's cancelation model is under question due to the need to relax its laws to support Scalaz 8's IO. References:

1. [Restate ConcurrentLaws to support auto-cancelable run loops](https://github.com/typelevel/cats-effect/pull/237), as a solution to [Cancelation laws presuppose no interruption](https://github.com/typelevel/cats-effect/issues/192)
2. [Technical Summary of the Problem](https://gist.github.com/alexandru/b970186208d1c1378c903d43588bd6a7)
3. [Proposal: Continual (type class?)](https://github.com/typelevel/cats-effect/issues/242)

Therefore, without consensus reached in Cats-Effect with the other collaborators, dropping `monix-tail` might happen in case these issues in Cats-Effect are not solved via a solution that admits the current behavior or an easy update, because in the proposed model `Iterant`'s encoding is broken.

Another possibility would be introducing the new encoding in progress at #678, however this would take at least one month by my estimates, if I and other Monix maintainers were to focus on this. And speaking for myself, I have a full time job, a family, upcoming exams, summer vacation coming up, etc. so 1 month to change that encoding is extremely optimistic, because it's a redo and I've got more important things to do at this point. It isn't fair for current users of `monix-eval` and `monix-reactive` to wait for so long on a final version 3.0.

Other possibilities would be:

1. delay Monix 3.0 indefinitely
2. release 3.0 soon, then reintroduce `monix-tail` at a later point after we find a better encoding (which implies the merging of this PR)
3. specialize `Iterant` for `Task`
4. fork the Cats-Effect type classes in a Monix sub-project, because the purpose of `Iterant` was to work with `Coeval` as well

Actually at this point there are no good solutions, unless consensus is reached for a solution that allows the current model to work.


hepin1989:
how about move the monix-tail and monix-itrent to a sub repos?

hepin1989:
I can't see what can't be achieved with the reactive-stream based one, so drop it would be a release for both the project and our daily life:)

PR #683, opened 2018-06-16
Iterant, version 2

alexandru:
This is a collaboration between myself and @oleg-py and @Avasil for redefining `Iterant` in a way that is compatible with https://github.com/typelevel/cats-effect/pull/237.

The feature is now feature complete and ready to be merged 🎊

TL;DR — this is a huge PR that modifies `Iterant`'s encoding to be compatible with Cats-Effect's exposed cancellation model. 

For example if you specify resources such as this ...

```scala
val resource = Iterant[Task].resource {
  Task(new BufferedReader(new FileReader(file)))
} { in =>
  Task(in.close())
}

// ...
resource.flatMap { in => Iterant.repeatEvalF(Task(in.readLine)) }
  .takeWhile(_ != null)
  .foldLeftL("")(_ + _)
```

Internally the `Iterant` now deals with resource handling via [Bracket](https://typelevel.org/cats-effect/typeclasses/bracket.html) and so cancellation of resulting tasks (via `Task`, `IO`, etc) is safe and reusable with any effect type that supports the Cats-Effect type classes.

`Iterant` is a data structure, an [ADT](https://en.wikipedia.org/wiki/Algebraic_data_type) with this definition:

```scala
  sealed abstract class Iterant[F[_], A]

  case class Next[F[_], A](item: A, rest: F[Iterant[F, A]])
    extends Iterant[F, A]

  case class Last[F[_], A](item: A)
    extends Iterant[F, A]

  case class NextCursor[F[_], A](cursor: BatchCursor[A], rest: F[Iterant[F, A]])
    extends Iterant[F, A]

  case class NextBatch[F[_], A](batch: Batch[A], rest: F[Iterant[F, A]])
    extends Iterant[F, A]

  case class Suspend[F[_], A](rest: F[Iterant[F, A]])
    extends Iterant[F, A]

  case class Halt[F[_], A](e: Option[Throwable])
    extends Iterant[F, A]

  case class Scope[F[_], A, B](
    acquire: F[A],
    use: A => F[Iterant[F, B]],
    release: (A, ExitCase[Throwable]) => F[Unit])
    extends Iterant[F, B]

  case class Concat[F[_], A](lh: F[Iterant[F, A]], rh: F[Iterant[F, A]])
    extends Iterant[F, A] 
```

What's new with this encoding:

1. the concept of `earlyStop` is gone from the nodes, being replaced by `Scope`
2. `Scope` is a new data constructor that mirrors `bracket`
3. `Concat` is a new data constructor that's meant for doing concatenation (`++`), needed due to the introduction of `Scope`

In order to keep the sub-project honest, `monix-tail` (exposing `Iterant`) no longer depends on `monix-eval` (exposing `Task`). This prevents us to take Task-related shortcuts, plus the `monix-tail` sub-project is now lighter and only depends on `monix-execution` (for low-level JVM utilities).

*Performance* has improved across the board due to internal usage of the "Visitor pattern", for eliminating memory allocations when interpreting the Interant data structure and due to introducing the `Concat` state. In many cases the improvement is ten fold.


volth:
Could you release ```3.0.0-RC2``` including these changes?

alexandru:
@volth I'm back from vacation and working on it.

Issue #688, opened 2018-07-20
Task.fromTry type constructor is being eagerly executed


jupco:
I have observed a weird behavior in the `Task.fromTry` function, when I had a  `Try` block which contained a long-time blocking call to a Java library.
What I could see was that my thread was blocked and `timeout` had no effect on this; while reviewing the code I found that this is the contract of `fromTry`
```scala
def fromTry[A](a: Try[A]): Task[A]
```
it receives an evaluated Try so no timeout can be applied because it's already a `Now` or an `Error` but it takes all the time needed to resolve the `Try`

alexandru:
`Task.fromTry` being eager is by design. Like many such functions being described in Monix, or in Typelevel Cats in general, strictness is the default and not laziness.

If you want lazy evaluation, you have to mark it as being lazy via one of Task's operators that deal with *suspending side effects*:

- `Task.eval` (also aliased to `delay` from [Cats-Effect](http://typelevel.org/cats-effect/))
- `Task.suspend`

If you're not doing that then the computation is considered *pure* and thus can be eagerly evaluated. Or in other words `Task.fromTry` expects a pure value. So the normal workflow is actually something like this:

```scala
Task.suspend {
  Task.fromTry(Try { 
    f() 
  })
}
```

You don't even need to do that of course, as you can just leave `Task` to catch the exception. But OK, in case you've got some function around that returns `Try`, you can just `suspend` that `fromTry`.

alexandru:
Closing as per #689.

PR #689, opened 2018-07-20
Fix/change task.fromTry to lazy parameter

jupco:
This PR addresses https://github.com/monix/monix/issues/688

Changes the contract of `fromTry` to receive the try as a lazy parameter and wraps its evaluation on a Task so it can be asynchronously executed


alexandru:
Hi @jupco, appreciate trying your hand at a commit 🙂

Can't approve changing fromTry's parameter to a by-name param though, I wrote on that issue as to why: https://github.com/monix/monix/issues/688#issuecomment-406955703

alexandru:
Thanks for PR. Will now close it. Hopefully you'll think of some other improvements in the future too.

Issue #695, opened 2018-08-05
Unwanted side-effect in FutureUtils.timeoutTo


LMnet:
When using `FutureUtils.timeoutTo` there is a possibility of fallback side-effect evaluation even when result is not fallback. It's easier to understand that from example:

```scala
def doSmth(): Future[Int] = ???
def fallback = {
  sideEffectfulFunction()
  Future.failed(new RuntimeException)
}

val result = doSmth.timeoutTo(1.second, fallback)
```

If `doSmth` and `fallback` will finish in almost same time, sometimes it leads to a situation when `sideEffectfulFunction` is evaluated, but `result` will contain result from `doSmth`. This situation is pretty rare, but I wrote a test that can simulate such a situation a lot of times and reproduce a bug.

An interesting moment about this issue: `Task` from 2.3.3 is also subject to this, but in 3.0.0-RC1 I can't reproduce this issue with `Task`.

LMnet:
This is fixed with #696

PR #696, opened 2018-08-05
Fix unwanted side effect in FutureUtils.timeoutTo (#695)

LMnet:
Fix for #695

To eliminate the possibility of race condition I moved `fallback` evaluation after promise resolving. It will lead to some overhead (extra `Either` allocation and `flatMap` on future), but I didn't find any better way to guarantee to fix the problem.

I added 2 test cases. First is pretty simple - just check that side-effect is not evaluated when future finished earlier than timeout. I placed it in the shared tests. 
The second one is a lot more interesting. I placed it on the JVM tests because this issue is reproducible only in environments with the real multithreading. This test runs on a real thread pool and tries to adjust timeout to find race condition. And when this race condition is found, it checks that the number of errors and side-effect evaluations is the same. This test is non-deterministic and takes some time to find race condition, but with this test, I was able to reproduce issue from #695 before this fix.

### DIFF
--- a/monix-execution/shared/src/main/scala_2.12/monix/execution/FutureUtils.scala
+++ b/monix-execution/shared/src/main/scala_2.12/monix/execution/FutureUtils.scala
@@ -66,18 +66,24 @@ object FutureUtils {
   def timeoutTo[A](source: Future[A], atMost: FiniteDuration, fallback: => Future[A])
     (implicit s: Scheduler): Future[A] = {
 
-    val promise = Promise[A]()
+    val promise = Promise[Either[Unit, Try[A]]]()
     val task = s.scheduleOnce(atMost.length, atMost.unit,
-      new Runnable { def run() = promise.tryCompleteWith(fallback) })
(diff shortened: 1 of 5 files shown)


monix-execution/shared/src/main/scala_2.12/monix/execution/FutureUtils.scala
```diff
@@ -66,18 +66,24 @@ object FutureUtils {
   def timeoutTo[A](source: Future[A], atMost: FiniteDuration, fallback: => Future[A])
     (implicit s: Scheduler): Future[A] = {
 
-    val promise = Promise[A]()
+    val promise = Promise[Either[Unit, Try[A]]]()
```
oleg-py: `Either[Unit, X]` is isomorphic to `Option[X]` for any `X`, and that would save allocations on `Left(())` vs `None`

LMnet: :man_facepalming: 
Fixed

alexandru:
No, looks good.

PR #698, opened 2018-08-21
[WIP] Upgrade to Cats-Effect 1.0.0-RC3

Avasil:
My attempt at upgrading Cats-Effect version.

I hacked whatever I could to make it compile, mostly converting between `IO` and `F`. I won't be surprised if there are more direct ways but I am yet to become more comfortable with cancelation related internals so I didn't think about it too much.

One thing I don't like is that having `Scheduler` in scope doesn't give us  `ContextShift` for types other than `Task` (it could be done with any `ExecutionContext` + `Effect` but probably in Cats-Effect to avoid extra imports). It also means we don't have `Concurrent[IO]` unless we provide `ContextShift` on our own.

`evalOn` is implemented on top of bracket - do we want something else?

Things that don't work right now:

- `monix.eval.TaskConversionsSuite` fails this test:
```scala
test("Task.fromEffect(task.to[IO]) preserves cancelability") { implicit s =>
  check1 { (task: Task[Int]) =>
    val lh = Task.fromEffect(task.to[IO]).start.flatMap(f => f.cancel *> f.join)
    lh <-> task.start.flatMap(f => f.cancel *> f.join)
  }
}
```
I don't see why looking at the code, I'll have to play with it unless this is due to https://github.com/typelevel/cats-effect/pull/311 and the test itself need fixing ?

- `monix.tail.TypeClassLawsForIterantIOSuite` sometimes fails with:

```
[info] - MonadError[Iterant[IO]].monadError.semigroupal associativity *** FAILED ***
[info]   Exception raised on property evaluation. (Checkers.scala:39)
[info]   > ARG_0: Suspend(IO$1672045507)
[info]   > ARG_1: Suspend(IO$1743155629)
[info]   > ARG_2: Suspend(IO$1080941687)
[info]   > Exception: java.lang.StackOverflowError: null
[info]     minitest.api.Asserts.fail(Asserts.scala:103)
[info]     minitest.api.Asserts.fail$(Asserts.scala:102)
[info]     minitest.api.Asserts$.fail(Asserts.scala:106)
[info]     minitest.laws.Checkers.check(Checkers.scala:39)
[info]     minitest.laws.Checkers.check$(Checkers.scala:36)
[info]     monix.tail.TypeClassLawsForIterantIOSuite$.check(TypeClassLawsForIterantIOSuite.scala:27)
[info]     monix.tail.BaseLawsSuite.$anonfun$checkAllAsync$4(BaseLawsSuite.scala:73)
[info]     monix.eval.TestUtils.liftedTree1$1(TestUtils.scala:39)
[info]     monix.eval.TestUtils.silenceSystemErr(TestUtils.scala:38)
[info]     monix.eval.TestUtils.silenceSystemErr$(TestUtils.scala:32)
[info]     monix.tail.TypeClassLawsForIterantIOSuite$.silenceSystemErr(TypeClassLawsForIterantIOSuite.scala:27)
[info]     monix.tail.BaseLawsSuite.$anonfun$checkAllAsync$3(BaseLawsSuite.scala:73)
[info]     minitest.SimpleTestSuite.$anonfun$test$1(SimpleTestSuite.scala:27)
[info]     minitest.api.TestSpec$.$anonfun$sync$1(TestSpec.scala:51)
[info]     minitest.api.TestSpec.apply(TestSpec.scala:27)
[info]     minitest.api.Properties.$anonfun$iterator$2(Properties.scala:38)
[info]     minitest.api.TestSpec.apply(TestSpec.scala:27)
[info]     minitest.runner.Task.loop$1(Task.scala:40)
[info]     minitest.runner.Task.$anonfun$execute$1(Task.scala:47)
[info]     scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:303)
[info]     ...
```

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/internal/TaskConversions.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/internal/TaskConversions.scala
@@ -59,7 +59,7 @@ private[eval] object TaskConversions {
     def cancelable(f: CancelableFuture[A])(implicit F: Concurrent[F]): F[A] =
       F.cancelable { cb =>
         f.underlying.onComplete(AttemptCallback.toTry(cb))(immediate)
-        f.cancelable.cancelIO
+        F.liftIO(f.cancelable.cancelIO)
       }
 
     source match {
@@ -88,8 +88,8 @@ private[eval] object TaskConversions {
   private def fromAsync0[F[_], A](fa: F[A])(implicit F: Effect[F]): Task[A] = {
     val start = (ctx: Context, cb: Callback[A]) => {
       try {
-        val io = F.runAsync(fa)(new CreateCallback[A](null, cb)(ctx.scheduler))
-        io.unsafeRunAsync(AttemptCallback.noop)
+        val syncIO = F.runAsync(fa)(new CreateCallback[A](null, cb)(ctx.scheduler))
+        syncIO.toIO.unsafeRunAsync(AttemptCallback.noop)
       } catch {
         case e if NonFatal(e) =>
           ctx.scheduler.reportFailure(e)
@@ -106,8 +106,8 @@ private[eval] object TaskConversions {
         val cancelable = SingleAssignCancelable()
         conn push cancelable
 
-        val io = F.runCancelable(fa)(new CreateCallback[A](conn, cb))
-        cancelable := Cancelable.fromIOUnsafe(io.unsafeRunSync())
+        val syncIO = F.runCancelable(fa)(new CreateCallback[A](conn, cb))
+        cancelable := Cancelable.fromIO(F.toIO(syncIO.unsafeRunSync()))
       } catch {
         case e if NonFatal(e) =>
           ctx.scheduler.reportFailure(e)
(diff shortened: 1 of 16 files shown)


monix-eval/shared/src/main/scala/monix/eval/internal/TaskConversions.scala
```diff
@@ -106,8 +106,8 @@ private[eval] object TaskConversions {
         val cancelable = SingleAssignCancelable()
         conn push cancelable
 
-        val io = F.runCancelable(fa)(new CreateCallback[A](conn, cb))
-        cancelable := Cancelable.fromIOUnsafe(io.unsafeRunSync())
+        val syncIO = F.runCancelable(fa)(new CreateCallback[A](conn, cb))
+        cancelable := Cancelable.fromIO(F.toIO(syncIO.unsafeRunSync()))
```
Avasil: With this change `Cancelable.fromIOUnsafe` is unused but maybe it should stay here - how can I assess whether I need to guarantee idempotency or not?

alexandru:
Thanks @Avasil. I've taken your commits and I'm integrating them into my "conversions" PR, because it's overlapping. Stay tuned.

alexandru:
So I integrated this PR in #686, which is now merged.

PR #699, opened 2018-08-27
Add Concat and Scope to Iterant description

Avasil:
It seems like we've forgotten to do it during `Iterant` rewrite

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
@@ -69,6 +69,13 @@ import scala.concurrent.duration.{Duration, FiniteDuration}
   *  - [[monix.tail.Iterant.Suspend Suspend]] is for suspending the
   *    evaluation of a stream.
   *
+  *  - [[monix.tail.Iterant.Concat Concat]] represents concatenation
+  *    of two streams.
+  *
+  *  - [[monix.tail.Iterant.Scope Scope]] is to specify the acquisition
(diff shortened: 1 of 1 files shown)


monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -69,6 +69,13 @@ import scala.concurrent.duration.{Duration, FiniteDuration}
   *  - [[monix.tail.Iterant.Suspend Suspend]] is for suspending the
   *    evaluation of a stream.
   *
+  *  - [[monix.tail.Iterant.Concat Concat]] represents concatenation
+  *    of two streams.
+  *
+  *  - [[monix.tail.Iterant.Scope Scope]] is to specify the acquisition
+  *    and release of resources. It is effectively encoding of
```
alexandru: "the encoding"

monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -69,6 +69,13 @@ import scala.concurrent.duration.{Duration, FiniteDuration}
   *  - [[monix.tail.Iterant.Suspend Suspend]] is for suspending the
   *    evaluation of a stream.
   *
+  *  - [[monix.tail.Iterant.Concat Concat]] represents concatenation
```
alexandru: "the concatenation"

Avasil:
Thanks @alexandru

PR #704, opened 2018-09-03
Introduce Observable.doOnStartTask and Observable.doOnStartEval

### LINKED ISSUE #697: Observable.doOnStartEval
`Observable` has a lot of `doOnX` functions, usually with async variant taking `Task` or other `Effect`. 

It's not the case for all the operators, e.g. [`doOnStart`](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala#L1143)

Would be cool to have

```scala
final def doOnStartEval[F[_]](cb: A => F[Unit])(implicit F: Effect[F]): Observable[A]
final def doOnStartTask(cb: A => Task[Unit]): Observable[A]
```
Anyone looking at implementing this operator could look at differences between similar functions like [EvalOnNext](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnNextAckOperator.scala) and [DoOnNext](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DoOnNextAckOperator.scala)

ilya-murzinov:
Resolves #697


monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnStartOperator.scala
```diff
@@ -0,0 +1,71 @@
...
+      private[this] var isStart = true
+
+      def onNext(elem: A): Future[Ack] = {
+        // Protects calls to user code from within the operator and
+        // stream the error downstream if it happens, but if the
+        // error happens because of calls to `onNext` or other
+        // protocol calls, then the behavior should be undefined.
+        try {
+          if (isStart) {
+            cb(elem).onErrorHandle(onError).runAsync
```
ilya-murzinov: Not sure about this 🤔

ilya-murzinov: I realised that this is incorrect, will fix shortly

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnStartOperator.scala
```diff
@@ -0,0 +1,76 @@
...
+
+      def onNext(elem: A): Future[Ack] = {
+        // Protects calls to user code from within the operator and
+        // stream the error downstream if it happens, but if the
+        // error happens because of calls to `onNext` or other
+        // protocol calls, then the behavior should be undefined.
+        try {
+          if (isStart) {
+            val t = for {
+              _ <- cb(elem).onErrorHandle(ex => { onError(ex); Stop })
```
Avasil: I'm not sure about this. @alexandru please correct me if I'm wrong.

Here we call `out.OnError` in case of error, return `Stop` but ignore it and still call `out.OnNext` which violates `Observer` contract (you shouldn't send anything after `onError`)

In my opinion we should short citcuit here in case of error. Ideally, we could write failing test now and then make it work after fix. Something similar to what you already have with hand written `Subscriber` but keeping track of `onNext` / `onError` calls

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -1145,6 +1145,12 @@ abstract class Observable[+A] extends Serializable { self =>
   final def doOnStart(cb: A => Unit): Observable[A] =
     self.liftByOperator(new DoOnStartOperator[A](cb))
 
+  final def doOnStartTask(cb: A => Task[Unit]): Observable[A] =
```
Avasil: Could you add scaladocs?
Similar to `doOnStart` but mentioning the difference and other variants (like `doOnNextAck` and others )

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnStartOperator.scala
ilya-murzinov: Thanks @Avasil I've noticed this bug and was thinking about how to fix it after I pushed yesterday :D
And short-circuting in case of error makes perfect sense, I just couldn't figure out how to implement it in a best way.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
ilya-murzinov: sure, will do

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnStartOperator.scala
```diff
@@ -0,0 +1,78 @@
...
+      private[this] var isDone = false
+      private[this] var isStart = true
+
+      def onNext(elem: A): Future[Ack] = {
+        if (isStart) {
+          // Protects calls to user code from within the operator and
+          // stream the error downstream if it happens, but if the
+          // error happens because of calls to `onNext` or other
+          // protocol calls, then the behavior should be undefined.
+          def tryExecute(): Task[Option[Ack]] = try {
```
ilya-murzinov: @Avasil idk how elegant this looks to you, please feel free to suggest improvements!

Avasil: I think we don't need to use `Option` at all. 

We could approach it like in this example (I've realized that `tryExecute()` is unused there, already fixed that with PR):
https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnNextAckOperator.scala#L52

Something like this:
```scala
cb(elem)
  .flatMap(_ => Task.fromFuture(out.onNext(elem))) // success
  .onErrorHandle(ex => { onError(ex); Stop) } ) // failure
```

And also `cb(elem)` in `try-catch` to protect against functions like `_ => throw ex`

ilya-murzinov: I tried this one as well, but I didn't like the fact that `onErrorHandle` also catches errors from `onNext`, do you think this is fine?

Avasil: Oh, you're right
Maybe we could match on `attempt`?

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnStartOperator.scala
```diff
@@ -0,0 +1,74 @@
...
+          // protocol calls, then the behavior should be undefined.
+          val t = try {
+            cb(elem)
+          } catch {
+            case NonFatal(ex) => Task.raiseError(ex)
+          }
+
+          isStart = false
+
+          t.attempt.flatMap {
```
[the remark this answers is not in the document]
ilya-murzinov: btw thanks for the hint!

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnStartOperator.scala
```diff
@@ -0,0 +1,74 @@
...
+            case Left(ex) => Task.eval { onError(ex); Stop }
+            case Right(()) => Task.deferFuture(out.onNext(elem))
+          }.runAsync
```
Avasil: We could use small optimization like [here](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnNextAckOperator.scala#L50) - it will be synchronous in case of `Stop` and sometimes `onNext` is synchronous too and just returning `Ack` at the end

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnStartOperator.scala
```diff
@@ -0,0 +1,74 @@
...
+            cb(elem)
+          } catch {
+            case NonFatal(ex) => Task.raiseError(ex)
+          }
+
+          isStart = false
+
+          t.attempt.flatMap {
+            case Left(ex) => Task.eval { onError(ex); Stop }
+            case Right(()) => Task.deferFuture(out.onNext(elem))
```
Avasil: We run it right away and need `Task` only for types so IMO we could use `Task.fromFuture`

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnStartOperator.scala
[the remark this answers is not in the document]
Avasil: Cool @ilya-murzinov ! 
I've added two more comments with nit picks but overall it looks great. :) Thanks a lot!

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnStartOperator.scala
ilya-murzinov: yep, makes sense, I just thought that you can't be "too lazy" in this case :D

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/EvalOnStartOperator.scala
ilya-murzinov: yep, thanks for anther hint!

ilya-murzinov: tbh, I don't completely understand how it works...
```
ack.value match {
  case Some(Success(sync)) => sync
  case _ => ack
}
```
I mean, `sync` is `Ack` and `ack` is `Future[Ack]`, so how do types align here? 🤔

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -1145,6 +1146,35 @@ abstract class Observable[+A] extends Serializable { self =>
...
 
+  /** Executes the given callback only for the first element generated
+    * by the source Observable, useful for doing a piece of
+    * computation only when the stream starts.
+    *
+    * @return a new Observable that executes the specified callback
+    *         only for the first element
+    * @see [[doOnStart]] for a simpler version that doesn't allow
+    *     asynchronous execution, or
+    *     [[doOnStartEval]] for a version that can do ev
```
Avasil: Oh, and sorry for the nitpicks but something went wrong with the formatting here (looks better on `doOnStartEval` version)

ilya-murzinov: yeah, totally, my bad!

PR #705, opened 2018-09-04
fix all Scaladocs (except Task's) in monix.eval

jozic:
this fixes almost all Scaladocs in `monix.eval` package
- `Task` excluded for now, as it is the biggest source of compile errors, just needs more time
- `TaskApp` is also excluded (though the code is compilable now) as it gives this strange warning (promoted to an error under `-X-fatal-warnings`)
```
Error:(40, 12) MyApp$2 has a main method with parameter type Array[String], but monix.eval.TaskAppDoctest.MyApp$2 will not be a runnable program.
  Reason: companion contains its own main method, which means no static forwarder can be generated.
    object MyApp extends TaskApp {
```
I think this is related to this object defined inside test method in generated test, not sure how to fix it
- also the Scaladoc in `TaskLike` lead me to removal of execution context parameter from `fromFuture`

so now instead of 
```scala
implicit def fromFuture(implicit ec: ExecutionContext): TaskLike[Future]
```
it's

```scala
implicit val fromFuture: TaskLike[Future]
```

looks like `ec` wasn't needed, but i'm not sure there wasn't some planned usage for it, lmk about this one

@oleg-py @alexandru please review

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/TaskCircuitBreaker.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/TaskCircuitBreaker.scala
@@ -109,7 +109,8 @@ import scala.util.{Failure, Success}
   *       throw new RuntimeException("dummy")
   *   }
   *
-  *   val task = circuitBreaker.protect(problematic)
+  *   val task = circuitBreaker
+  *     .flatMap(_.protect(problematic))
   * }}}
   *
   * When attempting to close the circuit breaker and resume normal
@@ -117,7 +118,7 @@ import scala.util.{Failure, Success}
   * failed attempts, like so:
   *
   * {{{
-  *   val circuitBreaker = TaskCircuitBreaker(
+  *   val exponential = TaskCircuitBreaker(
   *     maxFailures = 5,
   *     resetTimeout = 10.seconds,
   *     exponentialBackoffFactor = 2,
@@ -535,9 +536,9 @@ object TaskCircuitBreaker {
       * when the `Open` state is to transition to [[HalfOpen]].
       *
       * It is calculated as:
-      * {{{
+      * ```scala
       *   startedAt + resetTimeout.toMillis
-      * }}}
+      * ```
       */
     val expiresAt: Timestamp = startedAt + resetTimeout.toMillis
   }
--- a/monix-eval/shared/src/main/scala/monix/eval/TaskLike.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/TaskLike.scala
@@ -20,7 +20,7 @@ package monix.eval
 import cats.Eval
 import cats.effect.{ConcurrentEffect, Effect, IO, SyncIO}
 import scala.annotation.implicitNotFound
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.Future
 import scala.util.Try
 
 /** A lawless type class that provides conversions to [[Task]].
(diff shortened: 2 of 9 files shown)


Avasil:
> looks like ec wasn't needed, but i'm not sure there wasn't some planned usage for it, lmk about this one

That would be correct, it wasn't necessary for `deferFuture`.

Build fails with:
```
[0m[[0m[31merror[0m] [0m[0m/home/travis/build/monix/monix/monix-eval/shared/src/main/scala/monix/eval/TaskApp.scala:49:38: Variable name undefined in comment for trait TaskApp in trait TaskApp[0m
[0m[[0m[31merror[0m] [0m[0m  *           Task(println(s"Hello, ${name}.")).as(ExitCode.Success)[0m
[0m[[0m[31merror[0m] [0m[0m                                     ^[0m
[0m[[0m[31merror[0m] [0m[0mone error found[0m
[0m[[0m[31merror[0m] [0m[0m(Scalaunidoc / [31mdoc[0m) Scaladoc generation failed[0m
[0m[[0m[31merror[0m] [0m[0mTotal time: 46 s, completed Sep 4, 2018 4:16:43 AM[0m
```

alexandru:
Ah, my mistake. Yes, `ExecutionContext` isn't needed there, we'll have to fix it.

alexandru:
@Avasil btw that error is from the `$` character having special meaning in ScalaDoc. It has to be escaped, I don't remember how. Maybe by doubling it.

monix-eval/shared/src/main/scala/monix/eval/TaskLike.scala
```diff
@@ -79,7 +79,7 @@ object TaskLike extends TaskLikeImplicits0 {
   /**
     * Converts to `Task` from [[scala.concurrent.Future]].
     */
-  implicit def fromFuture(implicit ec: ExecutionContext): TaskLike[Future] =
+  implicit val fromFuture: TaskLike[Future] =
```
Avasil: Could you also fix `Observable`?

jozic: fixed in `ObservableLike`

jozic:
@alexandru can we merge this? or some changes required?

monix-eval/shared/src/main/scala/monix/eval/TaskCircuitBreaker.scala
```diff
@@ -535,9 +536,9 @@ object TaskCircuitBreaker {
       * when the `Open` state is to transition to [[HalfOpen]].
       *
       * It is calculated as:
-      * {{{
+      * ```scala
```
alexandru: Does this Github-flavored markdown annotation work?

alexandru:
@jozic we can merge it of course. Thanks a lot.

monix-eval/shared/src/main/scala/monix/eval/TaskCircuitBreaker.scala
jozic: sbt-doctest ignores it, if that's what you ask
i did that as the code refers to fields, and is not compilable as is

To me this particular scaladoc example is not really an example, but implementation explanation
to make it compilable we will have to write smth like
```scala
  val open = Open(1234, 5 seconds)
  require(open.startedAt + open.resetTimeout.toMillis == open.expiresAt)
```
but not sure if it's helpful or any better
wdyt?

jozic:
cool, thanks, will focus on fixing `Task`'s docs, that is a big one :)

monix-eval/shared/src/main/scala/monix/eval/TaskCircuitBreaker.scala
alexandru: It’s indeed not a sample. But we have to have ScalaDoc compile it without producing garbage and I don’t think this does.

One thing we can do is to do that inline via single back-ticks.

jozic: you mean 
```
|`startedAt + resetTimeout.toMillis`

instead of 

|```scala
|startedAt + resetTimeout.toMillis
|```
```
?

alexandru: Yes, that can work.

jozic: ok, will address that in PR for Task

Issue #703, opened 2018-09-03
Missing `SemigroupK[Task]` instance


jendakol:
Hello, is there some reason why there is `SemigroupK[Task]` instance **not** present in `Task` companion object (like there are instances for some other Cats typeclasses)? If not, please add it (or I can do it via PR).  
I tried to implement it and simple copy-paste from `cats.affect.IO` works:
```scala
implicit val taskSemigroupK: SemigroupK[Task] = new SemigroupK[Task] {
  def combineK[A](a: Task[A], b: Task[A]): Task[A] =
    ApplicativeError[Task, Throwable].handleErrorWith(a)(_ => b)
}
```

Avasil:
I don't think there is any specific reason, PR will be appreciated @jendakol :)

oleg-py:
I'm closing it, as we now have an instance 👍

PR #706, opened 2018-09-04
CatsSemigroupKForTask added

jendakol:
Solves issue #703


monix-eval/shared/src/main/scala/monix/eval/instances/CatsSemigroupKForTask.scala
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package monix.eval.instances
+import cats.SemigroupK
+import monix.eval.Task
+
+class CatsSemigroupKForTask extends SemigroupK[Task] {
+
+  def combineK[A](a: Task[A], b: Task[A]): Task[A] =
+    Task.catsAsync.handleErrorWith(a)(_ => b)
```
Avasil: I think we could use `a.onErrorHandleWith(_ => b)` directly but I'm not sure if it makes difference though

Avasil:
Thanks for PR @jendakol -- looks good!

But could you add test in `TypeClassLawsForTaskSuite` to make sure it is and remains lawful instance? :)

alexandru:
Hey, nice.

1. wouldn't this instance make better sense if integrated in `CatsBaseForTask`? https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/instances/CatsBaseForTask.scala
2. if you do it for `Task`, we need it for `Coeval` as well

alexandru:
Also, what @Avasil said, we need to test the laws, so checkout the `TypeClassLaws*` tests.

alexandru:
@jendakol yes, include the instance in [CatsBaseForTask](https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/instances/CatsBaseForTask.scala) for `Task` and in [CatsSyncForCoeval](https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/instances/CatsSyncForCoeval.scala) for `Coeval`.

jendakol:
Updated. Implemented for both `Task` and `Coeval`, enabled laws tests.

Issue #710, opened 2018-09-06
Plan for the Cats-Effect 1.0.0 Update


alexandru:
We have to release a version compatible with Cats-Effect 1.0.0 and soon.

Currently these are the pending issues:

1. make `Task` back-pressure on the finalizers on cancellation, like we did in Cats-Effect's `IO` — this is not a very easy change, will work on it either tonight or tomorrow night
2. make `Task` auto-cancelable by default (change the default setting); the design of the type classes clearly favors auto-cancelation by default, therefore the current model has to be opt-in; and we also need to remove the `cancelable` method and replace it with `continual`
3. change `MVar` and `TaskSemaphore`, to inherit from those in Cats-Effect, and possibly use those implementation as a short-cut until we can provide more optimal implementations, see discussion at #708

Not sure if I thought of anything.

Avasil:
I can do 3. today. 
How are we going to let users know that `read` semantics are changed?

alexandru:
`MVar.read` becoming atomic is a relaxation not an added constraint, so a change in ScalaDoc is fine.

Avasil:
@alexandru 
So I looked into it and both `MVar` and `TaskSemaphore` are based on their Async `Future` versions so we can't really mix them with Cats-Effect implementations.

It  comes down to whether we implement all missing operators there or change `MVar` / `TaskSemaphore` to call `Cats-Effect` constructors directly. Or get rid of it completely.

Would be cool to have it implemented in a way that both `Async` and `Task/F` versions have the same operations but I'm afraid it will take too much time and it's not worth prolonging the release. Adding new public methods to `Async` variants shouldn't break binary compatibility right? So could be targeted for `3.1.0` (on related note, I'd love to see `Task` version of `AsyncQueue` :P)

alexandru:
I forgot about the Future-enabled versions.

So it is fine if their implementations aren't based on the Future-enable version. But the API needs to be consistent. Not entirely sure how to proceed, but even the Future-enabled versions can be at first be based on `IO` and work from there.

So if you want to help with this, start with one of them and lets work from there.

alexandru:
What I mean is, lets get the API changes in and we can worry about providing good implementations later.

Avasil:
Seems like it's done, what's the plan forward? I'll have a spare time over the weekend so I can help getting closer to the release. :)

alexandru:
The `Task` back-pressuring on finalizers like in IO is not done, but isn't a priority either.

[...]

PR #711, opened 2018-09-06
Make MVar consistent with Cats-Effect impl

Avasil:
See discussion at #710 
<del>
Quick sketch showing how using cats-effect implementation could look like.
Note that we lose `withPadding` constructors.
<del>
It is easy to do it also for `AsyncVar` but it will be essentially using `MVar` and running it every operation. 
<del>
I'd rather stay with `AsyncVar` as is because if someone needs it with `Future` he can't use `MVar` anyway (needs to be `Async` / `Concurrent`). And if someone needs those semantics before we update implementation he can wrap it quite easily.


Avasil:
BTW I'm trying to add atomic reads to `AsyncVar` but I don't make any promises. x)

Avasil:
@alexandru @oleg-py 
I managed to do it, the only difference between Cats-Effect is that it doesn't support nulls because it uses them (I guess for performance reasons?) internally instead of `Option`.

I don't know if there are any benefits over Cats-Effect version apart from padding constructor. Let's decide what we do with `MVar` and I can either remove it or finish implementation (scaladocs and review)

PR #713, opened 2018-09-08
Expand Observable's scaladoc

Avasil:
I added primitive visualization to few `Observable` operators.
Might be useful for users looking to combine `Observable` in specific way.

Let me know what you think or have ideas how to make it nicer


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -495,9 +495,18 @@ abstract class Observable[+A] extends Serializable { self =>
...
+    * observable, by emitting elements combined in pairs.
+    *
+    * It emits an item whenever any of the source Observables emits an
+    * item (so long as each of the source Observables has emitted at
+    * least one item).
+    *
+    * Example:
+    *
+    * stream1: 1 - - 2 - - 3 - 4 - -
+    * stream2: 1 - - 2 - 3 - - - - 4
```
alexandru: @Avasil sorry for not reviewing this earlier.

This is really cool, however I don't think it will get formatted properly when ScalaDoc is generated, because newlines are not taken into account.

Avasil: In this case I'll try generating scaladoc and see if I can make it wiek / close PR otherwise

alexandru: Go for it. You probably need to introduce some sort of `<pre>` block.

Avasil: @alexandru `<pre>` tags were enough and IMO it looks quite nice:

image

The only issue is extra space but it's not too bad

alexandru: 👍 looks good

Issue #714, opened 2018-09-08
Implement Task.timed


Avasil:
@guizmaii has mentioned on Twitter that he's often missing following method:

```scala
final def timed: Task[(Duration, A)]
```

See [Scalaz-ZIO](https://github.com/scalaz/scalaz-zio/blob/master/core/shared/src/main/scala/scalaz/zio/IO.scala#L527)

`Task` has methods like `deferAction`, `async0` and `cancelable0` which allow access to the `Scheduler` (for clock) so it should be very easy to implement.

@guizmaii 
Do you still want to make PR or you don't mind if someone else takes this?

guizmaii:
Hi @Avasil 

I still want to do it but if someone does the job it's ok for me.

I'll be in a train all the day tomorrow, going to Scotland. I'll try to push you a PR. :)

guizmaii:
Hi @Avasil and @alexandru,

I've pushed an initial implementation here https://github.com/monix/monix/pull/715. 
Let me know your thinking about it.

PR #715, opened 2018-09-09
WIP: Implement Task `timed` method

guizmaii:
Fixes https://github.com/monix/monix/issues/714


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1916,6 +1917,34 @@ sealed abstract class Task[+A] extends TaskBinCompat[A] with Serializable {
...
     TaskCancellation.uncancelable(this)
+
+  /**
+    * TODO: doc
+    *
+    * @return
+    */
+  final def timedAttempt: Task[(Duration, Either[Throwable, A])] =
+    Task.deferAction { scheduler =>
+      val now: Task[Long] = Task.eval(scheduler.clockMonotonic(TimeUnit.NANOSECONDS))
```
oleg-py: There should be a globally accessible instance of `Timer[Task]` or `Clock[Task]` (I'm not sure if we've caught up with changes to cats-effect, so there might not be, but it was the case previously), which should remove the need to wrap the thing into eval and maybe even use `deferAction` at all.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1916,6 +1917,34 @@ sealed abstract class Task[+A] extends TaskBinCompat[A] with Serializable {
...
+
+  /**
+    * TODO: doc
+    *
+    * @return
+    */
+  final def timed: Task[(Duration, A)] =
+    this
+      .timedAttempt
+      .flatMap { case (duration, attempted: Either[Throwable, A]) =>
```
oleg-py: Why not implement `timed` on its own, and then `timedAttempt` would become just `.attempt.timed`?

guizmaii: Because I'm stupid 😅I didn't see the possibility to do `.attempt.timed`.

I thinks that `timedAttempt` isn't useful so.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
guizmaii: It's not what @Avasil says here: https://github.com/monix/monix/issues/714#issue-358293580 or on Twitter.

The `Timer` thing was implemented in Cats-Effect after the Monix 3 RC1, I think. So, I'm not sure that this concept is implemented here, for now at least but I could be wrong. I don't know all the Monix code.

oleg-py: Yeah, there's actually `Task.clock` method on a companion as a shortcut.

Here's a [test suite](https://github.com/monix/monix/blob/master/monix-eval/shared/src/test/scala/monix/eval/TaskClockTimerAndContextShiftSuite.scala#L30), you might get some hints from that.

Also, I think `clockRealTime` is more intended for measurements than `clockMonotonic`.

guizmaii: I'll take a look ! Thanks for the help Oleg ! :)

oleg-py: Actually, disregard the latest bit. I always mix the two up, sorry!

guizmaii: > Actually, disregard the latest bit. I always mix the two up, sorry!

Like everyone I think :)

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1916,6 +1915,42 @@ sealed abstract class Task[+A] extends TaskBinCompat[A] with Serializable {
     */
   final def uncancelable: Task[A] =
     TaskCancellation.uncancelable(this)
+
+  /** Times the `Task` execution and returns its duration and the computed value.
+    *
+    * Basic usage example:
+    *
+    * {{{
+    *   import scala.concurrent.duration.NANOSECONDS
```
alexandru: Why is the import needed in this example?

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1916,6 +1915,42 @@ sealed abstract class Task[+A] extends TaskBinCompat[A] with Serializable {
...
+    *
+    * To be able to time the execution, even if the `Task` fails, you have to call `.attempt` first:
+    *
+    * {{{
+    *   Task(1 / 0)
+    *     .attempt
+    *     .timed
+    *     .flatMap {
+    *       case (duration, Right(value)) => Task.eval(Logger.info(s"executed in ${duration.toMillis} ms")).map(_ => value)
+    *       case (duration, Left(e))      => Task.eval(Logger.warn(s"failed in ${duration.toMillis} ms")).flatMap(_ => Task.raiseError(e))
```
alexandru: Ah, one problem ... `$` will make ScalaDoc complain, you need to escape it.

alexandru: I think the escaping is done by doubling it, e.g. `$$`. Test with `sbt unidoc`.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
guizmaii: typo. Thanks !

guizmaii:
@alexandru are you ok with an implementation using for comprehension or do you prefer a `flatMap` based impl ? I know that the desugaring of the for comprehension is not that great.

alexandru:
TBH I don't know what that desugars into. It's probably OK like this.

alexandru:
Actually, it's your decision. If you think it adds overhead, then desugar it.

guizmaii:
I'm not a big fan of for comprehension, so I prefer the desugared version.

oleg-py:
I find `for` version more readable: it's obvious what the steps are and how dependent they are on each other. And as a local expert on for comprehension, I can assure that it desugars into plain map/flatmap calls, so there's no win 😄

guizmaii:
> And as a local expert on for comprehension, I can assure that it desugars into plain map/flatmap calls, so there's no win

😄

Ok. Thanks you :)

guizmaii:
@oleg-py rollbacked to the `for` version ;)

guizmaii:
> If you have more ideas for cool utility functions then keep them coming even if you don't have the time to implement. :D

Ok =)

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1916,6 +1915,40 @@ sealed abstract class Task[+A] extends TaskBinCompat[A] with Serializable {
...
+    *
+    * {{{
+    *   Task(1 + 1)
+    *     .timed
+    *     .flatMap { case (duration, value) =
+    *       Task.eval(Logger.info("executed in " + duration.toMillis + " ms").map(_ => value)
+    *     }
+    * }}}
+    *
+    * In the previuos example, if the initial `Task` fails, the following `flatMap` will not be executed.
```
Wogan: typo: `previous`

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1916,6 +1915,40 @@ sealed abstract class Task[+A] extends TaskBinCompat[A] with Serializable {
...
+    *   Task(1 / 0)
+    *     .attempt
+    *     .timed
+    *     .flatMap {
+    *       case (duration, Right(value)) => Task.eval(Logger.info("executed in " + duration.toMillis + " ms")).map(_ => value)
+    *       case (duration, Left(e))      => Task.eval(Logger.warn("failed in " + duration.toMillis + " ms")).flatMap(_ => Task.raiseError(e))
+    *     }
+    * }}}
+    */
+  final def timed: Task[(Duration, A)] =
```
Wogan: perhaps the type should be `Task[(FiniteDuration, A)]`?

guizmaii: Could question. What's the difference ?

guizmaii: Ok. Thanks

PR #718, opened 2018-09-11
fix Task scaladocs

jozic:
Fixes all scaladocs for `Task`, thus completing fixing scaladocs for `monix-eval` project

@oleg-py @alexandru please review

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -85,9 +85,10 @@ import scala.util.{Failure, Success, Try}
   *
   * {{{
   *   import monix.execution.CancelableFuture
+  *   import monix.execution.Scheduler.Implicits.global
   *
   *   val f: CancelableFuture[Unit] = sayHello.runAsync
-  *   //=> Hello World!
+  *   // => Hello World!
   * }}}
   *
   * The returned type is a
@@ -103,14 +104,16 @@ import scala.util.{Failure, Success, Try}
   * has real consequences. For example with `Task` you can do this:
   *
   * {{{
+  *   import scala.concurrent.duration._
+  *
   *   def retryOnFailure[A](times: Int, source: Task[A]): Task[A] =
-  *     source.onErrorRecoverWith { err =>
+  *     source.onErrorHandleWith { err =>
   *       // No more retries left? Re-throw error:
   *       if (times <= 0) Task.raiseError(err) else {
   *         // Recursive call, yes we can!
   *         retryOnFailure(times - 1, source)
   *           // Adding 500 ms delay for good measure
-  *           .delayExecution(500)
+  *           .delayExecution(500.millis)
   *       }
   *     }
   * }}}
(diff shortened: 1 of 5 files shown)


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -3871,7 +3929,11 @@ private[eval] abstract class TaskInstancesLevel1 extends TaskInstancesLevel0 {
     * {{{
     *   import cats.syntax.all._
     *
-    *   (taskA, taskB, taskC).parMap { (a, b, c) =>
+    *   val taskA = Task("a")
+    *   val taskB = Task("b")
+    *   val taskC = Task("c")
```
oleg-py: See my comment above regarding using `sleep`. Also maybe add a note that it would complete in about 3 seconds, if 3 is your longest delay.

jozic:
Thank you @oleg-py for your kind and encouraging words!
i've addressed your comments.

alexandru:
@jozic just dropping a note that improving the ScalaDocs is the kind of work that really makes a difference. Thanks!

jozic:
thanks @alexandru 

i've checked travis logs and it fail because code samples like 
```
    *         (() => ec.execute(() => {
    *           try
    *             cb.onSuccess(thunk)
    *           catch { case NonFatal(e) =>
    *               cb.onError(e)
    *           }
    *         })): Runnable,
```
don't compile for 2.11 (because of SAM types)
we can add `-Xexperimental` to make it available on 2.11 or rewrite examples to not use SAM 

which way do you guys prefer it?

alexandru:
I would change that to a plain `Runnable`. That cast looks weird anyway.

Or in case `ec` is a Monix `Scheduler`, another thing you could do is to use `executeAsync`, which is an extension method.

jozic:
jvm builds pass, but js ones don't as samples use jvm only methods, like `runSyncUnsafe` and `Scheduler.io`
any ideas how to solve that?

alexandru:
Can't we stop this verification for JS?

Alternatively, remove/ rewrite whatever causes issues.

jozic:
>Can't we stop this verification for JS?
of course we can, silly me :)

just enabled `doctestSettings` for `evalJVM` only, looks better now

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -154,7 +161,7 @@ import scala.util.{Failure, Success, Try}
   *
   * {{{
   *   // Some array of tasks, you come up with something good :-)
-  *   val list: Seq[Task[Int]] = ???
+  *   val list: Seq[Task[Int]] = Seq(Task(10), Task(20), Task(30))
   *
   *   // Split our list in chunks of 30 items per chunk,
   *   // this being the maximum parallelism allowed
```
oleg-py: Actually, I didn't realize it was splitting the list in the chunk of 30 below... Do `Seq.tabulate(100)(Task(_))`?

jozic: looks good

jozic:
how about now? good to go?

alexandru:
I merged it because some other changes are coming from myself and I don't want to break this merge.

If you find anything else, open different PRs.

PR #721, opened 2018-09-13
WIP: make Task auto-cancelable by default, remove Task.Options from API

alexandru:
This is a broken work in progress that I'm publishing as PR for early feedback.

PR changes `Task` to be auto-cancelable by default, thus with enabled auto-cancelable run-loops on `runAsync`. But the other, more significant change is that I want to get rid of `Task.Options` from the API entirely, I want it gone and all functions exposing it deleted.

This is an example of convenient complexity that has been creeping it and it's time to take a step back and simplify the API a little.

Changes needed:

- [x] make `Task` auto-cancelable by default
- [x] move `Task.Option` to `internals.TaskOptions` to completely hide it from the API
- [x] move `Task.Context` to `internals.TaskContext`, as it can no longer be public
- [x] remove `autoCancelable` operation on `Task`, to be replace with `continual` (which activates the non-cancelable run-loop)
- [x] remove `Task.executeWithOptions`
- [x] remove all `runAsyncOpt` variants and overloads (no more `Opt` execution)
- [x] `Effect[Task]` no longer takes a `Task.Option` implicit
- [x] `TaskApp` no longer exposes an implicit `Task.Option`
- [ ] `TaskLocal` initialization and usage needs to use `cats.effect.Resource`
- [ ] tests are terribly broken, need to be fixed
- [ ] ...

So one problem is that `TaskLocal` needs to have `localContextPropagation` set to `true` in Task's run-loop in order for it to work. This has always been a problem, its usage being very error prone due to the possibility that changes are not transported over async boundaries.

But we now have `Resource` from Cats-Effect and `TaskLocal` does in fact behave like a resource.
Therefore I'm thinking that the builder should look like this:

```scala
object TaskLocal {

  def apply[A](default: A): Resource[Task, TaskLocal[A]] =
     ???
}
```

This solves several problems:

1. on `use` a `release` will happen with which we can ensure that the internal `TaskOptions` gets restored to its previous value and maybe that the local gets restored to the `default`
2. the `TaskLocal` could have a "closed" state, so we can ensure safety by verifying that the local hasn't been released yet; therefore if the value escapes its `Resource#use` context, the user can be warned about it with an error
3. The Cats-Effect `Resource` can be transformed into an `Iterant` or into an FS2 `Stream`, with the right semantics (e.g. `flatMap` delays the closing of the resource until the end), so `TaskLocal` can be used with `Iterant` or FS2's `Stream` as well

This is a work in progress. Would like feedback!

/cc @Avasil @oleg-py


oleg-py:
I'm slightly 👎 on `Resource[TaskLocal]`. My crazy usage scenarios like logging and call tracing are even better without RT constructors 😅, so that's just more nuisance to work around then.

alexandru:
@oleg-py I would be interested in some samples and I think we can provide “unsafe” constructors if you think they’re useful.

But I want us to get rid of those Options from the API, I hate having them due to complexity and I don’t want to enable locals by default due to the performance degradation.

oleg-py:
@alexandru actually, all of my "smart" code right now relies on the fact that `Local`s are propagated together with `TaskLocal`. I don't think Resource would work here 🗡 

Can we just check if we have anything in the context map and enable local context propagation if we do, on async boundaries?

alexandru:
@oleg-py can you show me a piece of code?

I don't understand why `Resource[Task, TaskLocal]` wouldn't work because it is similar with `Task[TaskLocal]`, only better because `Resource` provides better scoping. So if you can make your code work with `Task[TaskLocal]`, you can make it work with `Resource` as well because `Resource#use` is basically like `Task#flatMap`.

So your usage might involve some global state created via unsafe constructs, instead of passing around such locals as parameters, as is recommended with FP usually. Or do you have something else in mind?

I'm open to modifying the run-loop to make it enabled all the time, however that's a cost that all users of Task will have to endure instead of just users of TaskLocal, therefore I need to understand the use-cases. I think you showed me some code before, but no longer have any links.

Because if we can get away with `Resource`, I think that's the better approach.

oleg-py:
@alexandru an example I have ready is my blog article [here](https://olegpy.com/better-logging-monix-1/)

pettyjamesm:
Agreed on the usage of Local, we're also depending on the propagation of context for distributed tracing. Having the trace context information available in logs is one such case.

I agree that trace context often behaves like Resource in that there is a bounding scope involved, but inside of that scope tracing API's usually need a way to grab whatever the local context is directly.

Question: since we're simulating "ThreadLocal" like behavior over async boundaries, isn't the run loop mostly not affected? I would assume only async context shifts to pay the cost of transporting the local context over the async boundary.

alexandru:
@pettyjamesm 

> I agree that trace context often behaves like Resource in that there is a bounding scope involved, but inside of that scope tracing API's usually need a way to grab whatever the local context is directly

I was thinking of using `Resource` because that way we can safely enable this feature (for everything that then goes on inside that resource context), but it would work just as now and you'd have access to raw `Local` as well.

> Question: since we're simulating "ThreadLocal" like behavior over async boundaries, isn't the run loop mostly not affected? I would assume only async context shifts to pay the cost of transporting the local context over the async boundary.

The run-loop handles jumping over async boundaries as well and so it is affected. And the cost is not insignificant.

pettyjamesm:
I'm not sure I understand. While the loop does manage where to jump over the async boundaries, I would expect there to be no additional cost for the cases where the run loop does not go async because the existing state continues to be valid unless explicitly modified. Are you saying that the additional cost of going async is the worry, or is there some hidden cost per iteration of the run loop that I'm not aware of?

alexandru:
> I would expect there to be no additional cost for the cases where the run loop does not go async

Yes, of course, if the processing stays synchronous there is no cost. But Monix's run-loop also auto-forks in batches by default, this being the `BatchedExecution` mode, configured by default for 1024 iterations, so it automatically forks after 1024 flatMaps and it also has an `AlwaysAsyncExecution` mode where it forks on each `flatMap`.

pettyjamesm:
Ok, so I'm not missing some hidden understanding of the synchronous case. The question of cost is only about not taxing performance of code that doesn't need the context propagation.

I suspect you know how to go about reducing that overhead better than I do, so I won't blindly suggest solutions to problems I don't fully understand- I just hope that I won't lose the ability to trace execution using Local. That's a huge feature that can't be easily added after the fact and one of the biggest factors in our choice to use Monix over the standard library futures.

alexandru:
@pettyjamesm we aren't losing the ability to use raw `Local` values.

And I am currently evaluating what to do next. But I no longer want users to do `executeWithOptions(_.enableLocalsPropagation)`, because it is error prone and a source of complexity. I want the notion of `Task.Options` to be gone from the public API, remaining only an internals thing.

And it's a good thing that you mentioned you are users of locals, as I'm going to ask for feedback 🙂

pettyjamesm:
I'm happy to do it. If you think it'll be easier to do via email or on a call, I'm happy to do that too. I've got had to use a few other workarounds with Local as implemented in RC1. The changes that have landed since RC1 look like they might have addressed those issues, but I haven't been able to verify since those haven't landed in a release yet.

oleg-py:
@alexandru so, how do I ensure LCP is enabled without using Resource? I think `Local` propagating along with `TaskLocal` is a neat feature for integrating APIs that rely on old bad `ThreadLocal`

alexandru:
@oleg-py in your blog post you're relying on `monix.execution.misc.Local` for the dirty part.

In that case you can simply create a `Local` and then wrap it in a `TaskLocal`.

```scala
val local = Local(0)

TaskLocal.wrap(Task(local)).use { local =>
  ???
}
```

oleg-py:
@alexandru I'd rather not have to explain this... but well, guess that's something

alexandru:
Yeah, but before this you had to explain that you had to do `task.executeWithOptions(_.enableLocalContextPropagation)`. At the very least this way the signature of the thing makes clear its usage, `Resource` being in your face.

This is a larger topic btw. I want us to push features in `Task` that are optional, without making all the use-cases that don't use those features to suffer from it. After the fixes to `TaskLocal`, its usage in the run-loop is significant, plus we are also wrapping all `Scheduler` values with `TracingScheduler` whenever we give it to users, etc... all to prevent locals leaking. It's not cheap.

oleg-py:
It's not clear that you need that magic rite to link the Local to Task execution :) Furthermore, it's not unlikely Local-for-Java-APIs usage would prefer to keep it private, so it's gotta be a dummy value. With `executeWithOptions` at least it's clear what the code does from what the code says.

BTW, what's wrong with just checking the size of context map instead of having the option at all?

oleg-py:
IOW, I understand why you're pushing for Resource. Just saying that for my use-cases API has became less convenient and straightforward.

alexandru:
> BTW, what's wrong with just checking the size of context map instead of having the option at all?

It's not as simple as that. Internally based on that option things get initialized differently.
For example this is a piece of code from `Task.Context`:

```scala
    val scheduler: Scheduler =
      if (options.localContextPropagation)
        TracingScheduler(schedulerRef)
      else
        schedulerRef
```

You can see how this can get complicated, because we do this in order to make `Task.async` safe. Also the restoration of the locals in most cases happens after an async boundary regardless if the map was empty or not before the jump. Because in a `Task.async` for example you don't care about what happens inside it, you just want the locals restored for its bind continuation, because in a `Task.async` the user might have forked using some thread-pool you cannot control.

So in that case it doesn't really matter if the map was empty or not. You still want to restore it, empty as it is, because if you had the defaults going in, you still want the defaults going out.

On `restartAsync`, which is the piece of code executing for the auto-fork behavior, we have this which is what you mean:

```scala
    val savedLocals =
      if (context.options.localContextPropagation) Local.getContext()
      else null
```

However we also have in `RestartCallback` logic, starting from the builder:

```scala
  def apply(context: Context, callback: Callback[Any]): TaskRestartCallback = {
    if (context.options.localContextPropagation)
      new WithLocals(context, callback)
    else
      new NoLocals(context, callback)
  }
```

And it's [TaskRestartCallback.WithLocals](https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/internal/TaskRestartCallback.scala#L150) doing the heavy lifting, the logic there being more complicated.

Either way, I know it sucks, but the point still stands ... some of these features have a high cost and we can't shove them all in the run-loop without the ability to turn them ON or OFF as needed. In this case reading and writing `ThreadLocal` values and making schedulers to be tracing is expensive.

We don't have the budget for it and if we allow this one by default, the next idea that we'll have requiring changes in the run-loop will be automatically rejected due to having no budget left.

---

So if we don't do `Resource`, it means we are stuck with `executeWithOptions`.

Will do another PR just with the auto-cancellation changes.

oleg-py:
@alexandru good point. I think Resource is a good pattern to use, would just prefer there to be a convenient option to turn LCP on for "unsafe" usage scenarios. Maybe introduce Resource, but not get rid of Task.Options just yet?

alexandru:
@oleg-py please look at the latest.

`Task.Options` is still gone. `Resource` is still used.

What I added is a `TaskLocal.Unsafe` object with unsafe stuff in it. You have more flexibility than before, except for the ability to set "options" at the edge.

```scala
TaskLocal.Unsafe.scopedTask {
  for {
    local <- Task(TaskLocal.Unsafe.create(0))
    _     <- local.write(10)
    _     <- Task.shift
    value <- local.read
  } yield value
}
```

alexandru:
Also, I have no more energy for this, have been toying around with multiple alternatives for how to proceed, this is the best that I could do.

alexandru:
So two things:

1. `TaskLocal.Unsafe.scopedTask` is basically the old `.executeWithOptions(_.enableLocalContextPropagation)`
2. you now also have unsafe builders that break referential transparency, so you can knock yourself out

alexandru:
Usage btw is greatly improved, also because I made those tasks yield errors in case the underlying support isn't active, so at the very least it is loud about the underlying support not being active.

Would appreciate a review.

alexandru:
Planning to merge this tonight btw. Immediately after I un-comment and fix the remaining tests.

oleg-py:
What I can say, except 👍 ? Unsafe stuff is very important to have too, thanks in advance for putting it in :)

alexandru:
Unfortunately I realized that I can't expose `Resource` values because `ContextSwitch` cannot work in the way I used it ... due to Task implementation details ☹️

Fixing it involves making the internal `Context` object mutable and that's not something I'm prepared to do because the complexity is overwhelming and I can't see if the implementation would be correct in all cases or not.

Therefore `Task.Options` has to stay.

PR #726, opened 2018-09-24
Change Task to sequence (back-pressure) async finalizers

alexandru:
This change is related to the changes I also did in Cats-Effect's `IO` implementation, see:

- [Issue 267](https://github.com/typelevel/cats-effect/issues/267)
- [PR 305](https://github.com/typelevel/cats-effect/pull/305)

`Task` changes to sequence async finalizers, specified via `bracket` or via `Task.cancelable`. What this means is that:

1. internally `Task[Unit]` is used, in addition to `Cancelable` values for specifying finalizers
2. triggering cancellation will now use and chain `Task` values
3. we expose additional `runAsync` variants:
  - `runAsyncF` that returns a `Task[Unit]` (aliased `CancelToken[Task]`)
  - `runAsyncAndForget`, an optimized way to run tasks in cases in which we care not about the result or about cancelling it
  - `runAsyncUncancelable`, which doesn't return a cancellation token
4. to make room for these, I also deprecated `runAsyncMaybe` (which was useless)

I also started a new sub-project of Monix called `monix-catnap`. This will provide utilities built on Cats-Effect's type classes. And the first traits introduced are:

- `monix.catnap.CancelableF`
- `monix.catnap.cancelables.BooleanCancelableF`

I don't have time to introduce pure versions of all cancelable types in `monix.execution`, but that's the plan. I also plan to introduce specialized `Ref` implementations, like we have in `monix.execution.atomic`. So `catnap` intends to provide generic and pure equivalents for what we have in `execution`.


monix-catnap/shared/src/main/scala/monix/catnap/CancelableF.scala
```diff
@@ -0,0 +1,158 @@
...
+    * Returns a dummy [[CancelableF]] that doesn't do anything.
+    */
+  def empty[F[_]](implicit F: Applicative[F]): CancelableF[F] =
+    new CancelableF[F] { def cancel = F.unit }
+
+  /** Builds a [[CancelableF]] reference from a sequence,
+    * cancelling everything when `cancel` gets evaluated.
+    */
+  def collection[F[_]](refs: CancelableF[F]*)(implicit F: Sync[F]): CancelableF[F] =
+    unsafeApply[F](cancelAll(refs:_*))
```
oleg-py: Should this not be `@UnsafeBecauseImpure`?

alexandru: The actual implementation has no state. Should it be?

alexandru: I mean, if it's unsafe, I'd rather wrap it in `F[_]`, but figured there's no point in doing it.

Not sure if I'm right. Maybe I should just wrap it in `F[_]` to be on the safe side.

oleg-py: I thought the `unsafeApply` allocates mutable state, no?

alexandru: Ugh, yes, that's a good point. I think my intention was to use `wrap`, because `CancelableF` references have to ensure idempotency on their own, no extra protections needed for such a "collection".

Thanks for pointing it out, definitely needs a fix.

oleg-py: Glad it helped 👍

Issue #735, opened 2018-10-05
Observable.bufferTimedWithPressure operator with custom counting function


ybr:
I need an operator that consider the sum of the size of elements as the count instead of the count of elements.

Something like:
def bufferTimedWithPressure[A](period: FiniteDuration, maxSize: Int, sizer: A => Int): Observable[Seq[A]]

The existing one could be rewritten in terms of the new:
def bufferTimedWithPressure[A](period: FiniteDuration, maxSize: Int): Observable[Seq[A]] = bufferTimedWithPressure(period, maxSize, _ => 1)

I implemented a version in our code base, I can propose a pull request.

ybr:
This issue look sa lot like https://github.com/monix/monix/issues/447 but on another operator.

Avasil:
If you already have implementation, PR should help in decision and having it in time for `3.0.0` :)

ybr:
Ok @Avasil I'll try my best to submit it promptly :)

PR #739, opened 2018-10-08
Add operator bufferTimedWithPressure with sizeOf on Observable

ybr:
- Changed the contract of BufferWithSelectorObservable to take an extra paramter sizeOf: A => Int
- Changed the way BufferWithSelectorObservable computes the size of the buffer using sizeOf
- Added the operator Observable[A].bufferTimedWithPressure(period: FiniteDuration, maxSize: Int, sizeOf: A => Int): Observable[Seq[A]]
- Changed operators using BufferWithSelectorObservable with default sizeOf set to 1 to keep the same behaviour
- Added myself to AUTHORS

Monix issue:
 https://github.com/monix/monix/issues/735


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -739,7 +739,47 @@ abstract class Observable[+A] extends Serializable { self =>
     */
   final def bufferTimedWithPressure(period: FiniteDuration, maxSize: Int): Observable[Seq[A]] = {
```
Avasil: I think we could leave one method with default `sizeOf` argument. It should also fix doc ambiguity issue

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/BufferWithSelectorObservable.scala
```diff
@@ -53,7 +54,8 @@ private[reactive] final class BufferWithSelectorObservable[+A,S](
           upstreamSubscriber.synchronized {
             if (downstreamIsDone) Stop else {
               buffer += elem
-              if (maxSize > 0 && buffer.length >= maxSize)
+              if (maxSize > 0 &&
+                  buffer.foldLeft(0)(_ + sizeOf(_)) >= maxSize) // sum size of elements in the buffer
```
Avasil: I wonder about performance penalty for bigger buffers (goes from O(1) to O(N)). Maybe we should track `buffer` size? What do you think @alexandru - is it something to worry about?

ybr: @Avasil I agree. The point of having the sizeOf paremeter and in our case was to decrease dramatically the count of elements in the buffer we reach at most 10 elements each containing x100K values so the penalty is very low. But since sizeOf can be hard to compute it certainly worths to care about that.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
ybr: Ok I'm going to change that. Thank you.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/BufferWithSelectorObservable.scala
ybr: @Avasil Yes I agree with you. Furthermore, sizeOf could be resource consuming so it will surely becomes unacceptable anyway.

Issue #742, opened 2018-10-14
Observable.collect causes double evaluation of partial functions


jfwilson:
The `Observable.collect` operator invokes separate calls to `.isDefinedAt` and then `.apply`. This results in any guard/unapply logic being evaluated twice, or in the case of `Function.unlift`ed partial functions, the entire function being evaluated twice.

The scala collections use `PartialFunction.applyOrElse` with a fallback value to avoid this double evaluation. Suggest `monix` does the same.

PR #743, opened 2018-10-14
Observable.collect improvement to avoid double evaluation (#742)

jfwilson:
Avoid double evaluation of `PartialFunction`s in `CollectOperator`.

I have replaced `.isDefinedAt` and `.apply` with a single call to `.applyOrElse` in the same way as done in standard scala collection `collect` implementations. This has the benefit that any guard or unapply logic within the partial function is only evaluated once.

This does require a small amount of casting (of a magic fallback value) to work. Alternatively could use `PartialFunction.lift` to return an `Option[B]`, but I'm less keen on this because in many cases this would require allocating an extra `Some(_)` on each element and the type casting is still happening (but is done inside `PartialFunction.Lifted` instead).


alexandru:
Thanks @jfwilson

PR #745, opened 2018-10-15
Avoid buffer size overflow in MapAsyncParallelSubscription

ddworak:
Targetting `series/2.x` - this no longer seems to be the case in 3.x.

When parallelism was sufficiently high in `.mapAsync`, it could overflow after internal `Platform.recommendedBatchSize` addition.


ddworak:
Well, I can obviously workaround it, but it would be nice to just upgrade at some point in our older projects which won't go Monix 3 soon.

PR #750, opened 2018-10-24
Fix handling start/end of Long range in RangeObservable

cakper:
Hello,

I've encountered a problem with `RangeObservable` that doesn't correctly handle start/end of Long range - current behaviour is that it overflows Long and continues emitting elements.

Problem happens both in Monix 2/3 and would be nice to have it fixed in 2 branch as at the moment I'm using a workaround (use case is that I need to scan whole Long range).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -48,3 +48,6 @@ https://github.com/lmnet
 
 Yohann Bredoux
 https://github.com/ybr
+
+Kacper Gunia
+https://github.com/cakper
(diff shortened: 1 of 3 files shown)


alexandru:
👍

Avasil:
Thanks @cakper can you make second PR targeting series/2.x branch?

There is another PR with bugfix for 2.x (#745 ) so we could get both in release

cakper:
@Avasil sure, many thanks for the merge. I'll get the other PR ready today :)

PR #759, opened 2018-11-01
Add Contravariant Observer and Subscriber

### LINKED ISSUE #748: contramap on Subscriber
We have `contramap` for `Consumer`, now we also have `Profunctor` instance for `Subject` and it might be useful to have `Contravariant` for `Subscriber`. 

[Implementation of Consumer.contramap](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/consumers/ContraMapConsumer.scala)

Avasil:
Fixes #748

TODO:

To test laws we need to implement `Eq` and `Gen` for `Observer` / `Subscriber`. So far I don't have any ideas that make sense so I'm issuing PR in case someone does


monix-reactive/shared/src/main/scala/monix/reactive/instances/CatsContravariantForObserver.scala
```diff
@@ -0,0 +1,61 @@
...
+  * References:
+  *
+  *  - [[https://typelevel.org/cats/ typelevel/cats]]
+  */
+object CatsContravariantForObserver extends Contravariant[Observer] {
+  override def contramap[A, B](fa: Observer[A])(f: B => A): Observer[B] =
+    new ContravariantObserver(fa)(f)
+}
+
+private[reactive] class ContravariantObserver[A, B](source: Observer[A])(f: B => A) extends Observer[B] {
```
alexandru: Make this `final` and a `private` member of `object `CatsContravariantForObserver`.

`private[package]`  is useful if you plan on reusing it within `package`, but it's riskier than a plain `private`, because `private[package]` is a Scala construct, not a Java one.

Avasil: @alexandru Ok, corrected in in all three cases!

monix-reactive/shared/src/main/scala/monix/reactive/instances/CatsContravariantForObserver.scala
```diff
@@ -0,0 +1,61 @@
...
+  *
+  *  - [[https://typelevel.org/cats/ typelevel/cats]]
+  */
+object CatsContravariantForObserver extends Contravariant[Observer] {
+  override def contramap[A, B](fa: Observer[A])(f: B => A): Observer[B] =
+    new ContravariantObserver(fa)(f)
+
+  private final class ContravariantObserver[A, B](source: Observer[A])(f: B => A) extends Observer[B] {
+    // For protecting the contract
+    private[this] var isDone = false
```
Avasil: @alexandru What's the rule whether we need this protection or not? Sometimes I see this sometimes I don't

Avasil:
@alexandru Done! :)

Issue #758, opened 2018-10-31
Add Observable#filterTask


yanns:
As we are using `Task` for everything, including logging, we would like to use `Observable.filter` but with a `Task[Boolean]` instead of a `Boolean` to be able to log what we needs.

Currently we need some boilerplate code, like:
```
      source
        .mapTask { element ⇒
          infoT(s"The log about the $element") >> Task.pure((element, element.keep))
        }
        .filter(_._2)
        .map(_._1)
```
Instead of:
```
      source
        .filterTask { element ⇒
          infoT(s"The log about the $element") >> Task.pure(element.keep)
        }
```

Avasil:
I think there should be `flatTap` if you have cats import to do the logging before but it's still in two passes so having `filterEval` could have its use. Would you like to try doing PR?

yanns:
Yes, I could try to do a PR for this if you think it's a good idea to have this in the library.
What about the naming? `filterTask` would be good?

alexandru:
@yanns the convention changed, functions taking a `Task` action should be named with an `Eval` suffix, instead of `Task`

yanns:
@alexandru thx for the quick feedback.
Do you think that this addition would be useful?

alexandru:
@yanns sure, go ahead.

But if we have `filterEval`, we need `filterEvalF` too.

PR #760, opened 2018-11-01
add Observable.filterEval

yanns:
Fix https://github.com/monix/monix/issues/758

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -3711,6 +3711,11 @@ abstract class Observable[+A] extends Serializable { self =>
 
   /** Only emits those items for which the given predicate holds.
     *
+    * @see [[filterEval]] for a version that works with a [[monix.eval.Task]].
+    * @see [[filterEvalF]] for a version that works with a generic
+    *      `F[_]` (e.g. `cats.effect.IO`, Scala's `Future`),
+    *      powered by [[monix.eval.TaskLike]]
+    *
     * @param p a function that evaluates the items emitted by the source
     *        returning `true` if they pass the filter
     * @return a new observable that emits only those items in the source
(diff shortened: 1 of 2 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -3719,6 +3724,33 @@ abstract class Observable[+A] extends Serializable { self =>
...
+    *
+    * @see [[filterEvalF]] for a version that works with a generic
+    *      `F[_]` (e.g. `cats.effect.IO`, Scala's `Future`),
+    *      powered by [[monix.eval.TaskLike]]
+    */
+  final def filterEval(p: A => Task[Boolean]): Observable[A] =
+    self
+      .mapEval(a ⇒ p(a).map((a, _)))
+      .filter(_._2)
+      .map(_._1)
```
alexandru: You can replace `filter` and `map` with `collect`. It would be slightly more efficient.

`filter(p).map(f) <-> collect { x if p(x) => f(x) }`

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -3719,6 +3724,33 @@ abstract class Observable[+A] extends Serializable { self =>
...
+    *
+    * So you can work among others with:
+    *
+    *  - `cats.effect.IO`
+    *  - `monix.eval.Coeval`
+    *  - `scala.concurrent.Future`
+    *  - ...
+    */
+  final def filterEvalF[F[_]](p: A => F[Boolean])(implicit F: TaskLike[F]): Observable[A] =
+    filterEval(a => Task.from(p(a))(F))
```
alexandru: This too needs a test.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
alexandru: There no way we're going to do that, since it requires the logic of `mapEval` and `mapEval` is extremely complicated and that logic is already duplicated from `flatMap`.

yanns: https://github.com/monix/monix/pull/760/commits/0c01301b27b6434ff91a67af2ffe8a795552f7d7

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
yanns: https://github.com/monix/monix/pull/760/commits/839b500082d314ea6a3b732afff6c5ddbe8b4c95

PR #790, opened 2018-12-05
Fix #780: improve fixed rate scheduling on JS

### LINKED ISSUE #780: AsyncScheduler.scheduleAtFixedRate should schedule at the specified fixed rate with ScalaJS
The documentation for `scheduleAtFixedRate` states:

>Schedules a periodic task that becomes enabled first after the given
>    * initial delay, and subsequently with the given period. Executions will
>    * commence after `initialDelay` then `initialDelay + period`, then
>    * `initialDelay + 2 * period` and so on.

In practice, however, the Scala.js implementation in Monix 3.0.0-RC2 drifts, adding a small amount of unwanted additional delay between executions.

This program shows the problem:

```scala
import monix.execution.Scheduler.{global => scheduler}

object App {
  def main(args: Array[String]): Unit = {
    val start = scheduler.clockMonotonic(MILLISECONDS)

    scheduler.scheduleAtFixedRate(1.second, 1000.milliseconds) {
      println(s"${scheduler.clockMonotonic(MILLISECONDS) - start} ms")
    }
  }
}
```

When compiled with Scala.js 0.6.25 + Scala 2.12.7 (with `fastOptJs`) and run under the current versions of both Firefox and Chrome, the console output is along the following lines:

```
1015 ms
2017 ms
3017 ms
4019 ms
5021 ms
6022 ms
7022 ms
8024 ms
9026 ms
10027 ms
11028 ms
12029 ms

...

95171 ms
96172 ms
97173 ms
98175 ms
99176 ms
100178 ms
```

As can be seen from this run with Chrome 70.0.3538.110 on Windows 10 October 2018 Update, the execution is 163 ms late after 100 seconds.

Many thanks for Monix, and for your attention on this.

oleg-py:
This is a simple fix that seems to give most consistency. The idea is basically to add scheduling lag time to task execution time. Test is a bit non-deterministic (and I don't know how Travis would accept these), but change seems to reduce the accumulated error by a factor of magnitude in this simple scenario.

### DIFF
--- a/monix-execution/js/src/test/scala/monix/execution/internal/AsyncSchedulerSuite.scala
+++ b/monix-execution/js/src/test/scala/monix/execution/internal/AsyncSchedulerSuite.scala
@@ -20,7 +20,8 @@ package monix.execution.internal
 import minitest.TestSuite
 import monix.execution.atomic.Atomic
 import monix.execution.schedulers.{AsyncScheduler, StandardContext}
-import monix.execution.{ExecutionModel, Scheduler, UncaughtExceptionReporter}
+import monix.execution.{Cancelable, ExecutionModel, Scheduler, UncaughtExceptionReporter}
+
 import scala.concurrent.Promise
 import scala.concurrent.duration._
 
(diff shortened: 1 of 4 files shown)


alexandru:
It's pretty good.

Another idea would be to have a start time and then schedule at `start + period`, `start + period * 2`, `start + period * 3` and so on. Not sure how well it works.

Merging.

PR #798, opened 2018-12-18
Update intervalWithFixedDelay scaladoc

Avasil:
Fixes one mistake and adds new sentence that is also present in `intervalAtFixedRate`

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -4855,10 +4855,13 @@ object Observable extends ObservableDeprecatedBuilders {
     new builders.Interleave2Observable(oa1, oa2)
 
   /** Creates an Observable that emits auto-incremented natural numbers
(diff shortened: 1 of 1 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -4855,11 +4855,14 @@ object Observable extends ObservableDeprecatedBuilders {
     new builders.Interleave2Observable(oa1, oa2)
 
   /** Creates an Observable that emits auto-incremented natural numbers
-    * (longs) spaced by a given time interval. Starts from 0 with no
-    * delay, after which it emits incremented numbers spaced by the
+    * (longs) spaced by a given time interval. Starts from 0 with `initialDelay`, 
+    * after which it emits incremented numbers spaced by the
     * `period` of time. The given `period` of time acts as a fixed
```
Avasil: I feel like `period` should be `delay` (name of the parameter) but I'm not sure

PR #802, opened 2018-12-30
Fix #787: encapsulate local ctx on task execution

### LINKED ISSUE #787: Local state remains on previous thread after context switch
**Monix Version: `3.0.0-RC2`**

## Observed behavior:
When using `Local` the state is left behind on the previous thread after a context switch. For example, given a `Local[Int](default = 0)`, you start on `thread 1` and set `local=42` and then shift to another thread. On `thread 2` you automagically see `local=42`  thanks to `enableLocalContextPropagation`... BUT `local=42` remains on `thread 1`, even though the task is done with that thread. It assumes the next task will clean up before starting, which IMO makes the `default` value somewhat useless (as the value never really means anything useful to you, as it's only in its initial state when a task is arbitrarily scheduled onto a virgin thread).

## Expected behavior:
When context switch occurs `thread 1 / local` should be reverted to the `default=0` value.

## Reproduction steps:

```scala
object Application extends App {
  implicit val scheduler: Scheduler = Scheduler.traced
  val local = Local[Int](0)

  (0 to 1000).foreach { _ =>
    Task
      .eval {
        if (local() > 0) {
          println("Uh-oh!")
        }
        local.update(local() + 1)
      }
      .asyncBoundary // If you comment this out, all works fine... but with it in, demonstrates the context is left behind on the previous thread.
      .map { _ =>
        local.update(local() - 1)
      }
      .executeWithOptions(_.enableLocalContextPropagation)
      .runSyncUnsafe()
  }
}
```

oleg-py:
I basically redid what was in one of my previous PRs. This ensures Local context is left unchanged when an unsafe method starts executing.

Not entirely sure this fixes all the possible issues (I'm suspicious about `.start`), but I think it's better than what we currently have.

### DIFF
--- a/monix-eval/jvm/src/test/scala/monix/eval/TaskLocalJVMSuite.scala
+++ b/monix-eval/jvm/src/test/scala/monix/eval/TaskLocalJVMSuite.scala
@@ -123,4 +123,20 @@ object TaskLocalJVMSuite extends SimpleTestSuite {
     val r = task.runSyncUnsafeOpt(Duration.Inf)
     assertEquals(r, 100)
   }
+
+  test("local state is encapsulated by Task run loop") {
+    import monix.execution.Scheduler.Implicits.global
+    implicit val opts = Task.defaultOptions.enableLocalContextPropagation
+    val local = TaskLocal(0).memoize
+
+    val task = for {
+      l <- local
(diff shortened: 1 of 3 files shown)


alexandru:
I like it.

PR #803, opened 2019-01-07
Eagerly null out dequeued elements in ChunkedArrayQueue

iravid:
This is a very simplistic fix for an issue I'm having, opening this for discussion (and to avoid keeping this in a local fork :-))

G1 Old Gen before the fix:
image

After the fix:
image


alexandru:
Thanks a lot @iravid for investigating this.

iravid:
Happy to help! Any chance you could restart the build?

PR #807, opened 2019-01-15
Improve encapsulation test, encapsulate locals on ContextShift

oleg-py:
I've improved the test to cover more ways to execute Task with LCP. Found a few related to `startFull` not being encapsulated, if the execution starts without LCP and gets it enabled halfway through, such as `task.executeWithOptions(_.enableLocalContextPropagation).unsafeRunSync()`.

### DIFF
--- a/monix-eval/jvm/src/test/scala/monix/eval/TaskLocalJVMSuite.scala
+++ b/monix-eval/jvm/src/test/scala/monix/eval/TaskLocalJVMSuite.scala
@@ -20,8 +20,9 @@ package monix.eval
 import minitest.SimpleTestSuite
 import monix.execution.ExecutionModel.AlwaysAsyncExecution
 import monix.execution.Scheduler
-import scala.concurrent.ExecutionContext
-import scala.concurrent.duration.Duration
+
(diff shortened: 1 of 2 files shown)


alexandru:
@oleg-py this is interesting. However I see that you've modified the run-loop.

So what does that fix?

oleg-py:
@alexandru this is basically a continuation of #802 / #787 fix. I tried to improve test to cover more execution methods, and found that the initial fix is not enough for the cases where you don't use `run...Opt` variants. The case where LCP is enabled by using `executeWithOptions`, it would still cause the "current" thread to not reset Local value to default after task is done executing.

alexandru:
Interesting. OK, if this fixes the problem, then it sounds good, but I hope you tested it well.

alexandru:
Btw, we'll have to make `Observable` compatible too. There are places like the `mapEval` operator, that could use some magic ... i.e. internally we currently use `runAsync`, however if `TracingScheduler` is detected we could enable LCP.

oleg-py:
Done 😄

PR #812, opened 2019-01-18
Update Minitest to 2.3.2

alexandru:
Made changes to Minitest to use standard `System.out.print` instead of sbt's loggers, in order to see which test froze the build.


alexandru:
Argh, it didn't fail 🙂

PR #813, opened 2019-01-18
Disable code coverage

alexandru:
Disabling automated code coverage because a 30 minutes overhead for Travis isn't worth it.

Now it's much better: https://travis-ci.org/monix/monix

image


alexandru:
I've also created an issue in `sbt-scoverage`: https://github.com/scoverage/sbt-scoverage/issues/276

PR #814, opened 2019-01-20
WIP: Make TaskLocal behave well in combination with Observable

alexandru:
This PR changes `Observable` operators that make use of `Task` to detect `TracingScheduler` and enable the `localContextPropagation` option when executing the tasks.

This should make `Observable` play well with `TaskLocal`.

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/util/TaskRun.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/util/TaskRun.scala
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2014-2019 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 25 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/internal/util/TaskRun.scala
```diff
@@ -0,0 +1,40 @@
...
+import monix.execution.schedulers.TracingScheduler
+
+private[reactive] object TaskRun {
+  /**
+    * Builds a [[monix.eval.Task.Options]] reference, for executing
+    * tasks in the context of `Observable`.
+    */
+  def options(implicit s: Scheduler): Task.Options =
+    s match {
+      case _: TracingScheduler => optionsWithLocalsRef
```
oleg-py: @alexandru we also have `TracingSchedulerService`, which doesn't extend `TracingScheduler` directly. Also not sure how well this would play with wrappers like `TrampolineScheduler`

alexandru: Indeed, I have to think more about it.

Right now I’m thinking of adding a `features` field to `Scheduler`.

Do you have any ideas?

oleg-py: Sounds good. I'd prefer anything that favors composition / polymorphism over casts :)

alexandru:
This isn't what I want to do. I have another branch I'm working on, will close this for now.

Issue #800, opened 2018-12-22
Add convenience method ObservableLike.filterNot


Otah:
Similarly as `TraversableLike` or ReactiveX's `Observable`, it would be very handy to have the `filterNot` method, the "negation" of `filter`.

Or is there any particular reason why this wasn't included in `ObservableLike` (version 2.3.3)?

I can also provide a pull request, if this is accepted.

alexandru:
Feel free to add a PR.

PR #816, opened 2019-01-21
Extend ObservableLike with filterNot method (#800)

Otah:
Introducing `ObservableLike.filterNot` operation for 2.x series.

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/observables/ObservableLike.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/observables/ObservableLike.scala
@@ -1019,6 +1019,16 @@ trait ObservableLike[+A, Self[+T] <: ObservableLike[T, Self]]
   def filter(p: A => Boolean): Self[A] =
     self.liftByOperator(new FilterOperator(p))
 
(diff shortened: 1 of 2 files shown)


alexandru:
All code needs some test. Can you add a test?

Also if we add `filterNot` in 2.3, we need it in 3.0 as well.

Otah:
Ah, yes, sorry for not adding the test. I will add it.

Regarding the 3.0, that’s my plan. However because in 3 series there is not only the simple `filter(A=>Boolean)` but also variants with `Task`, I haven’t yet had a chance to implement this for consistency.

Long story short, I will do both the tests and a separate PR for 3 series (to `master`, correct me if I am wrong). For now, I just need some time to do it.

Thanks!

alexandru:
I wouldn't worry about the `Task` variant in 3.0.

Otah:
Not sure why the build failed, seems like the Travis CI somehow timed out? The SBT build was passing locally.

I have opened a MR for series 3.x as well, see https://github.com/monix/monix/pull/819

Avasil:
Thank you @Otah :)

PR #822, opened 2019-02-01
Fix #815: remove dependency on cats-effect from monix-execution

### LINKED ISSUE #815: Extract Schedulers into a separate deployable module
Motivation:
Looks like Schedulers can be extracted to a separate dependency-free module, that can be used without Observables, Tasks or various Effects. Just as a light Scala-friendly replacement of Java's ScheduledExecutorService which also has control over clock! and thus has proper test implementation (TestScheduler).

`monix-execution` has dependency on `cats-effect` and `reactive-streams` which while doesn't disqualify it from using in "only Scheduler" use cases, but makes it much less attractable.
I believe this can benefit Scala eco-system in general

oleg-py:
Basically all the usages of cats-effect in execution are removed as discussed in #815.

For CancelableFuture instances, I've used a little known fact that package objects of parent packages are part of an implicit scope for type. That means if `monix-execution` does not define one, it can be defined in a separate project (monix-catnap defined one for `monix.execution`). Instances are immediately available without any extra imports, if you have catnap dependency.

This is an obscure hack, but it gives the best UX as in no extra imports are required. OTOH, it can be used to provide extension methods too. I've refrained from pushing it too hard for now.

I decided also to provide factories for Scheduler - cats-effect datatypes (`SchedulerEffect`, suggestion for better name are welcome) without syntax extensions - syntax doesn't really buy you much.

In #815, I forgot to mention Cancelable#toCancelToken. It has an optimization for dummy cancelables. We use it three times where it seems like it would not hit that optimization - scheduler calls and CancelableFuture usage.


monix-catnap/shared/src/main/scala/monix/execution/package.scala
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package monix
+
+import cats.Contravariant
+
+import scala.concurrent.ExecutionContext
+
+package object execution {
```
oleg-py: This is the hacky bit. It might seem innocent, but note that it's actually in the catnap subproject.

alexandru: This is an awesome hack if it allows us to keep the instances comfortable while removing the Cats-Effect dependency from `monix-execution`.

PR #826, opened 2019-02-02
Fix #712: add Iterant.uncons operation

### LINKED ISSUE #712: Iterant.headOptionL and Iterant.tail at once
It seems useful to have a function which would do ```Iterant.headOptionL``` and ```Iterant.tail``` at once for (head, rest)-iteration (yes, there is ```fold```-family of functions, but sometimes is it useful to perform "just one step", e.g. when merging many streams looking at their head's content)

Something like
```
  final def deconsL(implicit F: Sync[F]): F[(Option[A], Iterant[F, A])] =
    F.map(this.headOptionL) {
      case Some(hd) => (Some(hd), this.tail)
      case None     => (None, Iterant.empty)
    }
```

But this naive code won't work well: separate ```headOptionL``` and ```tail``` produce different "forks" of the ```Iterant```, evaluating ```Task```s inside the ```Iterant``` twice (besides the performance hit, it might result in double resource acquisition or in garbage output if the Task code is not deterministic)

oleg-py:
* Both Resource and Iterant versions (latter could be optimized, but now it's just convenience to avoid `fromResource`.
* Single element only. Batch versions can be made later.

@alexandru should I add separate tests to check how resource alloc works, or laws are enough on our setup?

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
@@ -2061,9 +2061,41 @@ sealed abstract class Iterant[F[_], A] extends Product with Serializable {
   final def toListL(implicit F: Sync[F]): F[List[A]] =
     IterantFoldLeftL.toListL(self)(F)
 
+  /**
+    * Pull the first element out of this Iterant and return it and the rest.
+    * If the returned Option is None, the remainder is always empty.
+    *
+    * The value returned is wrapped in Resource to preserve resource safety,
+    * and consumption of the rest must not leak outside of use.
+    *
+    * This can be used to implement custom operators. For example, `fold` is
+    * implementable by recursing into a Resource:
+    *
+    * {{{
+    *   def fold[F[_]: Sync, A: Monoid](iterant: Iterant[F, A]): F[A] = {
+    *     def go(iterant: Iterant[F, A], acc: A): Resource[F, A] =
+    *       iterant.unconsR.flatMap {
+    *         case (None, _) => Resource.pure(acc)
+    *         case (Some(a), rest) => go(rest, acc |+| a)
+    *       }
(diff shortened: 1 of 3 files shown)


alexandru:
@oleg-py this is very cool 🎉

What I'm wondering though is how useful it is, given that usage (of `Resource` at least) implies evaluation via `bracket`, which is not tail recursive, so I'm wondering if it's feasible to process an entire stream (that may be really big or infinite) via `unconsR` without memory blowing up.

How does it behave if it's a pure (no resources to close), but infinite stream. Can you do a quick test for yourself?

Otherwise `Resource` usage is fine, it communicates the result better, I don't think we need the `Iterant` version.

oleg-py:
@alexandru you are correct about memory usage. Seems that on every `.pure` of Resource, a lambda is allocated and is never released, so it is not suited for processing infinite streams. I rewrote it to use Iterant instead - with Iterant, all allocated nodes are eventually getting de-allocated (except for degenerate case of infinitely nested Scopes).

I'm very keen on preserving the Iterant version. I assume it's going to be used to implement custom operators, e.g. grouping in pairs:

```scala
def pairs(i: Iterant[F, A]): Iterant[F, (A, A)] = i.uncons.flatMap { case (opt, tail) =>
  tail.uncons.flatMap { case (opt2, tail2) =>
    (opt1, opt2).tupled.fold(Iterant.empty[F, A])(_ +: pairs(tail2))
  }
}
```

where you want to produce `Iterant` back as a result eventually. Resource version is better for _folds_, but Iterant version is better for transforms.

I've removed Resource version now. Can re-add it with a note that it's not safe for recursing on infinite streams.

alexandru:
Thanks for checking. The `Iterant` version is great if it works for pure and infinite streams.

Lets not add the `Resource` version for now.

PR #828, opened 2019-02-03
WIP: Add method to override UncaughtExceptionReporter on Scheduler

### LINKED ISSUE #799: API for overriding Scheduler.reportException?
On JS, I wanted to override `reportException` in a Scheduler and ended up in a bad situation:

* `AsyncScheduler` and `TrampolineScheduler` are final, so I cannot inherit from any of them.
* There are no utilities like `withExecutionModel` for creating one.
* There are no factory methods on Scheduler that take an `UncaughtExceptionReporter` for JS (there are for JVM).

I think we can port a number of JVM-only constructors to JS, such as this one:
```scala
def apply(reporter: UncaughtExceptionReporter, executionModel: ExecModel): Scheduler
```
and those taking ExecutionContext (since we have ECs in SJS, and schedulers are ECs themselves)

I'd also appreciate a method like `withExceptionReporter` too.

oleg-py:
Fixes #799

### DIFF
--- a/monix-execution/js/src/main/scala/monix/execution/schedulers/AsyncScheduler.scala
+++ b/monix-execution/js/src/main/scala/monix/execution/schedulers/AsyncScheduler.scala
@@ -65,6 +65,19 @@ object AsyncScheduler {
     *        [[monix.execution.ExecutionModel ExecutionModel]], a guideline
     *        for run-loops and producers of data.
     */
+  def apply(context: ExecutionContext, executionModel: ExecModel): AsyncScheduler =
+    new AsyncScheduler(context, executionModel, UncaughtExceptionReporter(context.reportFailure))
+
+  /** Builder for [[AsyncScheduler]].
+    *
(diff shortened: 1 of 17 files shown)


alexandru:
Thanks @oleg-py for taking the lead on this.

Initially I wanted to add a `withExceptionReporter`, but unfortunately it's not simple to do it, because in various implementations that exception reporter is also used by the thread pool implementation, not just by us and replacing it with your own instance for an already initialized thread pool isn't possible or I can't see how atm.

But for now this will do.

oleg-py:
@alexandru yes, lack of tests is why it's still WIP. Will do when I get a next chunk of time for OSS.

oleg-py:
I discovered we have a problem which is kind of a blocker for this: #829

oleg-py:
I think it'd be easier for me to solve the #829 together with #799 from scratch then to salvage this branch.

PR #838, opened 2019-03-06
[Suggestion] Add taskified variants of timeout combinators

tkroman:
I propose adding `timeoutL`/`timeoutLTo` variants of `timeout`/`timeoutTo`. Motivating examples are in the code. If you think this deserves to be in the library, I'll add tests etc.


oleg-py:
I'm 👍 on adding these, given we have Ref and TaskLocal stuff, these can be useful.

tkroman:
Cool. Any suggestions on names? I admit that `timeoutLTo` is far from being discoverable except by signature.

tkroman:
Another question - since timeout is a task now, should I try to recover somehow? If a timeout task fails, then the whole task fails. Not sure if this is the desired behaviour.

oleg-py:
As far as naming goes, I'd suggest keeping L as strictly a suffix. I.e. `timeoutToL`.

The issue I'm currently seeing is that currently you're waiting for a duration to compute, and then also for that duration. Maybe you should subtract the time it took to compute the duration from actual delay?

I don't think sensible recovery is possible. You can't just ignore the delay, as that might cause resource leak. You can return the result if it managed to compute before the delay failed, but I think that should be already done by race (although a test wouldn't hurt),

tkroman:
Good catch on the evaluation time. I changed the code to account for it but now is seems like some users would want to actually ignore that time. Should I put it under a flag? 

```scala
def timeoutL(timeout: Task[Duration], respectTimeoutTimings: Boolean = true) = ???
```

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -2210,21 +2209,121 @@ sealed abstract class Task[+A] extends Serializable {
   /** Returns a Task that mirrors the source Task but that triggers a
     * `TimeoutException` in case the given duration passes without the
     * task emitting any item.
+    *
+    * Useful when timeout is variable, e.g. when task is running in a loop
+    * with deadline semantics.
+    *
+    * Example:
+    * {{{
+    *   import cats.effect.IO
```
Avasil: reduntant import I think

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -2210,21 +2209,121 @@ sealed abstract class Task[+A] extends Serializable {
...
-    */
-  final def timeoutTo[B >: A](after: FiniteDuration, backup: Task[B]): Task[B] =
-    Task.race(this, Task.unit.delayExecution(after)).flatMap {
+    *
+    * Useful when timeout is variable, e.g. when task is running in a loop
+    * with deadline semantics.
+    *
+    * Example:
+    * {{{
+    *   import cats.effect.IO
```
Avasil:
```suggestion
```

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -2210,21 +2209,121 @@ sealed abstract class Task[+A] extends Serializable {
...
+    *                              E.g. if it took 3 seconds to evaluate `after`
+    *                              to a value of `5 seconds`, then,
+    *                              if the value of this parameter is `true`,
+    *                              this task will timeout in exactly 5 seconds
+    *                              from the momemnt computation started;
+    *                              and if the value of this parameter is `false`,
+    *                              this will timeout in 5 seconds _after_
+    *                              the evaluation of timeout task was completed.
+    **/
+  final def timeoutToL[B >: A](after: Task[FiniteDuration],
```
Avasil: Could you please change formatting like this:

```scala
final def timeoutToL[B >: A](
  after: Task[FiniteDuration],
  backup: Task[B],
  respectTimeoutTimings: Boolean = true): Task[B] = 
```

monix-eval/shared/src/test/scala/monix/eval/TaskRaceSuite.scala
```diff
@@ -251,6 +251,79 @@ object TaskRaceSuite extends BaseTestSuite {
...
+    assertEquals(f.value, None)
+
+    s.tick(1.seconds)
+    assert(f.value.isDefined)
+    assert(f.value.get.isFailure)
+    assert(f.value.get.failed.get.isInstanceOf[TimeoutException])
+  }
+
+  test("Task#timeoutL considers time taken to evaluate the duration task") { implicit s =>
+    // "3 seconds" after delay of 2s
```
Avasil: I don't think we need the comments

Avasil:
> Good catch on the evaluation time. I changed the code to account for it but now is seems like some users would want to actually ignore that time. Should I put it under a flag?
> 
> ```scala
> def timeoutL(timeout: Task[Duration], respectTimeoutTimings: Boolean = true) = ???
> ```

In other places we use different methods and call it `fixedRate` and `fixedDelay` but I think in this case flag is reasonable. WDYT @oleg-py ?

oleg-py:
We have `task.timed`. If we have one of these versions, the other one is always constructible via `timed` + `flatMap` + either subtraction or addition, depending on variant we have.

So, I think having one version is good enough (and I'm not a huge fan of boolean flags tbh). We don't need to address every use case with one method, if we have enough reasonably convenient composable pieces to let user solve the problem.

Which one should be default is the next question. I have a vague feeling that subtracting elapsed time would be less surprising, but I'm not insisting on it.

tkroman:
@Avasil @oleg-py addressed comments & merged. I also ran scalafmt on Task.scala which lead to a larger diff, should I revert?

Avasil:
@tkroman Revert scalafmt please, I think we were undecided whether to use it

tkroman:
Done. Feel free to merge :)

Issue #806, opened 2019-01-14
TaskLocal is not propagated when using bracket methods


oleg-py:
The following usage of `TaskLocal` seems to be broken with bracket-like methods:

```scala
import monix.eval._
import monix.execution.Scheduler.Implicits.global

val taskL = for {
  local <- TaskLocal(0)
  _ <- Task.unit.executeAsync.guarantee(local.write(1))
 // alternative:
//   _ <- Task.unit.guarantee(local.write(1).executeAsync)
  value <- local.read
} yield value

taskL
  .executeWithOptions(_.enableLocalContextPropagation)
  .runSyncUnsafe() // results in 0
```

@alexandru the code in `TaskBracket` seems to directly say to NOT restore locals. Not sure why should it be the case.

alexandru:
I don't remember what the issue was TBH and why it made sense.

Will take another look.

Feel free to investigate yourself.

alexandru:
@oleg-py does your latest PR (#807) fix this?

oleg-py:
@alexandru nope. It addresses a separate issue. Didn't yet have a chance to look into that one

Avasil:
@alexandru @oleg-py 

Actually, it seems like if we don't restore locals in `TaskShift` (set `restoreLocals = false` ) then it works and all the tests pass. I guess it's either the solution or we need more tests

The new local value is discarded here: 
```scala
override def prepareCallback: Callback[Throwable, Any] =
  new Callback[Throwable, Any] {
    def onSuccess(value: Any): Unit = {
      val locals = previousLocals
      if (locals ne null) Local.setContext(locals)
      callback.onSuccess(value)
     }
```

https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/internal/TaskRestartCallback.scala#L166

It was changed here: https://github.com/monix/monix/pull/647

Do you remember any context regarding this? I have been investigating this recently but I don't quite grasp Task's internals and can't really reason about potential repercusions. :( Though this case seems to be similar to `executeOn` explained in the PR.

mdedetrich:
Is there a conclusion to this problem? We are trying to use `TaskLocal` and this bit us before?

oleg-py:
@Avasil if the tests pass and it works, I'd say go for PR (since @alexandru seems to be busy for now).

mdedetrich:
PR is open, although it may make a sense to wait for a few days to see if @alexandru responds with the original reasoning for what was done.

[...]

PR #839, opened 2019-03-09
TaskLocal should propagate when used with Bracket Methods

Avasil:
Related to #806

Making PR as suggested by @oleg-py 

The difference between
```scala
Task.unit
    .executeAsync
    .guarantee(local.write(10))
```

and

```scala
Task.unit
    .guarantee(local.write(10))
```

is that the former starts `TaskRestartCallback` twice (the first one is just after evaluating `guarantee` body) and the latter just once. 

@alexandru If you could provide some explanation of `TaskRestartCallback` that would help immensely in putting all the pieces together. :) I feel like I'm getting somewhere with my understanding but there are a lot of guesses at this point


Avasil:
I'm merging then so we can move forward, we will revert it if it turns out to be wrong

Issue #841, opened 2019-03-12
MapParallelOrdered is not running all Tasks in parallel


lukestephenson:
This was originally mentioned to me by @cwmyers who I work with.  I've since looked into it a little more and thought about how it could be fixed.

# Problem / Issue description

With the following simple Observable using `mapParallelUnordered`, the process takes just over a second to complete as the Tasks are executed in parallel.
```
package monix

import monix.eval.Task
import monix.execution.Scheduler
import monix.reactive.Observable

import scala.concurrent.Await
import scala.concurrent.duration.Duration

object Example2 {
  implicit val scheduler = Scheduler.fixedPool("MyTest", 20)

  def createStream = {
    val last =
    Observable.range(1, 5)
        .mapParallelUnordered(20)(waitForABit)
        .dump("Int = ")
        .runAsyncGetLast
    val maybeInt = Await.result(last, Duration.Inf)
    maybeInt
  }

  def waitForABit(i: Long): Task[Long] = {
    Task {
      Thread.sleep(1000)
      println(Thread.currentThread().getName + s" - $i")
      i
    }
  }

  def main(args: Array[String]): Unit = {
    createStream
  }
}
```

Note the ordering is not guaranteed (as expected).
```
MyTest-14 - 4
MyTest-11 - 1
MyTest-12 - 2
MyTest-13 - 3
0: Int =  --> 3
1: Int =  --> 4
2: Int =  --> 2
3: Int =  --> 1
4: Int =  completed
```

Now if the above program is changed from `mapParallelUnordered` to `mapParallelOrdered` the behaviour changes significantly (beyond just the ordered output):

1. Everything occurs on the `main` thread.
```
main - 1
0: Int =  --> 1
main - 2
1: Int =  --> 2
main - 3
2: Int =  --> 3
main - 4
3: Int =  --> 4
4: Int =  completed
```
2. Execution is sequential
This program takes just over 4 seconds to complete.

While the use of `Thread.sleep` isn't ideal and replacing with a `Task.delayExecution` would fix the behaviour in this example, I don't think the current implementation of `mapParallelOrdered` is ideal.

I think users of the Monix library aren't expecting a radically different result when switching between `mapParallelOrdered` and `mapParallelUnordered`

Also, if the async Task was CPU bound, the implementation would also fail to achieve parallelism.

As such I'm proposing the current implementation is changed.

# Potential issue
Currently the elements are `process`ed when the parallelism semaphore permit is acquired.
[...]

PR #845, opened 2019-03-12
Ensure MapParallelOrdered runs tasks in parallel

lukestephenson:
See #841

Note the cancellation tests are not working and I still need
to investigate this.


Avasil:
I'm closing this since we decided to go with the other approach

PR #849, opened 2019-03-24
Specify exception on timeout

mdedetrich:
This PR allows you specify an exception when calling `.timout` on a `Task`. The reason why this PR is being made is that in one of our company projects using Monix we make heavy use of `.timeout` as well as recovering after `.timeout`. Due to needing to know if a Monix `Task` raised an exception due to a `timeout`, we had to create this ugly workaround 

```scala
private def isThrowableMonixTimeout(throwable: Throwable): Boolean =
  throwable match {
    // This is the exception that is thrown when you do .timeout on a Task
    case t: TimeoutException if t.getMessage.contains("Task timed-out after") => true
    case _                                                                    => false
  }
```

This simple PR allows you to specify your own type of `TimeoutException`

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -2212,7 +2212,16 @@ sealed abstract class Task[+A] extends Serializable {
     * task emitting any item.
     */
   final def timeout(after: FiniteDuration): Task[A] =
-    timeoutTo(after, raiseError(new TimeoutException(s"Task timed-out after $after of inactivity")))
+    timeoutWith(after, new TimeoutException(s"Task timed-out after $after of inactivity"))
+
+  /** Returns a Task that mirrors the source Task but that triggers a
+    * specified `Exception` in case the given duration passes
+    * without the task emitting any item.
+    * @param exception The `Exception` to throw after given duration
+    *                  passes
+    */
+  final def timeoutWith(after: FiniteDuration, exception: Exception): Task[A] =
+    timeoutTo(after, raiseError(exception))
 
   /** Returns a Task that mirrors the source Task but switches to the
     * given backup Task in case the given duration passes without the


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -2212,7 +2212,16 @@ sealed abstract class Task[+A] extends Serializable {
...
-    timeoutTo(after, raiseError(new TimeoutException(s"Task timed-out after $after of inactivity")))
+    timeout(after, new TimeoutException(s"Task timed-out after $after of inactivity"))
+
+  /** Returns a Task that mirrors the source Task but that triggers a
+    * specified `TimeoutException` in case the given duration passes
+    * without the task emitting any item.
+    * @param timeoutException The `TimeoutException` to throw after
+    *                         given duration passes
+    */
+  final def timeout(after: FiniteDuration, timeoutException: TimeoutException): Task[A] =
```
oleg-py: It feels overly restrictive to limit this to TimeoutException. Why not just `Throwable`?

oleg-py: Also, not sure if we should use overloading or have a different name like `timeoutWith`. Overloaded methods usually cooperate less with automatic eta expansion.

mdedetrich: I was actually thinking about making it `Throwable` in general, but I thought that if you are going to make your own error type which indicates a timeout exception you should make it extend `TimeoutException. I can change this though, don't feel to strongly about it

> Also, not sure if we should use overloading or have a different name like timeoutWith. 

I can avoid overloading by calling the method `timeoutWith` if desired, does Monix in general not do overloading (I personally prefer the overloaded method but I can change it if there is a precedent).

> Overloaded methods usually cooperate less with automatic eta expansion

Do you have any examples, or an article/issue that describes such issues?

oleg-py: > you should make it extend TimeoutException 

I would agree there if TimeoutException was a `trait` you could retroactively add with little effort, but it's a class, so it might not be that easy if you have a custom exception hierarchy.

> does Monix in general not do overloading?

I have seen _very_ few overloaded methods there. `Iterant.++` and `Iterant.suspend` are all I can remember, even. Default parameters occur more often, but it's still happens in a very niche cases, like `Iterant/Observable.dump` lets you override `PrintStream` or `task.executeOn` lets you switch off some of the added async boundaries.

> Do you have any examples, or an article/issue that describes such this in more detail?

Not at hand, unfortunately. I just vaguely remember running into issues here and there, but it's more of compiler edge cases than a general rule.

mdedetrich: Okay I will do the changes then.

Issue #848, opened 2019-03-22
TracingScheduler issues when mixing Future with Task


mdedetrich:
The equivalent issue can be found here https://github.com/mdedetrich/monix-mdc/issues/1. For basic scenarios when mixing `Task` and `Future` along with a custom `MDCAdaptor` that uses `TaskLocal`, everything seems to work fine (i.e. you can use `MDC.get`/`MDC.put` interchangeably and stuff works). However some more complex test scenarios don't seem to work, i.e. these tests currently fail 

https://github.com/mdedetrich/monix-mdc/blob/master/src/test/scala/MDCFutureSpec.scala#L56-L106

Where as the equivalent ones (that just use `Task`) work fine, i.e.

https://github.com/mdedetrich/monix-mdc/blob/master/src/test/scala/MDCBasicSpec.scala#L54-L66

The easiest way to replicate this is to just clone https://github.com/mdedetrich/monix-mdc/ and run test. Note that the repo uses `3.0.0-RC2-SNAPSHOT-9e79718-SNAPSHOT` which contains some `TaskLocal` fixes.

Its also entirely possible that this is an issue with the way that `MonixMDCAdapter` is written

mdedetrich:
So just an update on this, I redid the tests in monix-mdc (I know use property based tests to make sure the state is clean on each test run) and figured out that the issue occurs when you put a value into `TaskLocal` initially when using a `Future` and then try to modify the `TaskLocal` later within a `Task`.

If you put a value into the `TaskLocal` initially using `Task`, then you can retrieve the values (correctly) when using a `Future` afterwards.

oleg-py:
Our main problem currently is that this test would fail, if added:

```scala
  testAsync("multiple futures operation") {
    implicit val traced = Scheduler.traced
    val local = Local(42)
    for {
      f1 <- Future { local := 11 }
      f2 <- Future { local.get }
    } yield assertEquals(f2, 11)
  }
```

It seems that, for simple case like this:
```scala
// slightly desugared version of Future(X).flatMap(Y)
val x = Future { X } (ec)
val y = x.flatMap(Y)(ec)
```
Both X and Y would receive a copy of a current/parent/callsite state. Therefore, any modifications done by X will not be seen when it's time for Y to be called. We need to somehow "link" the changes across two call sites, or, stated in a different way, share the context state between the two, like a reference to a mutable state.

mdedetrich:
[...]

Issue #851, opened 2019-03-25
TaskLocal should carry over values on `join`


oleg-py:
When fiber is joined, the point at which we resume should contain the locals as they were upon fiber completion, i.e.:

```scala
  testAsync("TaskLocal should `pull` locals on join") {
    val t = for {
      local <- TaskLocal(0)
      _     <- local.write(10).start.flatMap(_.join)
      _     <- local.read.map(assertEquals(_, 10))
    } yield ()

    t.runToFutureOpt
  }
```

Currently fibers get a copy of the parent local context, but all values written to it are lost. That also means that quarantine function from #837 can be implemented as starting, then joining.

This breaks the following law from Concurrent typeclass:
```scala
  def startJoinIsIdentity[A](fa: F[A]) =
     F.start(fa).flatMap(_.join) <-> fa
```

mdedetrich:
Could this be related to https://github.com/monix/monix/issues/848#issuecomment-476186841 ?

oleg-py:
@mdedetrich unlikely. I think we need to manually carry over locals in this special case. #848 is more about TracingScheduler and Future.

oleg-py:
Fixed in #887

PR #852, opened 2019-03-25
Make Task.deferFuture not reset locals

oleg-py:
This is part of a fix for #848. We reset locals on boundary when going from Future, and it seems we shouldn't.

Used same approach as in #839.


Avasil:
Looks like `task.start keeps current Local.Context on join` in `TaskStartSuite` fails for Scala.JS and JVM `2.11` on CI

oleg-py:
@Avasil yeah, I can't see why though. It's as if #851 exists on my machines, but not on CI.

Has it failed before on CI? If you've got time, can you checkout my branch and see if it works for you?

Avasil:
@oleg-py It works for me locally

oleg-py:
I'll close this for now, we should fix it as we change the local propagation model

PR #861, opened 2019-04-07
Add fromTryL

tkroman:
Useful to easily taskify existing synchronous APIs.
Also adds tests for `fromTry`


Avasil:
Thank you for another contribution @tkroman 

Though I'm not sure if we need lazy variant. Here is an explanation why it's strict: https://github.com/monix/monix/issues/688#issuecomment-406955703

WDYT @alexandru @oleg-py ?

tkroman:
Oh yeah, `suspend` would work and doesn't need an additional method. Better to have fewer orthogonal methods than a bunch of ad-hoc one, feel free to close at will.

oleg-py:
It's actually much shorter in existing methods: `Task { functionReturningTry() }.dematerialize`

Avasil:
I'm closing then @tkroman but any other ideas about improving API are of course very welcome. :)

Issue #862, opened 2019-04-08
Support Scala 2.13.0


Avasil:
`Scala 2.13-RC1` has been released with many improvements to Scala's Future. Would be interesting to see how it affects `Observable` which uses `Future` internally.

Currently blocked by following releases:
- minitest
- cats
- cats-effect

`kind-projector` will need to be changed to
```scala
"org.typelevel" % "kind-projector" % "0.10.0"
```

Avasil:
New version of monitest: `2.4.0`
So just waiting for cats/cats-effect

SethTisue:
cats 2.0.0-M1 and cats-effect 2.0.0-M1 have been published for Scala 2.13.0-RC1

Avasil:
Thanks @SethTisue - I made a PR #875 
Not sure how fast we can release proper version but we could definitely do at least a snapshot soon

ALPSMAC:
Scala 2.13 has been released.  Would love to have a corresponding Monix 3.0 official release too :-)

Avasil:
@ALPSMAC We're going to release for Scala 2.13 as soon as stable Cats-Effect version supporting this release is available (I think it is planned in the next 2-3 weeks)

ALPSMAC:
Ah - perfect - thanks!

Daenyth:
@Avasil Could a milestone build be put out until then, for people to start doing WIP upgrades?

Avasil:
@Daenyth Sure, I will release something in a couple of hours

Daenyth:
Much appreciated!

Avasil:
@Daenyth 
The release is in progress but `2.13` should be already available
`3.0.0-eddc79d-SNAPSHOT`

The version is the latest master + dependencies supporting 2.13

Daenyth:
What resolver do I need for that?

Avasil:
@Daenyth I think it's `"Sonatype OSS Snapshots" at "http://oss.sonatype.org/content/repositories/snapshots/"`

alexandru:
`3.0.0` is now released with support for Scala 2.13

PR #863, opened 2019-04-08
Upgrade kind-projector to 0.10

Avasil:


Avasil:
Oh, it will have to be done with others: https://github.com/monix/monix/issues/862

PR #865, opened 2019-04-18
Adding FunctionK values for Task, Coeval

alexandru:
I'm using `Coeval` lately, for describing resources like `Resource[Coeval, ?]` and then I'm converting those to `Task` when needed. This is because you may want to initialize a resource in a dirty context, like Play Framework's dirty dependency injection and lifecycle mechanism and working with `Coeval` there is easier and more truthful.

But then doing a `mapK` to convert from `Resource[Coeval, ?]` to `Resource[Task, ?]` is freaking painful. Therefore we now have `liftTo` and `liftFrom` to help:

```scala
val res: Resource[Coeval, Something] = ???

res.mapK(Coeval.liftTo[Task]).use { r =>
  Task(???)
}

//------

val res: Resource[SyncIO, Something] = ???

res.mapK(Coeval.liftFrom[SyncIO]).use { r =>
  Task(???)
}
```

Summary of changes:

1. added `CoevalLift`
2. changed `TaskLift`, `TaskLike`, `ObservableLike` and `FutureLift` to extend `FunctionK[F, G]` (colloquially `F ~> G`)
3. added `liftTo`, `liftFrom` to `Coeval` and `Task`, also added `liftFrom` to `Observable` for consistency
4. deprecated `toIO` / `fromIO` and `fromEval` across the board
5. renamed `liftMap` from `Iterant` to `mapK`, for consistency with `Resource` in Cats-Effect

## 1. CoevalLift

This mirrors `TaskLift` and `LiftIO` from `cats.effect` and is used for lifting `Coeval` values into something else:

```scala
trait CoevalLift[F[_]] extends (Coeval ~> F)
``` 

Thus we can now do:

```scala
Coeval(1 + 1).to[SyncIO]

Coeval(1 + 1).to[Task]
```

## 2. TaskLift, TaskLike, ObservableLike

I changed them across the board to extend `FunctionK` from Cats:

```scala
trait CoevalLift[F[_]] extends (Coeval ~> F)

trait CoevalLike[F[_]] extends (F ~> Coeval)

trait TaskLift[F[_]] extends (Task ~> F)

trait TaskLike[F[_]] extends (F ~> Task)

trait ObservableLike[F[_]] extends (F ~> Observable)

trait FutureLift[F[_], Future[_]] extends (Lambda[A => F[Future[A]]] ~> F)
```

## 3. Added liftTo / liftFrom and variants

For `Coeval`:

```scala
object Coeval {

  def liftTo[F[_]](implicit F: CoevalLift[F]): Coeval ~> F = F

  def liftToSync[F[_]](implicit F: Sync[F]): Coeval ~> F

  def liftFrom[F[_]](implicit F: CoevalLike[F]): F ~> Coeval

}
```

For `Task`:

```scala
object Task {

  def liftTo[F[_]](implicit F: TaskLift[F]): (Task ~> F) = F

  def liftToAsync[F[_]](implicit F: cats.effect.Async[F], eff: cats.effect.Effect[Task]): (Task ~> F)

  def liftToConcurrent[F[_]](
    implicit F: cats.effect.Concurrent[F],
    eff: cats.effect.ConcurrentEffect[Task]): (Task ~> F)

  def liftFrom[F[_]](implicit F: TaskLike[F]): (F ~> Task)

  def liftFromEffect[F[_]](implicit F: Effect[F]): (F ~> Task)

  def liftFromConcurrentEffect[F[_]](implicit F: ConcurrentEffect[F]): (F ~> Task)

}
```

For `Observable`:

```scala
object Observable {

  def liftFrom[F[_]](implicit F: ObservableLike[F]): F ~> Observable

}
```

## 3. Deprecated .toIO, fromIO, fromEval across the board

These functions are not special, `IO` will soon no longer live in the base Cats-Effect package and `Eval` isn't so interesting as to have its own method.

Conversions for all of them are still supported via `Coeval.from` / `Task.from` / `Observable.from` and `coeval.to` / `task.to`.

## 5. Renamed `liftMap` from `Iterant` to `mapK`

Something that's bugging me a lot is naming consistency. Cats-Effect's `Resource` has `mapK`, but `Iterant` has `liftMap`. I don't like `mapK` that much, but consistency is important and I'd rather have naming consistency with `Resource`.

Old name is now a deprecated symbol.


monix-eval/shared/src/main/scala/monix/eval/CoevalLift.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import scala.annotation.implicitNotFound
+
+/**
+  * A lawless type class that specifies conversions from `Coeval`
+  * to similar data types (i.e. pure, synchronous).
+  */
+@implicitNotFound("""Cannot find implicit value for CoevalLift[${F}].
+Building this implicit value might depend on having an implicit
+s.c.ExecutionContext in scope, a Scheduler or some equivalent type.""")
```
oleg-py: Is there any type in the ecosystem that needs EC for a `Sync` instance? I'm afraid that might be more misleading than helpful.

monix-eval/shared/src/main/scala/monix/eval/CoevalLift.scala
```diff
@@ -0,0 +1,94 @@
...
+trait CoevalLift[F[_]] {
```
oleg-py: Proposal: 
```scala
trait CoevalLift[F[_]] extends Coeval ~> F // no body
```
Might cut out on the boilerplate in the impls and will show the intent in the code too

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -3957,6 +3958,48 @@ object Task extends TaskInstancesLevel1 {
...
     parMap6(fa1,fa2,fa3,fa4,fa5,fa6)((a1,a2,a3,a4,a5,a6) => (a1,a2,a3,a4,a5,a6))
 
+  /**
+    * Generates `cats.FunctionK` values for converting from `Task` to
+    * supporting types (for which we have a [[TaskLift]] instance).
+    *
+    * See [[https://typelevel.org/cats/datatypes/functionk.html]].
+    */
+  def toK[F[_]](implicit F: TaskLift[F]): FunctionK[Task, F] =
+    new FunctionK[Task, F] {
```
oleg-py: There's a macro method on FunctionK for this, IIRC `FunctionK.lift(F.taskLift)`
Also, kind-projector should let you write `λ[Task ~> F](F.taskLift)` too

oleg-py:
TBH, I'm not very big fan of lawless typeclasses, and wins over`res.mapK(FunctionK.lift(Task.coeval))` seem marginal, though I'd take `to[F[_]: Sync]` over `toIO`.

monix-eval/shared/src/main/scala/monix/eval/CoevalLift.scala
alexandru: No, just a left-over

monix-eval/shared/src/main/scala/monix/eval/CoevalLift.scala
alexandru: Sounds interesting 👍

alexandru:
@oleg-py I'm not a fan of lawless type classes either, but due to all the restrictions in Cats-Effect's `Bracket` and `Sync`, there are now many interesting types that are left out.

`Eval` and `Observable` are two such examples.

alexandru:
@oleg-py reliance on `FunctorK.lift` doesn't cut it, as you can easily end up with an error like this ...

```scala
missing parameter type for expanded function ((x$1: <error>) => x$1.task)
[error]       .mapK(FunctionK.lift(_.task))
```

The type inference does work with `FunctionK.lift(Task.coeval)`, but I don't think many people will know that. And there is no such function already described for `cats.effect.IO`.

Will take your suggestions under consideration. I especially like the idea of making the *Lift type classes inherit from `FunctorK`.

monix-catnap/shared/src/main/scala/monix/catnap/FutureLift.scala
```diff
@@ -59,8 +60,8 @@ import scala.concurrent.{Future => ScalaFuture}
   *   val sum: IO[Int] = IO(delayed(1 + 1)).futureLift
   * }}}
   */
-trait FutureLift[F[_], Future[_]] {
```
Avasil: A bit off-topic but it reminded me:

https://github.com/typelevel/cats-effect/pull/546

`Cats-Effect` changed `fromFuture` to do `shift`. Should we reconsider it on our part too?

alexandru: Yes, maybe.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -818,7 +817,8 @@ abstract class Observable[+A] extends Serializable { self =>
     *  - ...
     */
   final def bracketF[F[_], B](use: A => Observable[B])(release: A => F[Unit])(implicit F: TaskLike[F]): Observable[B] =
-    bracket(use)(release.andThen(F.toTask))
+    // TODO: pula
```
Avasil: leftover or unfinished business

alexandru: Oops 😂

monix-eval/shared/src/main/scala/monix/eval/CoevalLike.scala
```diff
@@ -119,7 +134,7 @@ private[eval] abstract class CoevalLikeImplicits0 {
     */
   implicit def fromComonad[F[_]](implicit F: Comonad[F]): CoevalLike[F] =
```
oleg-py: Actually, do we really want this instance? Apart from `Eval`, standard comonads from Cats are things like `NonEmptyList`, and a traditional case is something like a zipper, where `extract` would lose data - not that it makes sense to convert it to `Coeval` anyway.

monix-eval/shared/src/main/scala/monix/eval/TaskLift.scala
```diff
@@ -61,9 +63,22 @@ object TaskLift extends TaskLiftImplicits0 {
...
+        TaskConversions.toIO(task)(eff)
     }
+
+  /**
+    * Deprecated method, which happened on extending `FunctionK`.
+    */
+  implicit class Deprecated[F[_]](val inst: TaskLift[F]) {
+    /** DEPRECATED — switch to [[TaskLift.apply]]. */
+    @deprecated("Switch to TaskLift.apply", since = "3.0.0-RC3")
+    def taskLift[A](task: Task[A]): F[A] = {
```
oleg-py: Why not have this as a final method on TaskLift that just delegates to `apply`? Same for all the `Deprecated` classes

monix-eval/shared/src/main/scala/monix/eval/CoevalLike.scala
alexandru: OK, it probably doesn't make sense. Will just treat `Eval` as a special case and remove this.

monix-eval/shared/src/main/scala/monix/eval/TaskLift.scala
alexandru: Hmm, maybe, thinking about it.

Avasil:
@alexandru is it ready to merge? I could do the realese tomorrow evening :)

alexandru:
@Avasil yes, it's ready

PR #866, opened 2019-04-20
Allow sharing local context for Future interop

### LINKED ISSUE #837: TaskLocal and Local quarantine function
When calling third-party APIs you have no guarantee that it's not doing things that make you lose your Local instance, so I propose a function that does something similar to this:

```scala
  val globalLocal = Local[Int](0)
  def quarantine[A](task: => Task[A], scheduler: Scheduler): Task[A] = {
    val old = globalLocal.get
    task.asyncBoundary(scheduler).map { a =>
      globalLocal.update(old)
      a
    }
  }
```
This way we get to keep the local instance. 

I've written a `TaskLocalApp` that makes use of the above function using a `CompletableFuture` from java-8 which looses the `Local` instance.
https://gist.github.com/mihaisoloi/04f9ddbb16292fa3b6b3769d2744244e

oleg-py:
Attempt to fix #837 and #848, with `AtomicAny[Map[_, _]]` model.

As a side bonus, #851 is fixed automatically.

Needs more tests, `TaskLocal.isolate`, and I also would like to experiment on making `bind` not affect the reads of non-bound locals.

### DIFF
--- a/monix-eval/shared/src/test/scala/monix/eval/TaskStartSuite.scala
+++ b/monix-eval/shared/src/test/scala/monix/eval/TaskStartSuite.scala
@@ -52,7 +52,7 @@ object TaskStartSuite extends BaseTestSuite {
     assertEquals(f.value, Some(Success(1)))
   }
 
-  testAsync("task.start keeps current Local.Context on join") { _ =>
+  testAsync("task.start shares Local.Context with fibers") { _ =>
     import monix.execution.Scheduler.Implicits.global
     import cats.syntax.all._
     implicit val opts = Task.defaultOptions.enableLocalContextPropagation
@@ -60,14 +60,15 @@ object TaskStartSuite extends BaseTestSuite {
     val task = for {
       local <- TaskLocal(0)
       _ <- local.write(100)
-      f <- (Task.shift *> local.read <* local.write(200)).start
       v1 <- local.read
+      f <- (Task.shift *> local.read <* local.write(200)).start
+      // Here, before joining, reads are nondeterministic
       v2 <- f.join
       v3 <- local.read
     } yield (v1, v2, v3)
 
     for (v <- task.runToFutureOpt) yield {
-      assertEquals(v, (100, 100, 100))
+      assertEquals(v, (100, 100, 200))
     }
   }
 
--- a/monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
+++ b/monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
@@ -17,6 +17,9 @@
 
 package monix.execution.misc
 
+import monix.execution.atomic.AtomicAny
+
+import scala.annotation.tailrec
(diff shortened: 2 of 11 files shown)


monix-eval/shared/src/test/scala/monix/eval/TaskStartSuite.scala
```diff
@@ -52,22 +52,23 @@ object TaskStartSuite extends BaseTestSuite {
...
     import cats.syntax.all._
     implicit val opts = Task.defaultOptions.enableLocalContextPropagation
 
     val task = for {
       local <- TaskLocal(0)
       _ <- local.write(100)
-      f <- (Task.shift *> local.read <* local.write(200)).start
       v1 <- local.read
+      f <- (Task.shift *> local.read <* local.write(200)).start
+      // Here, before joining, reads are nondeterministic
```
Avasil: Would be possible to have isolation of locals after `start` until `join` happens?

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -210,24 +261,29 @@ final class Local[A](default: () => A) {
   /** Execute a block with a specific local value, restoring the
     * current state upon completion.
     */
+  @deprecated("Binds on Local prevent propagation of writes to other locals. Use Local.isolate directly", "3.0.0-RC3")
```
Avasil: I think there is also `bind` in `Local` companion object, should we deprecate/remove it too?

monix-eval/shared/src/test/scala/monix/eval/TaskStartSuite.scala
oleg-py: I don't think so. If both current and a forked fiber wrote to a Local, which one should win? That's the main question, and I don't think there's a good, generally applicable solution that is also not confusing

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
oleg-py: Binding on whole context is fine. The problem is that binding one variable implicitly binds whole context.

oleg-py:
We can merge it, but not release RC3 just yet. I'll make a PR on top of it to restore proper bind functionality, so we can see if we want to go with it or not.

Avasil:
Sounds good to me @oleg-py

mihaisoloi:
> We can merge it, but not release RC3 just yet. I'll make a PR on top of it to restore proper bind functionality, so we can see if we want to go with it or not.

Could we have an intermediary checksum release with this change though? This approach is much better than what I'm currently using

Avasil:
@mihaisoloi 
I've released snapshot version from this branch:
```
3.0.0-RC2-9c49e02-SNAPSHOT
```

Any feedback will be really valuable

Avasil:
Included in #887

PR #872, opened 2019-04-26
Add observable take while inclusive

### LINKED ISSUE #871: Add Observable.takeWhileInclusive
The current implementation of `takeWhile` means that the first item to fail the predicate is excluded from the resulting stream, in many cases this is fine, however there are cases where we might want to include the item that failed so that we can do further processing. The behavior itself is a minimal change to the `TakeByPredicate` operator but cannot be easily replicated with existing operators.

I realize that this may not mean much since `monix` has long since diverged from being another `Reactive Extensions` port, but the following are discussions that led up to the operators inclusion in rxjava and rxjs:

https://github.com/ReactiveX/rxjs/issues/2420
https://github.com/ReactiveX/RxJava/issues/1649

I'd be willing to running back a PR if there is an appetite for adding either an operator overload or a new operator which inclusively takes the last item (this could also include `dropWhile`).

allantl:
Resolve #871

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -1569,7 +1569,14 @@ abstract class Observable[+A] extends Serializable { self =>
     * predicate and returns a new observable that emits the rest.
     */
   final def dropWhile(p: A => Boolean): Observable[A] =
-    self.liftByOperator(new DropByPredicateOperator(p))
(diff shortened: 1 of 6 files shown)


oleg-py:
Can you also do it for a `dropWhile`?

allantl:
@oleg-py Sure, done!

oleg-py:
@allantl sorry for super slow responses. You need to add MiMa exclusions to `build.sbt` - it's fine to exclude anything in `internal` subpackage.

allantl:
@oleg-py No worries, I added the MiMa exclusions

PR #875, opened 2019-05-02
Scala 2.13.0 Cats-Effect 2.0.0-RC1

Avasil:
Related to #862

@alexandru @oleg-py We have to figure out whether our next release is on top of `cats-effect 2.0.0-M1` and/or `1.3.0`. I think there are some binary incompatibilities in `cats-effect` 2.0


oleg-py:
> I think there are some binary incompatibilities in cats-effect 2.0
IIRC cats-effect 2.0 is binary compatible with 1.x on Scala 2.12, same as cats.

3.0 final will be on cats-effect 2.0, but we can roll RC3 with fixed locals for CE 1.x if CE 2.0 isn't yet final by that time.

xuwei-k:
Any updates? 👀

Avasil:
@xuwei-k I don't know why CI fails but we will probably wait with release for stable Cats-Effect support if we use 2.0 for Monix 3.0 (or 3.1)

Would it be helpful to release a snapshot version from this branch?

niebloomj:
Any updates?

Avasil:
@niebloomj I have a snapshot from this branch if you're interested: `3.0.0-eddc79d-SNAPSHOT`

I'm not sure when stable `Cats-Effect` supporting `2.13` is released and I don't want to release on a milestone

project/MimaFilters.scala
```diff
@@ -2,6 +2,16 @@ import com.typesafe.tools.mima.core.ProblemFilters.exclude
 import com.typesafe.tools.mima.core._
 
 object MimaFilters {
+  lazy val changesFor_3_0_0__RC5: Seq[ProblemFilter] = Seq(
+    // Incompatible signatures should not cause linking problems.
```
Avasil: Should be fine but we might want to repeat this test: https://github.com/typelevel/cats-effect/pull/609#pullrequestreview-280142297

alexandru:
@Avasil see my PR: https://github.com/Avasil/monix/pull/4

alexandru:
The tests are now failing due to Mima, since we never published packages for 2.13.0.

alexandru:
We should only run Mima for the lowest version supported, in this case 2.11 and not for 2.13.

project/MimaFilters.scala
alexandru: Are we sure they don't cause linking problems?

What happened there?

alexandru: Oh, it's due to the changed `Parallel` from Cats or something?

alexandru: If so, then there's nothing we can do there.

PR #883, opened 2019-05-17
replace `case e if NonFatal(e)` with `case NonFatal(e)`

takayahilton:


Avasil:
Thanks @takayahilton 

Seeing how it is used a lot in the RunLoop - was there any specific reason @alexandru to do it like this?

oleg-py:
I can say that `case NonFatal(e)` causes an allocation of `Some(e)` in case error is actually non-fatal (which is most of errors happening inside Task), whereas calling `apply` that returns a boolean is not allocating anything. OTOH, in cats-effect IO run-loop it seems that using `case NonFatal(ex)` is fine.

alexandru:
`case e if NonFatal(e)` is indeed more efficient. In practice it doesn't really matter, but it does in benchmarks 🙂

alexandru:
@takayahilton let me ask you this ... any specific reason you don't like `case e if NonFatal(e)`?

I'm thinking that library code doesn't have to be idiomatic if we can avoid memory allocations and it's not like we're losing safety in this instance.

takayahilton:
@alexandru 
At first `case e if NonFatal(e)` seemed like a meaningless thing to me.
I understand that `NonFatal(e)` allocates `Some(e)`.
So I will close this PR.

Issue #882, opened 2019-05-15
Add Cron-based Observable


RomanIakovlev:
Cron is a popular mechanism to specify schedules, and Monix users will benefit from having it supported. I've implemented such an Observable (let's call it CronObservable going forward), and I can make open source, therefore I'm opening this issue to have a discussion on this topic. Particularly, let's talk about the following aspects.

1. Should CronObservable live in the Monix project itself, and be published in the main Monix artifact, or should it be in a separate artifact? My main concern here is that I've used a 3rd party library (https://github.com/alonsodomin/cron4s) to parse cron schedules and calculate the next execution date and time. Monix dependencies are rather minimal currently, and introducing a new one for a single tiny Observable might not be desireable.

At the same time, implementing required functionality (parsing cron expressions, calculating next execution time) from scratch seems like a waste of efforts, when an existing robust open source implementation exists. A way forward here might be to have a separate artifact with CronObservable (see next point for discussion).

2. If separate artifact for CronObservable is the best way to go, the question is, where the source code will live, and how it should be published. Source code can either live in Monix GitHub org, or in my personal one. Artifact can be published either under `io.monix` group id, or under my personal one as well. I'd personally like for this code to have Monix branding, but, if it's too much hassle for Monix maintainers for too little benefit, I can go ahead and do it all on my own, and then maybe add a link to this project from Monix documentation site.

Regarding the actual implementation, I've based CronObservable largely on the `monix.reactive.internal.builders.IntervalFixedRateObservable` class, because their functionality is quite similar. CronObservable emits incrementing integers on each cron tick.

Thanks for your time, and I would be glad to know your preference for moving this forward.

oleg-py:
I only have an opinion for 1st question now:

Adding a dependency to core monix is undesirable. If we can add some kind of time-interval compatible with cron4s, and make using both simple, it would be the preferred way to go. E.g. something along the lines of those:

```scala
def atGivenTimes(it: Seq[Instant]): Observable[Long]
def atGivenTimesUnfold[A](start: A)(next: A => Option[(Instant, A)]): Observable[Long]
```

Would totally fit in the monix-reactive and should make integration as simple as a one-liner, and would also fit some simpler use-cases that don't require an extra library.

RomanIakovlev:
[...]

PR #885, opened 2019-05-20
Add atGivenTimesUnfold Observable #882

RomanIakovlev:
This is an Observable suitable for implementing cron-like schedules,
where tick happens on each predefined instant (with back-pressure applied).

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -18,6 +18,7 @@
 package monix.reactive
 
 import java.io.{BufferedReader, InputStream, PrintStream, Reader}
+import java.time.Instant
 
 import cats.{Alternative, Applicative, Apply, CoflatMap, Eq, Eval, FlatMap, Functor, FunctorFilter, Monoid, NonEmptyParallel, Order, ~>}
 import cats.effect.{Bracket, Effect, ExitCase, IO, Resource}
@@ -4958,6 +4959,21 @@ object Observable extends ObservableDeprecatedBuilders {
     */
   def intervalAtFixedRate(period: FiniteDuration): Observable[Long] =
     new builders.IntervalFixedRateObservable(Duration.Zero, period)
+
+  /** Creates on Observable that emits auto-incremented natural numbers
+    * (longs) at a schedule defined by a `start` and `next` parameters.
+    * Particularly, `next` function is called continuously until it returns
+    * `None`, when the resulting Observable completes. Otherwise, its return
+    * value is used to determine next execution instant, as well as input to
+    * its own subsequent invocation.
+    *
+    * @param start input to the `next` function to produce the first tick instant
+    * @param next function emitting the next tick instants based on the result
+    *             of the previous invocation or the `start` parameter
+    * @tparam A type of the elements fed into the `next` function
+    */
+  def atGivenTimesUnfold[A](start: A)(next: A => Option[(A, Instant)]): Observable[Long] =
+    new builders.AtGivenTimesUnfoldObservable[A](start, next)
 
   /** Creates an Observable that emits auto-incremented natural numbers
     * (longs) at a fixed rate, as given by the specified `period`. The
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -18,7 +18,6 @@
 package monix.reactive
 
 import java.io.{BufferedReader, InputStream, PrintStream, Reader}
-import java.time.Instant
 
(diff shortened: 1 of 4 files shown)


oleg-py:
Good catch regarding Scala.js. I thought that `java.time` is included in SJS distribution, but apparently it's not.

Let's have a version in shared that takes `Long`s as a timestamp as a base implementation.

Let's also have a platform-specific version for `java.time.Instant` that is only available on JVM and when using Scala 2.12+, on a mixin to `object Observable`.

It has already been done for, e.g., CancelableFuture:
Platform-specific mixins:
https://github.com/monix/monix/blob/master/monix-execution/jvm/src/main/scala_2.11/monix/execution/internal/CancelableFutureForPlatform.scala
https://github.com/monix/monix/blob/master/monix-execution/js/src/main/scala/monix/execution/internal/CancelableFutureForPlatform.scala
https://github.com/monix/monix/blob/master/monix-execution/jvm/src/main/scala_2.12+/monix/execution/internal/CancelableFutureForPlatform.scala

mixin used here:
https://github.com/monix/monix/blob/master/monix-execution/shared/src/main/scala/monix/execution/CancelableFuture.scala#L180

If you feel adventurous, you can add support for java.util.Date (to shared, there's j.u.Date in SJS) and/or sjs native Date.

I also think a version that takes an Iterator or Seq, or something like that makes sense, as that would be more obvious and easy to try out when you aren't using cron4s.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -4959,6 +4960,21 @@ object Observable extends ObservableDeprecatedBuilders {
...
+    * `None`, when the resulting Observable completes. Otherwise, its return
+    * value is used to determine next execution instant, as well as input to
+    * its own subsequent invocation.
+    *
+    * @param start input to the `next` function to produce the first tick instant
+    * @param next function emitting the next tick instants based on the result
+    *             of the previous invocation or the `start` parameter
+    * @tparam A type of the elements fed into the `next` function
+    */
+  def atGivenTimesUnfold[A](start: A)(next: A => Option[(A, Instant)]): Observable[Long] =
```
oleg-py: Do you think it makes sense to emit `A` values instead? Could potentially be more general than emitting longs, and longs can be trivially recovered by doing things like `as.zipWithIndex.map(_._2)`.

RomanIakovlev: Agree, changed.

RomanIakovlev:
Ok, I've changed shared code to not depend on `java.time.Instant`. Let's see if tests pass now. I'll add shims for proper types on JS and JVM as a separate commit.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/AtGivenTimesUnfoldObservable.scala
```diff
@@ -0,0 +1,99 @@
...
+              // relationship between scheduleOnce invocations.
+              prevInstant = nextInstant
+              prevInput = nextInput
+              task := s.scheduleOnce(nextInstant - s.clockRealTime(MILLISECONDS), MILLISECONDS, self)
+            case None =>
+              o.onComplete()
+          }
+        } catch {
+          case NonFatal(ex) =>
+            try {
```
Avasil: I think this `try catch` is not necessary due to contract:

https://monix.io/docs/3x/reactive/observers.html#contract
(point 7)

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -4959,6 +4959,20 @@ object Observable extends ObservableDeprecatedBuilders {
...
+    * Observable to complete. Otherwise, its return value is used to
+    * determine next execution instant, as well as input to its own
+    * subsequent invocation.
+    *
+    * @param start input to the `next` function to produce the first tick instant
+    * @param next function emitting the next tick instants based on the result
+    *             of the previous invocation or the `start` parameter
+    */
+  def atGivenTimesUnfold[A](start: A)(next: A => Option[(A, Long)]): Observable[A] =
+    new builders.AtGivenTimesUnfoldObservable[A](start, next)
```
alexandru: This method is pretty cool. Does it do something like this?

```scala
Observable
  .fromAsyncStateAction(Option((start, 0))) { 
    case None => None
    case Some((a, instant)) =>
      Task.deferAction { s =>
        val now = s.clockRealTime(MILLISECONDS)
        val sleep = instant - now
        Task.now(next(a)).delayExecution(sleep.millis)
      }
  }
  .takeWhile(_.nonEmpty)
  .collect { case Some((a, _)) => a }
```

alexandru:
This was a cool suggestion, which we've let open, but I think it can be done with existing operators.

Closing it to clean up the issue tracker.

PR #897, opened 2019-06-12
WIP: configure CircleCI instead of Travis-ci

alexandru:
CircleCI might be better, I see improvements in speed and we might be able to do code coverage again (disabled because Travis-ci cannot handle it).


alexandru:
Too many challenges to configure a build matrix, giving up for now.

PR #898, opened 2019-06-12
Reformating via Scalafmt

alexandru:
This PR is reformatting all files via Scalafmt. I've also tuned it a little.

Found fragments which I did not like, however the result is totally worth it, plus we can fix as we go along.

N.B. we already had a Scalafmt config, however if people actually used it on a source file, it would then commit reformatting changes too, which is very annoying.

Also if we let the project be driven by Scalafmt, we can include a test for formatting in our CI build.


alexandru:
@Avasil the Scalafmt documentation is recommending against it, as it slows down the compilation process and it messes with the IDE. You'll get a lot of "an external change has occurred, reload file?" issues?

alexandru:
Alternative we could integrate a formatting check in `test`.

alexandru:
@Avasil I just added `scalafmtCheck` in testing, to be executed by the CI.

alexandru:
Not ideal to add more stuff to our build, we'll need to move to CircleCI.

alexandru:
Tests with `scalafmtCheck` pass 🙂 I'll assume you all agree 😛 I need to merge this in order to work relaxed on the other PR.

I think Scalafmt is doing a pretty good job for an automated tool, better then I expected, but as I said, we can fix it as we go along if we find fragments that aren't right.

xuwei-k:
https://github.com/monix/monix/pull/902

PR #901, opened 2019-06-14
avoid deprecated unicode arrow chars

xuwei-k:
https://github.com/scala/scala/commit/9e264483d04f48513f3a82e91365667f01a2f5f3


alexandru:
Cool. I think I had a phase like that before switching to the ligatures in Fira Code 🙂

Issue #856, opened 2019-03-31
Task.memoize doesn't play well with Local


guymers:
I am trying to use `Local` as a replacement for a `ThreadLocal` in order to implement OpenTracing support (https://github.com/opentracing/opentracing-java/blob/release-0.31.0/opentracing-util/src/main/java/io/opentracing/util/ThreadLocalScopeManager.java). However not all downstream `Task`s see the active `Scope`. Is what I am trying to accomplish a supported use case?

If I make this change everything works as I would expect.
```
diff --git a/monix-eval/shared/src/main/scala/monix/eval/internal/TaskRestartCallback.scala b/monix-eval/shared/src/main/scala/monix/eval/internal/TaskRestartCallback.scala
index 9752c396..6a91b3e0 100644
--- a/monix-eval/shared/src/main/scala/monix/eval/internal/TaskRestartCallback.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/internal/TaskRestartCallback.scala
@@ -155,8 +155,7 @@ private[internal] object TaskRestartCallback {
     private[this] var previousLocals: Local.Context = _
 
     override protected def prepareStart(task: Task.Async[_]): Unit = {
-      preparedLocals =
-        if (task.restoreLocals) Local.getContext() else null
+      preparedLocals = Local.getContext()
     }
 
     override def prepareCallback: Callback[Throwable, Any] =
```

The below test approximates what an average `Task` stack looks like:
```
val memoizedTask = Task.delay {
  5
}.memoize

val i = AtomicInt(0)

val t = for {
  local <- TaskLocal(0)
  ii <- Task.shift.flatMap(_ => Task.delay {
    i.incrementAndGet()
  })
  _ <- local.write(ii)
  result <- Task.parZip2(
    memoizedTask.flatMap { _ =>
      Task.sleep(100.millis).flatMap { _ =>
        local.read
      }
    },
    memoizedTask.flatMap { _ =>
      Task.sleep(100.millis).flatMap { _ =>
        local.read
      }
    },
  )
} yield {
  println(result)
}

t.runToFutureOpt
t.runToFutureOpt
t.runToFutureOpt
```

Running this will never return `(0,0)`, `(1,1)`, `(2,2)`.

However if you remove the `Task.shift` or remove the `memoize` it does.

Avasil:
We'd like to support such a use cases if it doesn't currently work as expected

What result do you get instead? I get:
```
(1,1)
(2,2)
(3,3)
```

[...]

PR #905, opened 2019-06-17
[POC] Remove restoreLocals flag

Avasil:
I don't have an intention to merge this PR, it's more to start a discussion because I feel like we can get rid of `TaskRestartCallback#WithLocals` because most of the heavy lifting is done by `Local` after @oleg-py changes.

Only two tests started failing - `TaskLocals get restored in Task.create on error` and `locals get transported with executeOn and shift` - I had to use `TracingRunnable` in cases with manual execution to the `ExecutionContext`. It might be cool to not require `TracingRunnable` so we probably can't just remove everything like in this PR but it seems to be possible to cut 90% of TaskLocal special cases

`restoreLocals` flag was introduced in https://github.com/monix/monix/pull/647

We should keep in mind https://github.com/monix/monix/issues/856 which is about issues with `TaskLocal` + `memoize` (we probably have to disable ability to modify `Local` from `memoize`)


alexandru:
Huh, interesting.

So we no longer need to do `Local.getContext` and `Local.setContext` on async boundaries?

oleg-py:
I have a haunting suspicion that it might introduce undesired sharing on Tasks running in parallel, that should be isolated. I don't think our test suite could detect that, since it mostly even doesn't use real parallelism.

Avasil:
> I have a haunting suspicion that it might introduce undesired sharing on Tasks running in parallel, that should be isolated. I don't think our test suite could detect that, since it mostly even doesn't use real parallelism.

As a user, I would like something that works like this:
```scala
for {
f1 <- task.start // copy "main" local but isolate for its own duration
f2 <- task.start // copy "main" local but isolate for its own duration 
_  <- Task.parZip2(f1.join, f2.join) // nondeterministic local value from either f1 or f2
} yield ()
```

```scala
for {
 // task1 and 2 copy "main" local but their own changes are isolated until entire Task returns
_ <- Task.parZip2(task1, task2)
_ <- ... // nondeterministic value from either task1 or 2
} yield ()
```

But I'm not sure how to achieve it and I can probably live with what we have right now as second best. :D Though I plan to play a bit in following days to make sure I can use it the way I would like to

oleg-py:
> `// nondeterministic local value from either f1 or f2`

I think that's problematic if your main fiber also interacts with locals before joining. You would lose all writes between start-join and would be forced to isolate `join` explicitly to avoid this.

In my (very small, mind you) experience with TaskLocal, it's best to write locals at the top level and only read or, worst case, `bind` them in layers below. I also think "stuff is shared unless said otherwise" is an okay model for reasoning about fibers and `parXXX` methods.

And with typelevel mouse and/or scala 2.13 you can use some chaining syntax to make it both more visible and convenient:
```scala
import TaskLocal.isolate
for {
f1 <- task.start |> isolate
f2 <- task.start |> isolate
_  <- Task.parZip2(f1.join, f2.join)
} yield ()
```

We can also have a helper like `.startIsolated` as a shorthand for that if we want to.

alexandru:
@Avasil is this PR still relevant?

Avasil:
> @Avasil is this PR still relevant?

I think we actually had a use case for a flag in `memoize` and `evalAsync` so might be not relevant.

PR #846, opened 2019-03-12
Ensure mapParallelOrdered runs in parallel

lukestephenson:
See #841

Is there a guarantee with the waiting for the semaphore permit that there will not be multiple permits blocked at once?  If so, the ordering might not be guaranteed with this approach.

PR #906, opened 2019-06-17
Fix race in MapParallelOrderedObservable

rfkm:
`MapParallelOrderedObservable` could be stuck forever.

Consider the following scenario:

1. get out of the `while` loop in `sendDownstreamOrdered` as `queue` is empty, but `sendDownstreamLock` is being still locked
2. another item `A` comes from upstream and is enqueued to `queue`
3. `A`'s future is completed, and then `sendDownstreamOrdered` is invoked
4. `sendDownstreamLock` is still locked, so the second call of `sendDownstreamOrdered` caused by `A` is rejected
5. `sendDownstreamLock` is finally unlocked
6. However, if there is no more item or there is no room in semaphore, there is no chance to kick `sendDownstreamOrdered` again

This can likely happen especially for small tasks (such tasks are not suitable for this operator in the first place, though). For example, I can reproduce it with the following simple observable:

```scala
Observable
  .range(1, 10000000)
  .mapParallelOrdered(10)(Task(_))
  .dump("O")
```

This patch will ensure that the polling of the queue is done as many as the number of elements.
As far as I have examined, there is no notable performance regression.

I'm not sure but #846 may have made the problem more likely.

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
@@ -18,7 +18,6 @@
 package monix.reactive.internal.operators
 
 import java.util.concurrent.ConcurrentLinkedQueue
-import java.util.concurrent.locks.ReentrantLock
 
 import monix.eval.Task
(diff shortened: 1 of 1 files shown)


Avasil:
Thank you!

I wish we could add regression test but it would probably take about a minute so probably not worth it.

PR #913, opened 2019-06-20
Optimize Task.guarantee

alexandru:
Currently `guarantee` is based on `bracket`, but its implementation could be simplified and made slightly more efficient.

This is something similar to what I did for `cats.effect.IO`, but I forgot about it.


monix-eval/shared/src/test/scala/monix/eval/TaskGatherUnorderedSuite.scala
```diff
@@ -134,12 +134,24 @@ object TaskGatherUnorderedSuite extends BaseTestSuite {
...
     var errorsThrow = 0
-    val task1 = Task.raiseError[Int](ex).executeAsync.doOnFinish { x =>
-      if (x.isDefined) errorsThrow += 1; Task.unit
-    }
-    val task2 = Task.raiseError[Int](ex).executeAsync.doOnFinish { x =>
-      if (x.isDefined) errorsThrow += 1; Task.unit
-    }
+
+    val task1 = Task
+      .raiseError[Int](ex)
```
oleg-py: Is it scalafmt style? That separation of companion object and a factory method on different lines kinda bothers me

alexandru: Unfortunately yes. Not sure what to do about it.

lorandszakacs: Huh, that's weird. I think you can do:
```
optIn {
  breaksInsideChains = false
  breakChainOnFirstMethodDot = false //pretty sure it's this one causing this.
}
```

monix-eval/shared/src/test/scala/monix/eval/TaskBracketSuite.scala
```diff
@@ -152,34 +152,68 @@ object TaskBracketSuite extends BaseTestSuite {
     }
   }
 
-  test("bracket is stack safe") { implicit sc =>
+  test("bracket works with auto-cancelable run-loops") { implicit sc =>
```
Avasil: Would it be possible to have a test that `acquire` is uncancelable?

Issue #908, opened 2019-06-19
Sporadic scala.MatchError inside Task.timeoutTo


Z1kkurat:
Sometimes `Task.timeout` fails with `scala.MatchError` with the following stacktrace.
```
scala.MatchError: null
	at monix.eval.Task.$anonfun$timeoutTo$1(Task.scala:2208)
	at monix.eval.internal.TaskRunLoop$.startFull(TaskRunLoop.scala:147)
	at monix.eval.internal.TaskRestartCallback.syncOnSuccess(TaskRestartCallback.scala:108)
	at monix.eval.internal.TaskRestartCallback$$anon$1.run(TaskRestartCallback.scala:128)
	at monix.execution.internal.Trampoline.monix$execution$internal$Trampoline$$immediateLoop(Trampoline.scala:66)
	at monix.execution.internal.Trampoline.startLoop(Trampoline.scala:32)
	at monix.execution.schedulers.TrampolineExecutionContext$JVMOptimalTrampoline.startLoop(TrampolineExecutionContext.scala:146)
	at monix.execution.internal.Trampoline.execute(Trampoline.scala:39)
	at monix.execution.schedulers.TrampolineExecutionContext.execute(TrampolineExecutionContext.scala:65)
	at monix.execution.schedulers.BatchingScheduler.execute(BatchingScheduler.scala:50)
	at monix.execution.schedulers.BatchingScheduler.execute$(BatchingScheduler.scala:47)
	at monix.execution.schedulers.ExecutorScheduler.execute(ExecutorScheduler.scala:34)
	at monix.eval.internal.TaskRestartCallback.onSuccess(TaskRestartCallback.scala:72)
	at monix.execution.Callback$Base.run(Callback.scala:244)
	at monix.eval.internal.TaskConnection$TrampolinedWithConn.run(TaskConnection.scala:227)
	at monix.execution.internal.Trampoline.monix$execution$internal$Trampoline$$immediateLoop(Trampoline.scala:66)
	at monix.execution.internal.Trampoline$ResumeRun$1.run(Trampoline.scala:51)
	at monix.execution.internal.forkJoin.AdaptedForkJoinTask.exec(AdaptedForkJoinTask.scala:27)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
[...]

PR #917, opened 2019-06-20
Callback received in Task.create should be thread-safe and should shift back to default Scheduler

alexandru:
Changes in these PR:

- in attempting to investigate #908, it became clear that the `Callback` passed in `Task.async` (and variants) should be thread-safe
  - this should be the user's responsibility, but in case the user fails to uphold the contract, it falls on us to investigate potential bugs and it creates frustration for the user 
  - in case `onSuccess` / `onError` are called twice, an `IllegalStateException` is still being thrown, the difference is that now this conflict is able to be detected in a multi-threading context
  - this should give us or the users better error messages
- removing the `allowContinueOnCallingThread` flag on `Task.async`, `Task.cancelable`, `Task.create`, etc... introduced in #921
  - we don't need this flag
  - this is now replaced with a cool optimization that detects if the `register` function actually forks and if so, then we shift back, if not then we don't shift
- also changes`Task.fromFuture` to use `Future.onComplete` as it was meant to be used


alexandru:
Released test version: `3.0.0-RC2-cb74787`

project/MimaFilters.scala
```diff
@@ -4,6 +4,9 @@ import com.typesafe.tools.mima.core.ProblemFilters.exclude
 object MimaFilters {
 
   lazy val changesFor_3_0_0: Seq[ProblemFilter] = Seq(
+    // Internals for https://github.com/monix/monix/pull/917
+    exclude[MissingClassProblem]("monix.eval.internal.TaskConnection$TrampolinedWithConn"),
+    exclude[DirectMissingMethodProblem]("monix.eval.internal.TaskConnection.trampolineCallback"),
     // Breaking changes for https://github.com/monix/monix/pull/921
```
Avasil: We could remove this to make sure it's no longer breaking bincompat

monix-eval/jvm/src/test/scala/monix/eval/TaskAsyncAutoShiftJVMSuite.scala
```diff
@@ -0,0 +1,943 @@
...
+      r
+    }(ec)
+
+    started.await()
+    (CancelableFuture(f, Cancelable.empty), latch)
+  }
+
+  def repeatTest(times: Int)(f: => Future[Unit])(implicit ec: ExecutionContext): Future[Unit] = {
+    if (times > 0) f.flatMap(_ => repeatTest(times - 1)(f))
+    else Future.successful(())
```
Avasil: nitpick: `Future.unit`

monix-eval/jvm/src/test/scala/monix/eval/TaskAsyncAutoShiftJVMSuite.scala
```diff
@@ -0,0 +1,943 @@
...
+        assert(name.startsWith(ThreadName), s"'$name' should start with '$ThreadName'")
+      }
+    }
+  }
+
+  testAsync("Task.deferFuture(completed) should not shift to the main scheduler") { s =>
+    implicit val ec: Scheduler = Scheduler.global
+
+    repeatTest(1000) {
+      val f = Future.successful(())
```
Avasil: nitpick: Future.unit

monix-eval/jvm/src/test/scala/monix/eval/TaskAsyncAutoShiftJVMSuite.scala
```diff
@@ -0,0 +1,943 @@
...
+        assert(name.startsWith(ThreadName), s"'$name' should start with '$ThreadName'")
+      }
+    }
+  }
+
+  testAsync("Task.deferFutureAction(completed) should not shift to the main scheduler") { s =>
+    implicit val ec: Scheduler = Scheduler.global
+
+    repeatTest(1000) {
+      val f = Future.successful(())
```
Avasil: nitpick: Future.unit

monix-eval/shared/src/main/scala/monix/eval/internal/TaskCreate.scala
```diff
@@ -184,44 +166,85 @@ private[eval] object TaskCreate {
...
+          state.set(2)
+          this.value = null.asInstanceOf[A]
+          this.error = e
+          isSameThread = true
+          TrampolineExecutionContext.immediate.execute(this)
+      }
+    }
+
+    override def run(): Unit = {
+      if (!isSameThread) {
```
Avasil: I'm not sure I understand how it works. I will appreciate explanation

We create this `Callback` / `Runnable` in `Task.create`. At some point we run it, `isSameThread` is `false` so we reset the frameRef and then go to `startExecution()`, setting `isSameThread` variable to `true` or `false`. So I assume the instance of this callback is reused somewhere? Is it for each time given `Task.create` is used? Why do we reset it at least once?

monix-eval/shared/src/test/scala/monix/eval/TaskAsyncSuite.scala
```diff
@@ -156,29 +156,29 @@ object TaskAsyncSuite extends BaseTestSuite {
     assertEquals(f.value, Some(Success(10000)))
   }
 
-  test("Task.async0(allowContinueOnCallingThread = true) should execute") { implicit s =>
-    val task = Task.async0[Int]({ (ec, cb) =>
+  test("Task.async0() should execute") { implicit s =>
```
Avasil: I think all of the tests that used to be with `allowContinueOnCallingThread = true` are duplicated in this file now so we could remove them / revert to the state from before my change

monix-eval/shared/src/test/scala/monix/eval/TaskCancelableSuite.scala
```diff
@@ -140,92 +140,97 @@ object TaskCancelableSuite extends BaseTestSuite {
...
   }
 
-  test("Task.cancelable0(allowContinueOnCallingThread = true) should be stack safe on repeated, right-associated binds") {
-    implicit s =>
-      def signal[A](a: A): Task[A] =
-        Task.cancelable0[A]({ (_, cb) =>
-          cb.onSuccess(a)
-          Task.unit
-        }, allowContinueOnCallingThread = true)
+  test("Task.cancelable0() should be stack safe on repeated, right-associated binds") { implicit s =>
```
Avasil: I think all of the tests that used to be with `allowContinueOnCallingThread = true` are duplicated in this file now so we could remove them / revert to the state from before my change

monix-eval/shared/src/test/scala/monix/eval/TaskFromFutureSuite.scala
```diff
@@ -88,264 +88,253 @@ object TaskFromFutureSuite extends BaseTestSuite {
     assertEquals(result.value, Some(Success(1)))
   }
 
-  test("Task.fromFuture(allowContinueOnCallingThread = true) should be faster for completed futures, success") {
-    implicit s =>
-      val t = Task.fromFuture(Future.successful(10), allowContinueOnCallingThread = true)
-      val f = t.runToFuture
-      assertEquals(f.value, Some(Success(10)))
+  test("Task.fromFuture() should be faster for completed futures, success") { implicit s =>
```
Avasil: We could remove some of the tests here too and just keep those that check for faster completion

Perhaps it will be faster to just revert to the state from before my changes

monix-eval/jvm/src/test/scala/monix/eval/TaskAsyncAutoShiftJVMSuite.scala
oleg-py: Isn't that 2.12+

Avasil: if that's a case, disregard my comment :D

monix-eval/shared/src/main/scala/monix/eval/internal/TaskCreate.scala
```diff
@@ -184,44 +166,85 @@ private[eval] object TaskCreate {
...
+    }
+
+    private def startExecution(): Unit = {
+      // Optimization — if the callback was called on the same thread
+      // where it was created, then we are not going to fork
+      isSameThread = Platform.currentThreadId() == threadId
+      try {
+        ctx.scheduler.execute(
+          if (isSameThread)
+            this
```
oleg-py: if it's the same thread, do we really need an async boundary from `execute` here? Can't we just call `run()`?

monix-eval/shared/src/main/scala/monix/eval/internal/TaskCreate.scala
oleg-py: This is my understanding:

`StartAsyncBatchRunnable` forces an async boundary if we're on the other thread
Resetting frameRef ensures that internal batching mechanism isn't deoptimized by thinking there has been a lot of runs without async boundaries. This is only done if we're inserting the async boundary.
We only call methods once. `startExecution` is called first, upon completion, to set the `isSameThread`. `run` is called later via scheduling.

Avasil: Thanks, I misread which `onSuccess` / `onError` is called in `run()`, makes sense now

monix-eval/jvm/src/test/scala/monix/eval/TaskAsyncAutoShiftJVMSuite.scala
alexandru: Yeah, `Future.unit` is not available in 2.11 ☹️

project/MimaFilters.scala
alexandru: Will do.

monix-eval/shared/src/main/scala/monix/eval/internal/TaskCreate.scala
alexandru: An additional detail ... the callback inherits from `TrampolinedRunnable`. If executed as is (`scheduler.execute(this)`), then it gets executed via the trampoline on the current thread.

On the other hand by wrapping it in a `StartAsyncBatchRunnable`, we ensure that a fork happens.

monix-eval/shared/src/main/scala/monix/eval/internal/TaskCreate.scala
alexandru: Not really, because we'd run into stack safety issues. Some sort of async boundary, either via a real fork or via the trampoline, is needed.

Before this change the callback was wrapped in a `Callback.trampoline()`. It's also why we have the settings `trampolineBefore` and `trampolineAfter` on `Async`. The point of those settings was to avoid wrapping callbacks in additional instances and we don't need to use those here, since we are wrapping the callback anyway.

monix-eval/shared/src/test/scala/monix/eval/TaskAsyncSuite.scala
alexandru: 👍 will take care of it.

PR #918, opened 2019-06-21
Fix #793: switchMap should wait for last child to complete

### LINKED ISSUE #793: Version of switchMap that waits for the child observable to complete
Current behavior of `switchMap` in Monix (3.0.0-RC2):

If the "parent" observable completes, an observable that was created using `switchMap` will complete immediately. This is different from RxScala (and apparently also from fs2), where the "switchMapped" observable will not complete until its current ("latest") child observable is completed.

(Find a minimal example here: https://gist.github.com/MartinHH/6977ee66963e438eaf9a4d8f6314893b)

While I could imagine that the current behavior is desirable in certain situations (and I guess it might be dangerous to just change the behavior of Monix' `switchMap`), I think it would be good to have a convenient method that supports the other behavior (wait for current child to complete).

oleg-py:


monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/SwitchMapObservable.scala
```diff
@@ -61,15 +63,24 @@ private[reactive] final class SwitchMapObservable[A, B](source: Observable[A], f
           activeChild := childObservable.unsafeSubscribeFn(new Observer[B] {
             def onNext(elem: B) =
               self.synchronized {
-                if (upstreamIsDone || myChildIndex != activeChildIndex)
+                if (myChildIndex != activeChildIndex)
                   Stop
                 else {
                   ack = out.onNext(elem).syncOnStopOrFailure(_ => cancelFromDownstream())
```
Avasil: BTW it seems like this `ack` var is unused other than this two lines

oleg-py: It's set below to `Stop` in `cancelFromDownstream`

PR #919, opened 2019-06-21
Fix #719: propagate cancellation to tasks in async Consumers

### LINKED ISSUE #719: Limited cancellation logic for asynchronous consumers
When we have an observable that we want to consume asynchronously (via, for example, `FoldLeftAsyncConsumer`), we're often interested in having a solid story for cancellation of the tasks that are being run in the background.
  
For example, imagine I have an observable of file watch events and I want to read the contents of a file every time there's such an event asynchronously. Whenever the observable is cancelled, I want all the running tasks that are reading contents from files to be correctly cancelled too.
  
By default, `FoldLeftTaskConsumer` does not do that https://github.com/monix/monix/blob/2e8a6a8790d1b971cf9ecb7ca45327636d532997/monix-reactive/shared/src/main/scala/monix/reactive/internal/consumers/FoldLeftTaskConsumer.scala#L75, it only returns a dummy assignable.
  
However, this is rarely what we want. We often want a consumer that will cancel all the running tasks, like I do in bloop via my own `FoldLeftAsyncConsumer` https://github.com/scalacenter/bloop/blob/825fb39bc94923a94ac7bcfbf4a484b6506ad8b2/frontend/src/main/scala/bloop/monix/FoldLeftAsyncConsumer.scala#L62-L67
  
However, changing the consumer is not enough to get the behavior I want. The definition of `Consumer` assumes that all consumers return an empty assignable cancelable and assigns the subscription cancelable to the consumer one https://github.com/monix/monix/blob/2e8a6a8790d1b971cf9ecb7ca45327636d532997/monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala#L67-L80
  
This is problematic because if the assignable is a multi cancellable, it will override the previous tasks. If it is single, it will ignore the cancelable given by the source subscription. In both cases, we lose cancellation capabilities.
  
Instead of the current logic, I propose to upstream my redefinition of the `apply` method in `Consumer` (which in 3.0.0-RC1 is final and cannot be overridden):
  
```scala
 Task.cancelable0 { (scheduler, cb) => 
   val (out, consumerSubscription) = createSubscriber(cb, scheduler) 
   val sourceSubscription = source.subscribe(out) 
   if (consumerSubscription.isInstanceOf[Cancelable.IsDummy]) 
     sourceSubscription 
   else CompositeCancelable(sourceSubscription, consumerSubscription)
}
```
  
If I don't miss anything, I think this should fix the issue upstream. We've been using this technique in bloop for quite some time, so I think it should work.

oleg-py:


monix-reactive/shared/src/main/scala/monix/reactive/internal/consumers/ForeachAsyncConsumer.scala
```diff
@@ -58,6 +67,12 @@ private[reactive] final class ForeachAsyncConsumer[A](f: A => Task[Unit]) extend
         }
     }
 
-    (out, AssignableCancelable.dummy)
+    (out, SingleAssignCancelable.plusOne(Cancelable { () =>
+      out.synchronized {
+        isDone = false
```
Avasil: Shouldn't this be `isDone = true`?

monix-reactive/shared/src/main/scala/monix/reactive/internal/consumers/ForeachAsyncConsumer.scala
```diff
@@ -31,13 +31,22 @@ import scala.concurrent.Future
 private[reactive] final class ForeachAsyncConsumer[A](f: A => Task[Unit]) extends Consumer[A, Unit] {
 
   def createSubscriber(cb: Callback[Throwable, Unit], s: Scheduler): (Subscriber[A], AssignableCancelable) = {
+    var isDone = false
+    var lastCancelable = Cancelable.empty
+
     val out = new Subscriber[A] {
       implicit val scheduler = s
       private[this] var isDone = false
```
Avasil: I think you've forgotten to remove it

monix-reactive/shared/src/main/scala/monix/reactive/internal/consumers/MapTaskConsumer.scala
```diff
@@ -43,7 +45,10 @@ private[reactive] final class MapTaskConsumer[In, R, R2](source: Consumer[In, R]
             try {
               val task = f(value)
               streamErrors = false
-              task.runAsync(cb)
+              self.synchronized {
+                if (!isCancelled)
```
Avasil: I feel like we don't have to synchronize `isCancelled` if it's `volatile` (or it doesn't have to be `volatile` if it's only used in synchronized block)

monix-reactive/shared/src/main/scala/monix/reactive/internal/consumers/MapTaskConsumer.scala
```diff
@@ -30,7 +30,9 @@ private[reactive] final class MapTaskConsumer[In, R, R2](source: Consumer[In, R]
   extends Consumer[In, R2] {
 
   def createSubscriber(cb: Callback[Throwable, R2], s: Scheduler): (Subscriber[In], AssignableCancelable) = {
-    val asyncCallback = new Callback[Throwable, R] {
+    @volatile var lastCancelable: Cancelable = Cancelable.empty
```
Avasil: Does it have to be `volatile`?

monix-reactive/shared/src/test/scala/monix/reactive/consumers/MapTaskConsumerSuite.scala
```diff
@@ -24,6 +24,8 @@ import monix.execution.exceptions.DummyException
 import monix.reactive.{BaseTestSuite, Consumer, Observable}
 import scala.util.Failure
 
+import monix.execution.atomic.Atomic
```
Avasil: unused import?

monix-reactive/shared/src/main/scala/monix/reactive/internal/consumers/MapTaskConsumer.scala
oleg-py: Yeah, I'm synchronizing access across two variables here. Volatile probably is redundant

Avasil:
Missing `scalafmtAll` :D

PR #921, opened 2019-06-23
Shifting back to the default pool in Task.fromFuture, Task.async and …

### LINKED ISSUE #916: Shifting back to the default pool in Task.fromFuture, Task.async and others
Currently code like this:

```scala
Task.async[Unit] { cb =>
  foreignEC.submit(new Runnable {
    override def run(): Unit = {
      cb(Right(()))
    }
  })
}
```
Stays on `foreignEC` for the next operations until async boundary. Similar to `Future` and probably the rest similar operators. 

I'm not familiar with any situation when it's desirable but it can easily catch off-guard and is a source of confusion, see [recent blog post](https://blog.softwaremill.com/thread-shifting-in-cats-effect-and-zio-9c184708067b) and it was already changed for `fromFuture` in Cats-Effect: https://github.com/typelevel/cats-effect/pull/546

I believe the rest is planned to be fixed in `3.0`. Do you think we should do it and can we allow it such a change right now? cc @oleg-py @alexandru 

I feel like it wouldn't be noticable to the users unless there are use cases from current behavior that I'm not aware of.

If we agree to do it, I can provide a PR

Avasil:
…others

Fixes #916

Should we have old behavior available? New method (`unsafeXXX`)? Flag? Deprecation?  If I were just introducing these builders, I wouldn't include not-shifting variant so the API is safer but in our situation I am not so sure.


monix-eval/shared/src/main/scala/monix/eval/internal/TaskFromFuture.scala
```diff
@@ -42,14 +46,14 @@ private[eval] object TaskFromFuture {
             rawAsync(startSimple(_, _, f))
         }
       case Some(value) =>
-        Task.fromTry(value)
+        Task.async(cb => cb(value))
```
Avasil: It sucks but I don't know if there is any way to detect whether `Future` has been built using `successful` or not

monix-eval/shared/src/test/scala/monix/eval/TaskAsyncSuite.scala
```diff
@@ -88,7 +95,19 @@ object TaskAsyncSuite extends BaseTestSuite {
...
   test("Task.async0 works for immediate successful value") { implicit sc =>
     val task = Task.async0[Int]((_, cb) => cb.onSuccess(1))
-    assertEquals(task.runToFuture.value, Some(Success(1)))
+    val f = task.runToFuture
+    sc.tick()
+
+    assertEquals(f.value, Some(Success(1)))
+  }
+
+  test("Ta0sk.async0 works for immediate error") { implicit sc =>
```
oleg-py: Typo?

Avasil:
Oh, I didn't run `monix-reactive` tests. If we go with non-shifting versions I can rewrite them to use those, otherwise I'll fix tests

monix-eval/shared/src/main/scala/monix/eval/internal/TaskFromFuture.scala
oleg-py: Pretty sure those futures have `.value` set to `Some(try)`

Avasil: `Future` crossing async boundaries can also be  `Some(try)` if it finished earlier

oleg-py:
@Avasil @alexandru do you think it would make sense for `Task.contextShift.evalOn` to wrap EC into a Scheduler and use `executeOn` instead of bracketing two shifts?

It's pretty fragile as it is, and it seems more so with auto-shifting in builders that we're using in a typeclass instance.

Avasil:
@oleg-py I was wondering about it too and feel like it's the way to go.

Avasil:
@oleg-py @alexandru 
What do you think about the name `createDirect`, `asyncDirect` etc. instead of `createUnsafe`, `asyncUnsafe` ? "Direct" because we are calling `Task` callback directly and continue on the last thread directly

oleg-py:
@Avasil honestly, probably a boolean flag with a default value and a consistent descriptive name on main builders (something like `allowContinueOnCallingThread = false` or `shiftToMainPoolAfterResult = true`) would be a better option. Otherwise, you still have to explain what you mean by `direct`, as you had to explain what `unsafe` meant.

Default parameters would probably make most sense here, as the method is still conceptually doing the same thing (lifting the asynchronous/cancelable operation into a Task) with a minor performance/thread-management option, and we don't have to worry much about eta-expansion issues it can cause due to it being FFI method that requires very explicit handling anyway.

Another option is to add an extra flag to `Task.Options` so you can `executeWithOptions` or set it in `TaskApp` and actually control from the outside if the shifting is going to happen. A positive side to it is that you can configure typeclass behavior this way. A negative one is that it's less intuitive, as we don't keep these two related things together - which might not be too bad for an option intended for power users.

Avasil:
> @Avasil honestly, probably a boolean flag with a default value and a consistent descriptive name on main builders (something like `allowContinueOnCallingThread = false` or `shiftToMainPoolAfterResult = true`) would be a better option. Otherwise, you still have to explain what you mean by `direct`, as you had to explain what `unsafe` meant.

I will update a PR tomorrow, in my opinion flag fits quite nicely here. Also, if we change our minds it will be easier to deprecate it and introduce extra builders than in the other direction.

> Another option is to add an extra flag to Task.Options so you can executeWithOptions or set it in TaskApp and actually control from the outside if the shifting is going to happen. A positive side to it is that you can configure typeclass behavior this way. A negative one is that it's less intuitive, as we don't keep these two related things together - which might not be too bad for an option intended for power users.

I always look at `Task.Options` as something to configure `TaskRunLoop` itself and here we deal with a `builder`. I also feel like those users who will use both variants may mix them a lot depending on the methods they invoke and that won't be very convenient

Avasil:
@oleg-py Updated! If you feel like the old way was better then it's easy to revert

oleg-py:
Nah, I think it's good. Not ambiguous like `Unsafe`, very explicit if you name boolean parameters, and arguably not less convenient to use than `async`-style builders already are.

Avasil:
Awesome! I'll just do sanity check if I didn't miss anything important in a hurry and merge it

PR #922, opened 2019-06-24
Update README.md

arlequin-nyc:
Added Agoda (brand belonging to Booking Holdings)


Avasil:
Thank you @arlequin-nyc :)

PR #933, opened 2019-06-27
Add flatmap combinators

### LINKED ISSUE #929: Task.*> and <* combinator

```scala
final def *>[B](that: => Task[B]): Task[B]
final def <*[B](that: => Task[B]): Task[A]
```

```scala
taskA *> taskB // runs taskA, then taskB, returns the result of taskB
taskA <* taskB // runs taskA, then taskB, returns the result of taskA
```

valenterry:
Closes #929


valenterry:
> It would be nice to have a small example in scaladocs (like a print + return value)

Added examples 👍

alexandru:
Hey guys, see my comments on: https://github.com/monix/monix/pull/935

valenterry:
@Avasil @alexandru changed it accordingly to #935, please re-review

PR #935, opened 2019-06-27
Coeval flatmap combinator

### LINKED ISSUE #930: Coeval.*> and <* combinator

```scala
final def *>[B](that: => Coeval[B]): Coeval[B]
final def <*[B](that: => Coeval[B]): Coeval[A]
```

```scala
coevalA *> coevalB // runs coevalA , then coevalB, returns the result of coevalB
coevalA <* coevalB // runs coevalA , then coevalB, returns the result of coevalA
```

valenterry:
Closes #930

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
@@ -183,6 +183,32 @@ import scala.util.{Failure, Success, Try}
 sealed abstract class Coeval[+A] extends (() => A) with Serializable { self =>
   import monix.eval.Coeval._
 
+  /** Evaluates this underlying computation first and then, when successful, the given one.
+    * Returns the result of this underlying computation.
+    *
+    * Example:
+    * {{{
+    *   val combined = Coeval{println("first"); "first"} <* Coeval{println("second"); "second"}
+    *   // Prints "first" and then "second"
+    *   // Result value will be "first"
+    * }}}
+    */
+  final def <*[B](that: => Coeval[B]): Coeval[A] =
+    this.flatMap(a => that.map(_ => a))
+
+  /** Runs this underlying computation first and then, when successful, the given one.
+    * Returns the result of the given underlying computation.
+    *
+    * Example:
+    * {{{
+    *   val combined = Coeval{println("first"); "first"} *> Coeval{println("second"); "second"}
+    *   // Prints "first" and then "second"
+    *   // Result value will be "second"
+    * }}}
+    */
+  final def *>[B](that: => Coeval[B]): Coeval[B] =
+    this.flatMap(_ => that)
+
   /** Evaluates the underlying computation and returns the result.
     *
     * NOTE: this can throw exceptions.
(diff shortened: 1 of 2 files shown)


monix-eval/shared/src/main/scala/monix/eval/Coeval.scala
```diff
@@ -183,6 +183,32 @@ import scala.util.{Failure, Success, Try}
...
+    * Returns the result of the given underlying computation.
+    *
+    * Example:
+    * {{{
+    *   val combined = Coeval{println("first"); "first"} *> Coeval{println("second"); "second"}
+    *   // Prints "first" and then "second"
+    *   // Result value will be "second"
+    * }}}
+    */
+  final def *>[B](that: => Coeval[B]): Coeval[B] =
```
alexandru: This isn't consistent with Cats's `*>`, which works with strict parameters.

See: https://github.com/typelevel/cats/blob/v1.6.1/core/src/main/scala/cats/Apply.scala#L119

What you're thinking of is `>>`:

https://github.com/typelevel/cats/blob/v1.6.1/core/src/main/scala/cats/syntax/flatMap.scala#L33

valenterry: @alexandru I can merge the PRs. Should we just rename `*>` to `>>` then or add both `*>` and `>>` with different semantics?

Avasil: @alexandru strict parameter is not safe with infinite recursion and it sometimes catches users off-guard in cats-effect / monix gitter channels. You can't explain it as just `flatMap(_ => a)` because of this catch 

In my opinion if we go with one operator (to not pollute API too much) it's better to use `*>` exactly because the one from `Cats` has gotchas. And I think there is a law in cats that `flatMap` and `map2` should be consistent so I'd prefer to have it without corner cases. :D

oleg-py: I'd go with both sharks and `>>`, or _only_ `>>`. IMO that's the least surprising operator that we should promote, and most of the times its "right-bias" is exactly what you need for quick effect combination that doesn't yet need for-comprehension.

Avasil: @oleg-py what makes `<*` better than `<<` and what's your opinion on consistency vs safety? I'd rather avoid strict `*>` so either have it like in the PR or `>>` instead

oleg-py: IMO having inconsistency that is safer only as long as you don't abstract Task/Coeval due to extension methods being shadowed isn't really safer. It'll eventually bite the user.

I didn't say that `<<` is better, but it is both deprecated and strict in cats, so I'd not add it for same consistency reasons.

I'm fine with strict `*>` for cats consistency as long as we have non-strict `>>`.

Avasil: Okay, fair enough. I guess we can let @valenterry decide. :D 
by-name `>>` only or `>>` with strict `*>` and `<*`

If we go with the latter, I would add a comment in scaladoc with a warning

valenterry: I think I'd go for `>>` only then, even though it is a bad sad that `<<` deprecated in cats.

valenterry:
@alexandru changed it accordingly, please re-review

PR #940, opened 2019-07-01
fix typo

takayahilton:


monix-catnap/shared/src/main/scala/monix/catnap/cancelables/AssignableCancelableF.scala
```diff
@@ -71,7 +71,7 @@ object AssignableCancelableF {
 
   /**
     * Represents an [[AssignableCancelableF]] with no internal state and that
-    * doesn't do anything, either on assignment or on cancelation.
+    * doesn't do anything, either on assignment or on cancellation.
```
Avasil: I was under impression it was intended, is that correct @alexandru ?

alexandru: Indeed, cancelation is correct with both one and two Ls. With one L it's basically the American spelling.

And we used `Cancelable` as a type name, it's good to be consistent.

alexandru:
Will now close this, hope the explanation was clear.

Issue #914, opened 2019-06-20
Add missing F[_]-polymorphic methods to Observable


Avasil:
Many operators, like `mapEval` have equivalent working for `F[_]` which is convenient for people using other Effect type or writing using Tagless Final.

There are missing methods, e.g.:
- `mapParallelOrdered`
- `mapParallelUnordered`
- ` fromIterator`
- `consumeWith`
- `completedL`
And more. 

In some cases we might be able to express them in terms of Cats-Effect typeclasses but converting to/from `Task` is ok too.

entangled90:
Hi, I'm picking this up if nobody has started working on it.
One simple question.
What's the standard way of converting a Task` to an `F[_]`? 
The `TaskLike` constraint only wokrs on the other direction, from `F[_]` to `Task`.

Avasil:
Solved in #941 
There are more methods but the most important ones are covered so closing this for now

PR #941, opened 2019-07-02
Added some polymorphic methods for Observable

entangled90:
Added some polymorphic methods for types which are `TaskLike` or `TaskLift`.
Let me know if someone is interested in other methods. 
I did not find additional test for existing polymorphic variants of methods, therefore I didn't add any test.

Issue: #914

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -4561,7 +4561,7 @@ object Observable extends ObservableDeprecatedBuilders {
   def fromIterator[A](resource: Resource[Task, Iterator[A]]): Observable[A] =
     Observable.fromResource(resource).flatMap(fromIteratorUnsafe)
 
-  /** Version of [[fromIterator]] that can work with generic
+  /** Version of fromIterator that can work with generic
     * `F[_]` tasks, anything that's supported via [[monix.eval.TaskLike]]
     * conversions.
     *
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -4572,7 +4572,7 @@ object Observable extends ObservableDeprecatedBuilders {
     *  - `scala.concurrent.Future`
     *  - ...
     */
-  def fromIteratorF[A, F[_]](iteratorF: F[Iterator[A]])(implicit F: TaskLike[F]): Observable[A] =
+  def fromIteratorF[F[_], A](iteratorF: F[Iterator[A]])(implicit F: TaskLike[F]): Observable[A] =
     fromIterator(F(iteratorF))
 
   /** Converts any `Iterator` into an observable.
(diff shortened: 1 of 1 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -4479,6 +4561,20 @@ object Observable extends ObservableDeprecatedBuilders {
   def fromIterator[A](resource: Resource[Task, Iterator[A]]): Observable[A] =
     Observable.fromResource(resource).flatMap(fromIteratorUnsafe)
 
+  /** Version of [[fromIterator]] that can work with generic
```
Avasil: I think you need to use `[[fromIterator[A](task* fromIterator(task)]]` or just `fromIterator` to successfully generate doc (currently it fails because of ambiguity)

entangled90: Removed the scaladoc link, considering the other method is just a few lines above.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -607,6 +607,20 @@ abstract class Observable[+A] extends Serializable { self =>
...
+    * conversions.
+    *
+    * So you can work among others with:
+    *
+    *  - `cats.effect.IO`
+    *  - `monix.eval.Coeval`
+    *  - `scala.concurrent.Future`
+    *  - ...
+    */
+  final def consumeWithF[R, F[_]](f: Consumer[A, R])(implicit F: TaskLift[F]): F[R] =
```
Avasil: Sorry for not noticing right away, the convention is to have `F[_]` first
Otherwise looks great :D

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -2166,6 +2180,33 @@ abstract class Observable[+A] extends Serializable { self =>
...
+    *
+    * @param f is the mapping function that produces tasks to execute
+    *        in parallel, which will eventually produce events for the
+    *        resulting observable stream
+    *
+    * @see [[mapParallelUnorderedF]] for a variant that does not preserve order
+    *     which may lead to faster execution times
+    * @see [[mapEvalF]] for serial execution
+    */
+  final def mapParallelOrderedF[B, F[_]](parallelism: Int)(
```
Avasil:
```suggestion
  final def mapParallelOrderedF[F[_], B](parallelism: Int)(
```

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -2194,6 +2235,34 @@ abstract class Observable[+A] extends Serializable { self =>
...
+    *        back-pressured
+    *
+    * @param f is the mapping function that produces tasks to execute
+    *        in parallel, which will eventually produce events for the
+    *        resulting observable stream
+    *
+    * @see [[mapParallelOrdered]] for a variant that does preserve order
+    * @see [[mapEval]] for serial execution
+    */
+  final def mapParallelUnorderedF[B, F[_]](parallelism: Int)(
```
Avasil:
```suggestion
  final def mapParallelUnorderedF[F[_], B](parallelism: Int)(
```

Avasil:
Thank you @entangled90 :)

PR #945, opened 2019-07-07
Add collectWhile observable

### LINKED ISSUE #925: Observable.collectWhile
```scala
final def collectWhile[B](pf: PartialFunction[A, B]): Observable[B]

val o: Observable[Option[A]] = ???
// Observable[A] that termiantes when there is a None value
o.collectWhile { case Some(value) => value }
```
Equivalent to:
```scala
o
  .takeWhile(_.isDefined)
  .collect { case Some(value) => value }
``` 
But described as its own operator. Basically merging these two:
- https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/TakeByPredicateOperator.scala
- https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/CollectOperator.scala

Not sure if there is noticable performance gain (might be interesting to measure!) but it shouldn't be too difficult to write anyway

Also there is similar thing in fs2: https://github.com/functional-streams-for-scala/fs2/blob/series/1.1/core/shared/src/main/scala/fs2/Stream.scala#L2580

allantl:
Closes #925

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -865,6 +865,16 @@ abstract class Observable[+A] extends Serializable { self =>
   final def collect[B](pf: PartialFunction[A, B]): Observable[B] =
     self.liftByOperator(new CollectOperator(pf))
 
+  /** Takes longest prefix of elements that satisfy the given partial function
+    * and returns a new Observable that emits those elements.
+    *
+    * @param pf the function that filters and maps the source
+    * @return an observable that emits the transformed items by the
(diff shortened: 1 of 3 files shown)


Avasil:
Thank you @allantl !
Looks great but would you mind adding 1 test to make sure it completes early? e.g. equivalence to `takeWhile` + `collect` with property based test, similar to [Iterant tests](https://github.com/monix/monix/blob/master/monix-tail/shared/src/test/scala/monix/tail/IterantCollectSuite.scala#L467)

allantl:
@Avasil Sure, its done

Issue #923, opened 2019-06-25
Local is exposing too many implementation details


alexandru:
We should make the `Context` class in `Local` private or at least its implementation.

Is there any reason for the user to have access to the methods of `Context` or to `Bound` / `Unbound`?

The problem is that I'm sure this isn't the final `Local` implementation and it would be a pity to break binary compatibility for simple fixes.

alexandru:
> Is there any reason for the user to have access to the methods of Context or to Bound / Unbound?

Actually I see that the macros are calling those methods now 🙂 I'm a little worried that we are exposing methods with `Any` in their signature.

Also are the `Bound` / `Unbound` classes required to be public?

oleg-py:
Pretty sure Bound / Unbound being public is an oversight.

We could remove macros for other cases, if they aren't really buying us much performance. I preserved them as they were already here.

Also we can make things `private[monix]` (should be bincompatible, IIRC that translates to JVM public anyway). I had some ideas for a Local that can work for any F if its ContextShift is backed by a `TracingScheduler`, which is why I left things quite lax.

At least parts that are used in `TaskLocal` should remain public for anybody daring to do such thing. One thing in particular is that `isolate` needs to be reimplemented in terms of effect type's `bracket` with `getContext`/`setContext`.

Avasil:
Fixed in #946, I assume we don't do anything else related to it

PR #946, opened 2019-07-08
Expose less implementation details in Local

Avasil:
My take on #923

Just added few `private`s. 
I've left `Context.mkIsolated` and `Context.bind` public because they are used by `TaskLocal`

### DIFF
--- a/monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
+++ b/monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
@@ -60,6 +60,9 @@ object Local {
   def clearContext(): Unit =
     localContext.set(defaultContext())
 
+  /** Execute a block of code without propagating any `Local.Context`
+    * changes outside.
+    */
   def isolate[R](f: => R): R =
     macro Macros.isolate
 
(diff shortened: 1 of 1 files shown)


monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -114,7 +114,7 @@ object Local {
     }
 
   /** Macros implementations for [[bind]] and [[bindClear]]. */
-  class Macros(override val c: whitebox.Context) extends InlineMacros with HygieneUtilMacros {
+  private class Macros(override val c: whitebox.Context) extends InlineMacros with HygieneUtilMacros {
```
oleg-py: Is that okay to make that private? I don't even know how scoping works for macros.

Avasil: me neither but it shouldn't compile if it's wrong, right? Or is there a way it could still fail somewhere?

oleg-py: That's the point, I have no clue 😆

Avasil: @oleg-py 

I changed `package` of `LocalSuite` to something completely else and it passed
Also moved it to `monix-eval` subproject with different package (without monix prefix) and worked as well

Looks like it can be private as long as it is in scope of public methods accessing it. I asked someone with some experience with macros and that was his thought too

Avasil:
Cool, I'm merging it then but will leave issue open for now in case @alexandru wants to do more

PR #948, opened 2019-07-09
Make Task.memoize play well with Local

### LINKED ISSUE #856: Task.memoize doesn't play well with Local

Avasil:
Fixes #856

Weird fix but seems to work

### DIFF
--- a/monix-eval/shared/src/test/scala/monix/eval/TaskMemoizeOnSuccessSuite.scala
+++ b/monix-eval/shared/src/test/scala/monix/eval/TaskMemoizeOnSuccessSuite.scala
@@ -672,7 +672,7 @@ object TaskMemoizeOnSuccessSuite extends BaseTestSuite {
 
     val t = for {
       local <- TaskLocal(0)
-      ii <- Task.evalAsync(i.incrementAndGet())
+      ii    <- Task.evalAsync(i.incrementAndGet())
       _     <- local.write(ii)
       result <- Task.parZip2(
         memoizedTask.flatMap { _ =>
--- a/monix-eval/shared/src/test/scala/monix/eval/TaskMemoizeOnSuccessSuite.scala
+++ b/monix-eval/shared/src/test/scala/monix/eval/TaskMemoizeOnSuccessSuite.scala
@@ -18,6 +18,7 @@
 package monix.eval
 
 import monix.execution.Callback
+import monix.execution.atomic.AtomicInt
 import monix.execution.exceptions.DummyException
 import monix.execution.internal.Platform
 import scala.concurrent.Promise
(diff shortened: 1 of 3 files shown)


monix-eval/shared/src/test/scala/monix/eval/TaskMemoizeSuite.scala
```diff
@@ -663,4 +665,39 @@ object TaskMemoizeSuite extends BaseTestSuite {
...
+        },
+        memoizedTask.flatMap { _ =>
+          local.read.executeAsync
+        }
+      )
+    } yield result
+
+    val f1 = t.runToFutureOpt
+    val f2 = t.runToFutureOpt
+    val f3 = t.runToFutureOpt
```
oleg-py: BTW, one of the crazy Scala features is that you can write:
`val f1, f2, f3 = t.runToFutureOpt`

Not sure if you want to, though 😄

monix-eval/shared/src/test/scala/monix/eval/TaskMemoizeOnSuccessSuite.scala
```diff
@@ -645,13 +646,13 @@ object TaskMemoizeOnSuccessSuite extends BaseTestSuite {
...
 
-  testAsync("local.write.memoize works") { _ =>
+  testAsync("local.write.memoizeOnSuccess works") { _ =>
     import monix.execution.Scheduler.Implicits.global
     implicit val opts = Task.defaultOptions.enableLocalContextPropagation
 
     val task = for {
       local <- TaskLocal(0)
-      v1    <- (local.write(100).flatMap(_ => local.read)).memoizeOnSuccess
+      v1    <- local.write(100).flatMap(_ => local.read).memoizeOnSuccess
```
oleg-py: Just for the sake of inner peace, can you execute the memoized task two or more times to ensure that it doesn't change the local state after first time?

Avasil: I upgraded to
```scala
val task = for {
    local <- TaskLocal(0)
    memoizeTask = local.write(100).flatMap(_ => local.read).memoize
    v1    <- memoizeTask
    _     <- Task.shift
    v2    <- local.read
    _     <- local.write(200)
    _     <- memoizeTask
    _     <- Task.shift
    v3    <- local.read
} yield (v1, v2, v3)

for (v <- task.runToFutureOpt) yield {
    assertEquals(v, (100, 100, 200))
}
```

and seems to still work. :D 

Initially I changed this test to say that `memoize` + `local` don't work, updated scaladoc etc. and then to my surprise the test was failing because it actually still worked... I guess `restoreLocals` can still be useful if we want to customize locals for `Task`

PR #951, opened 2019-07-13
Add void to Task + Coeval

allantl:
I know we have this in cats. But its pretty annoying having to import just for `void`.


allantl:
Done! Please take a look again.

PR #953, opened 2019-07-18
Make default scheduler remove cancelled tasks

### LINKED ISSUE #912: Make default scheduler remove cancelled tasks
We need to look into this:

https://github.com/typelevel/cats-effect/pull/568/

As this might prevent some leaks with `timeout`.

Avasil:
Fixes #912

By extension it will also set this policy in many other builders.

### DIFF
--- a/monix-execution/jvm/src/main/scala/monix/execution/schedulers/Defaults.scala
+++ b/monix-execution/jvm/src/main/scala/monix/execution/schedulers/Defaults.scala
@@ -17,15 +17,19 @@
 
 package monix.execution.schedulers
 
-import java.util.concurrent.{Executors, ScheduledExecutorService}
+import java.util.concurrent.{ScheduledExecutorService, ScheduledThreadPoolExecutor}
+
 import monix.execution.UncaughtExceptionReporter
 
 private[schedulers] object Defaults {
   /**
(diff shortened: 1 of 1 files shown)


monix-execution/jvm/src/main/scala/monix/execution/schedulers/Defaults.scala
```diff
@@ -17,15 +17,19 @@
...
   /**
     * Internal. Provides the `Scheduler.DefaultScheduledExecutor` instance.
     */
-  lazy val scheduledExecutor: ScheduledExecutorService =
-    Executors.newSingleThreadScheduledExecutor(
-      ThreadFactoryBuilder("monix-scheduler", UncaughtExceptionReporter.default, daemonic = true)
-    )
+  lazy val scheduledExecutor: ScheduledExecutorService = {
+    val tp = new ScheduledThreadPoolExecutor(
+      1,
```
Avasil: `Cats-Effect` uses 2 threads, not sure what's the reasoning behind this

alexandru:
@oleg-py @Avasil no specific reason.

IMO the important thing is to ensure that scheduled tasks don't get executed on that scheduler.
We could use 2 threads as well.

PR #954, opened 2019-07-23
Add gather + wander n

### LINKED ISSUE #926: Bounded versions for Task.gather and Task.wander
Recently `Cats-Effect`  added [`parSequenceN` and `parTraverseN`](https://github.com/typelevel/cats-effect/pull/498/files) and it can be used with `Task` but it will use generic `parSequence` / `parTraverse` from cats which is much slower than hand-optimized version we have.

[ZIO also has this operator named as foreachParN](https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/ZIO.scala#L1734)

allantl:
Closes #926

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -3472,6 +3472,12 @@ object Task extends TaskInstancesLevel1 {
   def gather[A, M[X] <: Iterable[X]](in: M[Task[A]])(implicit bf: BuildFrom[M[Task[A]], A, M[A]]): Task[M[A]] =
     TaskGather[A, M](in, () => newBuilder(bf, in))
 
+  /**
+    * Similar to `gather`, but bounded by specified parallelism.
+    */
+  def gatherN[A](parallelism: Int)(in: Iterable[Task[A]]): Task[List[A]] =
+    TaskGatherN[A](parallelism, in)
+
   /** Given a `Iterable[A]` and a function `A => Task[B]`,
     * nondeterministically apply the function to each element of the collection
     * and return a task that will signal a collection of the results once all
@@ -3495,6 +3501,14 @@ object Task extends TaskInstancesLevel1 {
     */
   def wander[A, B, M[X] <: Iterable[X]](in: M[A])(f: A => Task[B])(implicit bf: BuildFrom[M[A], B, M[B]]): Task[M[B]] =
     Task.eval(in.map(f)).flatMap(col => TaskGather[B, M](col, () => newBuilder(bf, in)))
+
+  /**
+    * Similar to `wander`, but bounded by specified parallelism.
+    */
+  def wanderN[A, B](parallelism: Int)(in: Iterable[A])(f: A => Task[B]): Task[List[B]] =
+    Task.suspend {
+      TaskGatherN(parallelism, in.map(f))
+    }
 
   /** Processes the given collection of tasks in parallel and
     * nondeterministically gather the results without keeping the original
--- a/monix-eval/shared/src/main/scala/monix/eval/internal/TaskGatherN.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/internal/TaskGatherN.scala
@@ -1,0 +1,75 @@
+/*
+ * Copyright (c) 2014-2019 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.eval.internal
+
+import cats.effect.ExitCase
+import cats.effect.concurrent.Deferred
+import monix.catnap.ConcurrentQueue
+import monix.eval.Task
+
+private[eval] object TaskGatherN {
+
+  def apply[A](
+    parallelism: Int,
+    in: Iterable[Task[A]]
+  ): Task[List[A]] = {
+    val itemSize = in.size
+
+    if (itemSize == 0) {
+      Task.pure(List.empty)
+    } else if (itemSize == 1) {
+      in.head.map(List(_))
+    } else {
+      for {
+        error <- Deferred[Task, Throwable]
+        queue <- ConcurrentQueue.bounded[Task, (Deferred[Task, A], Task[A])](itemSize)
+        pairs <- Task.traverse(in.toList)(task => Deferred[Task, A].map(p => (p, task)))
+        _     <- queue.offerMany(pairs)
+        workers = Task.gather(List.fill(parallelism.min(itemSize)) {
+          queue.poll.flatMap {
+            case (p, task) =>
+              task.redeemWith(
+                err => error.complete(err).attempt >> Task.raiseError(err),
+                p.complete
+              )
+          }.loopForever.start
+        })
+        res <- workers.bracketCase { _ =>
+          Task
+            .race(
+              error.get,
+              Task.sequence(pairs.map(_._1.get))
+            )
+            .flatMap {
+              case Left(err) =>
+                Task.raiseError(err)
+
+              case Right(values) =>
+                Task.pure(values)
+            }
+        } {
+          case (fiber, exit) =>
+            exit match {
+              case ExitCase.Completed => Task.unit
+              case _ => Task.traverse(fiber)(_.cancel).void
+            }
+        }
+      } yield res
+    }
+  }
+}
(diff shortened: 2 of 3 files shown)


monix-eval/shared/src/test/scala/monix/eval/TaskGatherNSuite.scala
```diff
@@ -0,0 +1,107 @@
...
+      Task.evalAsync(3).delayExecution(3.seconds))
+    val f = Task.gatherN(1)(seq).runToFuture
+
+    s.tick()
+    assertEquals(f.value, None)
+    s.tick(1.second)
+    assertEquals(f.value, None)
+
+    f.cancel()
+    s.tick(1.second)
```
oleg-py: That's actually not checking anything. The `f` can only complete after 6 seconds. I recommend having an effect in `doOnCancel` on 1st Task which you can actually check and then effect on 2nd Task which you can see _not_ happening even after ticking for `1.day`

monix-eval/shared/src/test/scala/monix/eval/TaskGatherNSuite.scala
```diff
@@ -0,0 +1,107 @@
...
+
+object TaskGatherNSuite extends BaseTestSuite {
+
+  test("Task.gatherN should execute in parallel for async tasks") { implicit s =>
+    val seq = Seq(
+      Task.evalAsync(1).delayExecution(1.second),
+      Task.evalAsync(2).delayExecution(2.seconds),
+      Task.evalAsync(3).delayExecution(3.seconds)
+    )
+    val f = Task.gatherN(1)(seq).runToFuture
```
oleg-py: - This isn't actually checking parallelism with N = 1
- It should execute in parallel even if Tasks aren't async to begin with.
- `delayExecution` already makes things async, btw.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -3486,6 +3493,13 @@ object Task extends TaskInstancesLevel1 {
   def wander[A, B, M[X] <: Iterable[X]](in: M[A])(f: A => Task[B])(implicit bf: BuildFrom[M[A], B, M[B]]): Task[M[B]] =
     Task.eval(in.map(f)).flatMap(col => TaskGather[B, M](col, () => newBuilder(bf, in)))
 
+  /**
+    * Similar to `wander`, but bounded by specified parallelism.
+    */
+  def wanderN[A, B, M[X] <: Iterable[X]](parallelism: Int)(in: M[A])(f: A => Task[B])(
+    implicit bf: BuildFrom[M[A], B, M[B]]): Task[M[B]] =
+    Task.eval(in.map(f)).flatMap(col => TaskGatherN[B, M](parallelism, col, () => newBuilder(bf, in)))
```
oleg-py: If you want to catch errors thrown by `f` call, using `Task.suspend` could be clearer and would allocate a bit fewer things. I.e.:
```
Task.suspend {
  TaskGatherN[B, M](parallelism, in.map(f), () => newBuilder(bf, in))
}
```
just leave a comment if catching from `f` is what you intend to do.

allantl:
Thanks for the comments. I've updated the tests btw.

oleg-py:
Good job, now I can see that it does what it says on the cover 👍

allantl:
Actually, I just did a quick testing comparing my implementation in the PR vs 

```
  def wanderN[T, U](parallelism: Int)(items: Seq[T])(f: => T => Task[U]): Task[Seq[U]] =
    for {
      queue <- ConcurrentQueue.bounded[Task, (Deferred[Task, U], T)](items.length)
      pairs <- Task.traverse(items)(item => Deferred[Task, U].map(p => (p, item)))
      _ <- queue.offerMany(pairs)
      fibers <- Task.wander(0.until(parallelism).toList) { _ =>
        queue.poll.flatMap { case (p, a) =>
          f(a).flatMap(p.complete)
        }.loopForever.start
      }
      res <- Task.sequence(pairs.map(_._1.get)).guarantee(
        Task.traverse(fibers)(fiber => Task.eval(fiber.cancel)).void
      )
    } yield res
```
```
  def wanderN[T, U](parallelism: Int)(items: Seq[T])(f: => T => Task[U]): Task[List[U]] =
    Observable.fromIterable(items).mapParallelOrdered(parallelism)(f).toListL
```

The implementation in the PR seems much slower. Probably, I'm doing something not efficient. Will need to investigate.

Avasil:
First off all, great job @allantl :D 

> The implementation in the PR seems much slower. Probably, I'm doing something not efficient. Will need to investigate.

What's the difference on the benchmarks?

Unfortunately I'm not able to take a closer look/experiment (benchmarking it with profiler on sounds fun) with the implementation right now but I have two random ideas that could work and improve performance:

1. We could probably start first `n` tasks without asking for any permit
2. Instead of using `AsyncSemaphore`, we could run `n` tasks and then each time one of the tasks is finished, we start a new one.

allantl:
@Avasil 👍 I'm on vacation currently. I'll take a look when I get back.

allantl:
@Avasil I've updated the PR following your suggestion no 2, so no more async sempahore.

Here's the benchmark [(source)](https://github.com/allantl/monix-benchmark/blob/master/src/main/scala/benchmark/GatherNBenchmark.scala):
```
Benchmark                                                   Cnt    Score    Error  Units

// May run out of of memory if list is very big
GatherNBenchmark.catsEffectParTraverseNBenchmark            10  116.799 ± 14.034  ops/s

GatherNBenchmark.manualImplementationWanderNBenchmark       10   56.101 ±  0.743  ops/s
GatherNBenchmark.observableMapParallelOrderedBenchmark      10   15.459 ±  0.450  ops/s
GatherNBenchmark.observableMapParallelUnorderedBenchmark    10   39.675 ±  0.866  ops/s

// I didn't handle error/cancellation scenario on this one
GatherNBenchmark.wanderNConcurrentQueueBenchmark            10  357.297 ±  4.390  ops/s

// May run out of of memory if list is very big
GatherNBenchmark.wanderNMonixSemaphoreBenchmark             10    6.022 ±  0.132  ops/s
```

Avasil:
Thank you for your investigation @allantl !

It's pretty surprising to me but I guess in this situation it's not worth to go with low-level solution. We can always revisit it later if we want to optimize.

BTW - any reason why `wanderNSemaphore` implementation is different than `parTraverseN` (other than parTraverse => wander)?  I think it could be responsible for such a score

allantl:
The difference is `cats.effect.concurrent.Sempahore` vs `import monix.catnap.Semaphore`

allantl:
I've updated the implementation to use concurrent queue instead. Benchmark is 

```
Benchmark                                               Mode  Cnt    Score   Error  Units
GatherNBenchmark.manualImplementationWanderNBenchmark  thrpt   10  344.197 ± 0.796  ops/s
``

So, around 3x faster compared to cats effect par traverse

File: monix-eval/shared/src/main/scala/monix/eval/internal/TaskGatherN.scala [hunk 4]
```diff
@@ -0,0 +1,69 @@
...
+  ): Task[List[A]] = {
+    val itemSize = in.size
+
+    if (itemSize == 0) {
+      Task.pure(List.empty)
+    } else if (itemSize == 1) {
+      in.head.map(List(_))
+    } else {
+      for {
+        error <- Deferred[Task, Option[Throwable]]
```
Avasil: I think we could have `Deferred[Task, Throwable]` directly

monix-eval/shared/src/main/scala/monix/eval/internal/TaskGatherN.scala
```diff
@@ -0,0 +1,69 @@
...
+          )
+          .flatMap {
+            case Left(err) =>
+              Task.raiseError(err.get)
+
+            case Right(values) =>
+              Task.pure(values)
+          }
+          .guarantee(
+            Task.traverse(fiber)(_.cancel).void
```
Avasil: We don't have to cancel everything on success, we could use `guaranteeCase` to only act on `cancelation` / `error`

monix-eval/shared/src/main/scala/monix/eval/internal/TaskGatherN.scala
```diff
@@ -0,0 +1,69 @@
...
+        _     <- queue.offerMany(pairs)
+        fiber <- Task.gather(List.fill(parallelism.min(itemSize)) {
+          queue.poll.flatMap {
+            case (p, task) =>
+              task.redeemWith(
+                err => error.complete(Some(err)).attempt >> Task.raiseError(err),
+                p.complete
+              )
+          }.loopForever.start
+        })
```
Avasil: I think if we cancel entire `Task.gatherN` at this point, it can get dirty

If we turn `AlwaysAsyncExecution` and `enableAutoCancelableRunLoops` then any `flatMap` could be canceled

I think `bracketCase` could solve it. `acquire` would be this `Task.gather`, `use` as `Task.race` and `release` instead of `guarantee`. I will try to think about good test for it later

allantl:
@Avasil I've updated the PR. I have one question, is cancellation needed when it exits on failure?

Avasil:
> @Avasil I've updated the PR. I have one question, is cancellation needed when it exits on failure?

Yes, we should return the failure and cancel the rest, otherwise it would just leak cpu resources

And thank you! I will play with it in a few hours. We plan to do a `3.0.0` release very soon (in ~ 10 hours) and hopefully we can include it!

Avasil:
Okay, I think it is correct, I have an idea for small improvement but I'll do a PR myself in a moment to deliver it for `3.0.0`

Good job @allantl :)

PR #960, opened 2019-07-27
Scheduler.features

alexandru:
The primary purpose of this PR is to make `Observable` play well with `Local` and `TaskLocal` when used in combination with `Task` (e.g. in `Observable.mapEval`).

For this purpose it adds a `def features` property to `Scheduler`.

This helps with discovery of what the underlying scheduler supports, because `instanceOf` checks are problematic to say the least due to all the wrapping. Features that can be exposed by schedulers currently:

- `BATCHING` — special handling of `TrampolinedRunnable`, for schedulers that inherit from `BatchingScheduler`
- `TRACING` — handling of `Local`, implemented by `TracingScheduler` and `TracingSchedulerService`

With a `Scheduler` instance you might be tempted to do this:

```scala
// WRONG
if (scheduler.isInstanceOf[TracingScheduler]) {
  //...
}
```

However this is wrong for one because `TracingSchedulerService` does not inherit from `TracingScheduler` and also because people might implement their own `Scheduler` wrappers and implementations that might use an underlying `TracingScheduler`, which would support `Local`, but which would fail an `isInstanceOf[TracingScheduler]`.

With a `Scheduler` instance you can now do this:

```scala
if (scheduler.features.contains(Scheduler.TRACING)) {
  // ...
}
```

Using this, `Task.runAsync` can now activate the "local context propagation" in case the `Scheduler` is a `TracingScheduler`:

```scala
implicit val scheduler = TracingScheduler(global)

// This will have the "local context propagation" active
task.runAsync
```

This helps in `Observable`, because `Observable`'s `mapEval` does not take the local context propagation into account:

```scala
Observable(1, 2, 3).mapEval(task).completedL.runAsyncOpt
```

Currently the tasks evaluated inside of `mapEval` won't take any "local context propagation" option enabled via `runAsyncOpt`, this being IMO an API bug.


alexandru:
@mdedetrich would love your input as well.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -4685,8 +4708,11 @@ private[eval] abstract class TaskInstancesLevel0 extends TaskParallelNewtype {
     */
   implicit def catsEffect(
     implicit s: Scheduler,
-    opts: Task.Options = Task.defaultOptions): CatsConcurrentEffectForTask =
-    new CatsConcurrentEffectForTask
+    opts: Task.Options = Task.defaultOptions): CatsConcurrentEffectForTask = {
+
+    val opts2 = opts.withSchedulerFeatures
```
Avasil: It might be redundant because running `Task` (e.g. in `runAsync`) checks it anyway

alexandru: Ah, good point.

alexandru: Fixed.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -2597,7 +2597,7 @@ abstract class Observable[+A] extends Serializable { self =>
     *        throws an error.
     */
   final def onErrorRecover[B >: A](pf: PartialFunction[Throwable, B]): Observable[B] =
-    onErrorHandleWith(ex => (pf.andThen(Observable.now(_))).applyOrElse(ex, Observable.raiseError _))
+    onErrorHandleWith(ex => (pf.andThen(Observable.now)).applyOrElse(ex, Observable.raiseError))
```
Avasil: AFAIK it won't compile with latest `Scala 2.13` but it's fine unless you will have extra changes to do, I can fix it in PR updating to `2.13.0`

mdedetrich:
@alexandru Will look at it in the next couple of days, I fell sick last week so wasn't really available

mdedetrich:
I had a brief look at this PR last week, and its unrelated to the issues I am experiencing. We are not using Observable at all.

I definitely agreed with the direction of the PR though, for library authors its great to see what features the `Scheduler` has reliably (and I think I could use this feature in my monix-mdc/monix-opentracing to warn users if they don't have `TracingScheduler` enabled)

alexandru:
@mdedetrich this PR is also enabling `TaskLocal` to work in the presence of a `TracingScheduler` via `runAsync`. So if you're using a `TracingScheduler`, you no longer need to do a `runAsyncOpt` with the right `Task.Options`.

PR #965, opened 2019-07-28
Fix #938 — fixes resource handling in Iterant.repeat, adds Iterant.retryIfEmpty

### LINKED ISSUE #938: Iterant.resource(...).repeat doesn't release any resources until the very end
I'm not sure if this is intentional, but if I construct an Iterant via the `.resource` helper, then `.repeat` it, it will repeatedly allocate the resource, then when the stream is done, it will deallocate all of the resources. I expected it to release the allocated resource each time a new "repeat" iteration started.

A minimal example:

```scala
val nums = Iterant.resource(Task{ println("Acquire") })(_ => Task{ println("Release") })
  .flatMap(_ => Iterant[Task].of(1,2,3)
```

If I run `nums.repeat.take(9).foreach(println)` I get:
```text
Acquire
1
2
3
Acquire
1
2
3
Acquire
1
2
3
Release
Release
Release
```

But if I run `(nums ++ nums ++ nums).foreach(println)` I get what I was expecting from the "repeat" version:
```text
Acquire
1
2
3
Release
Acquire
1
2
3
Release
Acquire
1
2
3
Release
```

I'm using Monix 3.0.0-RC2

alexandru:
See #938 for the bug description.

This PR fixes `iterant.repeat` to handle resources (`Scope`) correctly.

Because `repeat` returns empty on empty streams, this PR also adds a new `retryIfEmpty` operator:

```scala
import scala.util.Random

val stream = Iterant[Coeval].suspend(Coeval {
  val nr = Random.nextInt()
  if (nr % 10 != 0)
    Iterant[Coeval].empty[Int]
  else
    Iterant[Coeval].of(1, 2, 3)
})

stream.retryIfEmpty(None).repeat
```

### DIFF
--- a/monix-tail/shared/src/test/scala/monix/tail/IterantRepeatSuite.scala
+++ b/monix-tail/shared/src/test/scala/monix/tail/IterantRepeatSuite.scala
@@ -21,6 +21,7 @@ import cats.laws._
 import cats.laws.discipline._
 import cats.syntax.eq._
 import monix.eval.{Coeval, Task}
+import monix.execution.atomic.Atomic
 import monix.execution.exceptions.DummyException
 import monix.execution.internal.Platform
 import monix.tail.batches.{Batch, BatchCursor}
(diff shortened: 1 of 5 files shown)


monix-tail/shared/src/test/scala/monix/tail/IterantRepeatSuite.scala
```diff
@@ -113,6 +114,18 @@ object IterantRepeatSuite extends BaseTestSuite {
     assertEquals(source4.repeat.toListL.value(), List.empty[Int])
   }
 
+  test("Iterant.repeat discards Scopes properly") { implicit s =>
+    val acquired = Atomic(0)
+    val sum = Iterant
+      .resource(Coeval(acquired.incrementAndGet()))(_ => Coeval(acquired.decrement()))
```
oleg-py: This actually won't discriminate vs old behavior where finalizers are executed at the end. You need to check that list is `List(1, 1, 1, 1, 1, 1...)` at least

alexandru: It does discriminate, because with the old behavior you get `1, 2, 3, ... 10` as elements, so you get the total sum to be `55` instead of `10`.

I can test with a list of course, instead of a sum.

oleg-py: Right, disregard that 😓

PR #968, opened 2019-07-31
Add cats compatible combinators *> (productR) and <* (productL) for Task and Coeval

### LINKED ISSUE #929: Task.*> and <* combinator

valenterry:
Closes #929


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -1186,6 +1186,45 @@ sealed abstract class Task[+A] extends Serializable {
...
+    * {{{
+    *   val combined = Task{println("first"); "first"} productR Task{println("second"); "second"}
+    *   // Prints "first" and then "second"
+    *   // Result value will be "second"
+    * }}}
+    *
+    * As this method is strict, it can lead to an infinite loop / stack overflow for self-referring tasks.
+    * @see [[>>]] for the version with a non-strict parameter
+    */
+  final def productR[B](tb: Task[B]): Task[B] =
```
Avasil: Thank you @valenterry !
Though I don't think we need `productR` / `productL`, just `*>` and `*<` should be enough, I would be surprised if there is anyone using these with `Task`. :D

oleg-py: Agreed

valenterry: I thought it might be nice to stick to how cats does it and keep the code compatible - I also like to have names for methods that can be pronounced. But I can also see that if no one needs these methods, they just make maintenance harder.

oleg-py: Since we already have it in cats which monix requires anyway, it's only point is to avoid having to import syntax for most commonly needed operations. In my experience, `productL`/`R` aren't commonly used

Avasil: From our perspective, we just want to provide the most essential operators to avoid too much clutter in the API. This syntax is available from Cats anyway and removing things from a library is a hard thing so we need to be careful about functions we choose to include. 
I think if someone is working with `Task` directly and wants to use something with a name that person will just go for `flatMap`. I think we could always add how it's pronounced (but I think `*>` is called left or right shark :D ) to the scaladoc

alexandru:
@valenterry the following changes are needed for this PR to get merged:

1. remove `productR` and `productL`, since this PR is strictly about syntax and people wanting those methods can just `import cats.implicits._`
2. add `*>` and `<*` to `Coeval` as well, as we need to ensure consistency
3. add some simple tests

valenterry:
> 3. add some simple tests

What tests do you have in mind?

Avasil:
> > 1. add some simple tests
> 
> What tests do you have in mind?

Not convinced about the value of a test here but I suppose sanity check like the following:

```scala
test("fa *> fb <-> fa.flatMap(_ => fb)") { implicit s =>
  check2 { (fa: Task[Int], fb: Task[Int]) =>
    fa *> fb <-> fa.flatMap(_ => fb)
  }
}
```

should do the trick

jozic:
is any help needed here? would love to see this in next release
@valenterry @Avasil

valenterry:
> is any help needed here? would love to see this in next release
> @valenterry @Avasil

Thank you for the ping! I added the property tests for `Task` and `Coeval` basked on `Int` only (seems to be the standard in the other tests).

valenterry:
The tests fail, but it seems unrelated to my changes 🤔

Avasil:
Thanks, I have restarted the job, I will take a look in a few days

jozic:
@Avasil can this make it to 3.2.0?

Avasil:
Sure!

Thanks for PR @valenterry

Issue #955, opened 2019-07-23
Add Observable.unfold


Avasil:
```scala
def unfold[S, A](seed: => S)(f: S => Option[(A, S)]): Observable[A]
```

Similar to [fromStateAction](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/StateActionObservable.scala) but stops as soon as `None` is generated

oleg-py:
Why is `seed` the 2nd parameter list here? Wouldn't that cause worse inference, as `S` would fail to infer?

Avasil:
Eh, I'd prefer it to be the first parameter but `fromStateAction` has it in this order... But in everywhere else it's the way you describe.

Maybe we can deprecate old methods and switch the order?

alexandru:
Even though the ordering doesn't play well with Scala's type inference, the `seed` as the second parameter is the natural order in `fromStateAction`:

```scala
val randomNumbers = Observable.fromStateAction { (state: Long) =>
  val state2 = (state * 0X5DEECE66DL + 0XBL) & 0XFFFFFFFFFFFFL
  val n = (state2 >>> 16).toInt
  (n, state2)
} _
```

You just need an explicit type in that function to do this. What you gain however is that `randomNumbers` is a function and that seed is especially important ...

```scala
Task.deferAction { sc =>
  randomNumbers(sc.clockRealtime(MILLISECONDS))
}
```

In other words this isn't necessarily like `foldLeft` where you're always dealing with the same empty / zero element.

Avasil:
Thanks for explanation @alexandru 
Makes sense to me, I withdraw my deprecation idea

Though `unfold` itself should follow the order of `fold` and other streaming libraries, right?

alexandru:
Not sure. What are other libraries doing?

Avasil:
@alexandru 

Akka streams: 
[def unfold[S, E](s: S)(f: S => Option[(S, E)]): Source[E, NotUsed]
](https://doc.akka.io/docs/akka/current/stream/operators/Source/unfold.html)

fs2:
```
def unfold[F[x] >: Pure[x], S, O](s: S)(f: S => Option[(O, S)]): Stream[F, O]
```

ZIO: 
```
final def unfold[S, A](s: S)(f0: S => Option[(A, S)]): Stream[Nothing, A]
```

alexandru:
@Avasil thanks for looking into it.

Then do the state as the first param.

[...]

PR #970, opened 2019-08-02
Observable.unfold

pk044:
Solves #955 

i have a few doubts though:
- i'm afraid the code is duplicated, it differs very little from `fromStateAction`. i was thinking of extracting the common parts somewhere and try to reuse them in both functions - not to mention that more unfold-based functions are planned to be implemented in Monix (unfoldEval, unfoldEvalF), is there perhaps a generic way to handle this?
- i tried to write a property based test, i'm not too familiar with it and I kind of gave up after having the test go out of memory :)

```
  test("expected not to hang and throw OOM errors but it does") { implicit s =>
    check3 { (seed: Int, f: Int => Option[(Int, Int)], i: Int) =>
      val n = i % (recommendedBatchSize * 2)

      val received = Observable.unfold(seed)(f).takeWhile(_ < n).toListL
      val expected = Task.now((seed to n).toList)

      received <-> expected
    }
  }
```
after dabbling with it for a little while it looks like it hangs because the size of the generated list (`seed to n toList`) might be a little too big. I think I'm supposed to have that list in the parameters, and come up with some sort of law so that ScalaCheck would always generate the expected list, instead of letting me doing it myself. (not sure if that was clear enough)

do you guys perhaps have any suggestions for any of the above? thanks!


monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/UnfoldObservable.scala
```diff
@@ -0,0 +1,101 @@
...
+
+    @tailrec
+    def fastLoop(syncIndex: Int): Unit = {
+      val ack = try {
+        f(seed) match {
+          case Some((nextA, newState)) =>
+            this.seed = newState
+            o.onNext(nextA)
+          case None =>
+            Stop
```
Avasil: We should also emit `o.onComplete()` to let know downstream that `Observable` is ending

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/UnfoldSuite.scala
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.reactive.internal.operators
+
+import monix.execution.Ack.Continue
+import monix.reactive.{BaseTestSuite, Observable}
+object UnfoldSuite extends BaseTestSuite {
```
Avasil: We could copy few tests from [StateActionObservableSuite](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/test/scala/monix/reactive/internal/builders/StateActionObservableSuite.scala) - especially to do things like:

```scala
val cancelable = Observable
    .fromStateAction(int)(s.clockMonotonic(MILLISECONDS))
    .unsafeSubscribeFn(new Subscriber[Int] {
      implicit val scheduler = s
      def onNext(elem: Int) = {
        sum += 1
        Continue
      }

      def onComplete() = wasCompleted = true
      def onError(ex: Throwable) = wasCompleted = true
})
```

to see if `Observable` completed. I think it could fail currently. Alternatively we can call methods like `sumL` or `completedL` to complete it in nicer way and execute as a `Task`. `subscribe` just processes it asynchronously in the background

Also I'd suggest naming the suite to `UnfoldObservableSuite` - we have also `Iterant` and if we add `unfold` for it then it can be a bit confusing to search for

Avasil:
Thank you @pk044 - great job overall! I've left some comments.

> * i'm afraid the code is duplicated, it differs very little from `fromStateAction`. i was thinking of extracting the common parts somewhere and try to reuse them in both functions - not to mention that more unfold-based functions are planned to be implemented in Monix (unfoldEval, unfoldEvalF), is there perhaps a generic way to handle this?

It's fine to have duplicated code, we often do it in purpose for a little gain in efficiency. I wouldn't do it in a "normal" code base but we have an excuse. :)

> * i tried to write a property based test, i'm not too familiar with it and I kind of gave up after having the test go out of memory :)
>
> ```
>   test("expected not to hang and throw OOM errors but it does") { implicit s =>
>     check3 { (seed: Int, f: Int => Option[(Int, Int)], i: Int) =>
>       val n = i % (recommendedBatchSize * 2)
> 
>       val received = Observable.unfold(seed)(f).takeWhile(_ < n).toListL
>       val expected = Task.now((seed to n).toList)
> 
>       received <-> expected
>     }
>   }
> ```

I think `seed to n` could be potentially -2147483648 to n

BTW It would be cool if we can come up with a property that tests `Observable.unfold(seed)(f).toListL <-> Observable.fromStateAction(f)(seed).takeWhile(p).toListL` instead

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -5197,6 +5197,14 @@ object Observable extends ObservableDeprecatedBuilders {
   def fromStateAction[S, A](f: S => (A, S))(seed: => S): Observable[A] =
     new builders.StateActionObservable(seed, f)
 
+  /** Given an initial state and a generator function that produces the
```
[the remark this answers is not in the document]
pk044: could you take a peek again and see if it provides enough details?

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/UnfoldSuite.scala
pk044: i added the test with `subscribe` for now, I tried to write it like this:
  test("should be cancelable") { implicit s =>

```
    val cancelable = Observable
      .unfold(s.clockMonotonic(MILLISECONDS))(intOption)
        .sumL
        .runToFuture

    cancelable.cancel()

    s.tick()

    assertEquals(cancelable.value, Some(Success(s.executionModel.recommendedBatchSize * 2)))
  }
```

but it returns None - i need to dig deeper and find out why

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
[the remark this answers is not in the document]
Avasil: Perfect, thank you!

pk044:
@Avasil thanks! i'll see what I can do about the property test :)

pk044:
@Avasil i'm not too familiar with property-based testing yet, I came up with this so far:

```scala
  test("unfold and fromStateAction lists should be equal given generated inputs") { implicit s =>
    check {
      forAll(Gen.choose(0, Platform.recommendedBatchSize * 2), Gen.choose(0, Platform.recommendedBatchSize * 2)) {
        (seed, max) =>
          val f: Int => Option[(Int, Int)] = i => if (i < max) Some((i, i + 1)) else None
          val f2: Int => (Int, Int) = i => (i, i + 1)

          Observable.unfold(seed)(f).toListL <-> Observable.fromStateAction(f2)(seed).takeWhile(_ < max).toListL
      }
    }
  }
```
is it what we wanted more or less?

monix-reactive/shared/src/test/scala/monix/reactive/internal/builders/UnfoldObservableSuite.scala
```diff
@@ -0,0 +1,108 @@
...
+
+    s.tick()
+
+    assertEquals(sum, s.executionModel.recommendedBatchSize * 2)
+    assert(!wasCompleted)
+  }
+
+  test("unfold and fromStateAction results should be equal given generated inputs") { implicit s =>
+    check {
+      forAll(Gen.choose(0, Platform.recommendedBatchSize * 2), Gen.choose(0, Platform.recommendedBatchSize * 2)) {
```
Avasil: I think the test is good but I'd recommend doing just `check2` and `val n = i % (recommendedBatchSize * 2)` like in many `Iterant` tests, it's a bit less noise when reading tests

pk044: so basically something like

```
    check2 { (s: Int, i: Int) =>
          val seed = s % (recommendedBatchSize * 2)
          val n = i % (recommendedBatchSize * 2)
```
?

Avasil: I feel like `seed` should be fine but there's no harm in doing it this way 👍

PR #971, opened 2019-08-02
Callback tryOnSuccess/tryOnFailure

alexandru:
This is a continuation of #917.

Thread-safe implementations of `Callback` (that protect the `Callback`'s contract) are supposed to throw `CallbackCalledMultipleTimesException` in case the result was signaled more than once.

The callbacks returned the `Callback.safe`, `Callback.trampoline` and `Callback.forked` wrappers are now all thread-safe. The callback injected in `Task.create` (and variants) is thread-safe as well.

Also added 2 methods on `Callback`: `tryOnSuccess` and `tryOnFailure`. These forward by default the call to `onSuccess` and `onError`. If these throw, then the exception gets caught in the `try*` versions.


monix-eval/shared/src/main/scala/monix/eval/internal/TaskShift.scala
```diff
@@ -68,9 +68,7 @@ private[eval] object TaskShift {
         })
       } catch {
         case e: RejectedExecutionException =>
-          Callback
-            .trampolined(cb)(context.scheduler)
-            .onError(e)
+          cb.onError(e)
```
Avasil: why is it no longer necessary?

alexandru: I made `Callback.trampolined()` thread-safe too and I don't want to use it just for signaling errors in such cases. But point taken, I've reintroduced the trampolined execution here.

alexandru:
This one is ready for merging.

I added tests and changed the behavior of the `Task.async` builders to also report errors, thrown in the registration function, via the callback. We couldn't do that before because the callback wasn't safe.

Avasil:
Thanks @alexandru , we'll get it in for the release

Avasil:
I'm merging it @oleg-py but if you have any objections/suggestions we can patch it in next PR

PR #972, opened 2019-08-02
 Rename `Task.forkAndForget` to `Task.startAndForget`

### LINKED ISSUE #877: Make a RT "fire and forget" function
So we kind of have this problem in our services where we do a lot of "fire and forget" tasks. These are tasks that follow these properties

* Do not block, even in the resulting computation, i.e.
```scala
for {
  _ <- doSomething
  _ <- fireAndForgetTask.executeAsync
  _ <- anotherFireAndForgetTask.executeAsync
} yield ()
```

In this example, by fire and forget we mean that this for expression will immediately exist after `doSomething`, we do not wait for either `fireAndForgetTask` or `anotherFireAndForgetTask` to complete, i.e. the for comprehension will return immediately after `doSomething` is finished.

* Do not prevent the current computation in case of any error (this can be currently already be done with the various recover functions)

The issue is I don't think this is currently possible without having to unsafely run a `Task` into a `Future` (which is what we currently do, i.e. `runToFuture`). This has actually kind of caused incidents because said functions are not RT (so they executed immediately when they weren't meant to since they weren't suspended)

Is it possible to create a RT "fire and forget" style function. Note this may also already be possible, but from all of the reading online people say the only real way to do true "fire and forget" is to use `runToFuture` (in which case it may be a documentation issue and/or we may also need to create a handy helper function).

Avasil:
Fixes #877


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -2425,6 +2425,20 @@ sealed abstract class Task[+A] extends Serializable {
     */
   final def void: Task[Unit] =
     this.map(_ => ())
+
+  // Deprecation directly in `Task` to avoid binary breakage just before the release
```
Avasil: I was getting `DirectMissingMethodProblem` if I put it in `Deprecations` syntax so wanted to be safe

alexandru:
👍

alexandru:
@Avasil from a code organization perspective, I don't like leaving `@deprecated` methods on the classes themselves.

One idea would be to create a new trait for deprecated methods in the `TaskDeprecated` object and put this method there.

Avasil:
@alexandru 👍

PR #973, opened 2019-08-06
Fix local isolate

### LINKED ISSUE #966: Local.isolate sometimes doesn't isolate properly in between async boundaries
See the comment in the snippet:

```scala
import monix.execution.schedulers.TracingScheduler

import scala.concurrent.{Await, ExecutionContext, Future}

object Main extends App {
  implicit val s = TracingScheduler(ExecutionContext.global)

  val local = Local(0)

//  Sometimes fails with:
//  scala-execution-context-global-306-0
//  scala-execution-context-global-286-50
//  scala-execution-context-global-286-isolate1-50
//  scala-execution-context-global-306-isolate2-50
//  scala-execution-context-global-286-100
  val f1 = for {
    _ <- Future{
      println(Thread.currentThread.getName + s"-${local()}")
      assert(local() == 0)
      local := 50
    }
    _ <- Future{
      println(Thread.currentThread.getName + s"-${local()}")
      assert(local() == 50)
    }
    _ <- Local.isolate{
      println(Thread.currentThread.getName + s"-isolate1-${local()}")
      Future  {
          println(Thread.currentThread.getName + s"-isolate2-${local()}")
          assert(local() == 50)
          local := 100
      }
    }
    _ <- Future{
      println(Thread.currentThread.getName + s"-${local()}")
      assert(local() == 50)
    }
  } yield ()

  Await.result(f1, Duration.Inf)
}
```
cc @alexandru @oleg-py

Avasil:
Fixes #966 

Not finished but might be easier to discuss


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -580,9 +580,11 @@ sealed abstract class Task[+A] extends Serializable with TaskDeprecated.BinCompa
...
   def runToFutureOpt(implicit s: Scheduler, opts: Options): CancelableFuture[A] = {
     val opts2 = opts.withSchedulerFeatures
-    Local.bindCurrentIf(opts2.localContextPropagation) {
-      TaskRunLoop.startFuture(this, s, opts2)
-    }
+    Local
+      .bindCurrentAsyncIf(opts2.localContextPropagation) {
+        TaskRunLoop.startFuture(this, s, opts2)
+      }
+      .asInstanceOf[CancelableFuture[A]]
```
Avasil: I hope there is a way to do it without it

monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
```diff
@@ -268,4 +268,88 @@ object TaskLocalSuite extends SimpleTestSuite {
...
+        Task.evalAsync(local := 100).flatMap(_ => Task.raiseError(DummyException("boom")))
+      }.attempt
+      v <- Task.evalAsync(local())
+      _ <- Task.now(assertEquals(v, 50))
+    } yield ()
+
+    test.runToFutureOpt
+  }
+
+  testAsync("TaskLocal.isolate should properly isolate during async boundaries on cancelation") {
```
Avasil: This test fails currently

monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
```diff
@@ -268,4 +268,88 @@ object TaskLocalSuite extends SimpleTestSuite {
...
+        Task.evalAsync(local := 100).start.flatMap(_.cancel)
+      }
+      v <- Task.evalAsync(local())
+      _ <- Task.now(assertEquals(v, 50))
+    } yield ()
+
+    test.runToFutureOpt
+  }
+
+  testAsync("Running task isolates locals") {
```
Avasil: This one fails too even for `runToFuture`, not sure why

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -78,12 +95,18 @@ object Local {
...
+  /** If `b` evaluates to `true`, execute a synchronous block of code using a current
     * state of `Local.Context` and restore the current state when complete.
     */
   def bindCurrentIf[R](b: => Boolean)(f: => R): R =
     macro Macros.localLetCurrentIf
 
+  /** If `b` evaluates to `true`, execute an asynchronous block of code using a current
+    * state of `Local.Context` and restore the current state when complete.
+    */
+  def bindCurrentAsyncIf[R](b: => Boolean)(f: => Future[R]): Future[R] =
```
Avasil: couldn't just overload because of `(b: => Boolean)` I suppose

monix-execution/shared/src/test/scala/monix/execution/misc/LocalSuite.scala
```diff
@@ -73,8 +75,10 @@ object LocalSuite extends SimpleTestSuite {
     local2 := 100
 
     val f = Local.isolate {
-      local1 := 100
-      Future(local1.get + local2.get)
+      Future {
```
Avasil: with current changes
```scala
Local.isolate {
  local1 := 100
  Future (...)
}
```
currently goes for `Future` variant but doesn't isolate properly :( unless we stick modification inside
```

File: monix-execution/shared/src/test/scala/monix/execution/misc/LocalSuite.scala [hunk 6]
```diff
@@ -73,8 +75,10 @@ object LocalSuite extends SimpleTestSuite {
...
+        local1 := 100
+        local1.get + local2.get
```
Avasil: it returns updated `local1` (100) + `local2` (999) from below unless we do something use `local2` here

Avasil: fixed both cases

monix-execution/shared/src/test/scala/monix/execution/misc/LocalSuite.scala
Avasil: fixed both cases

monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
Avasil: Small update:
```scala
val f = p.future

f.map(_ => {
    assert(local() == 0, s"received ${local()} != expected 0 in $method")
})
```

fails if `f` is not isolated

monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
Avasil: Fixed by not restoring locals in `TaskStart`

monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
Avasil: @oleg-py Is it a bug or feature? :D

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
oleg-py: There's actually no point in having `b` being by-name here. It will always be strictly evaluated.

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -60,12 +62,34 @@ object Local {
...
+    * changes outside.
+    */
+  def isolate[R](f: => Future[R]): Future[R] = {
+    // TODO: rewrite as macro
+    val prev = Local.getContext()
+    val current = Local.getContext().mkIsolated
+    Local.setContext(current)
+
+    try {
+      f.transform(result => {
```
oleg-py: Can't you use `Try[A] => Try[B]` overload here to have just one lambda?

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -294,8 +324,24 @@ final class Local[A](default: () => A) {
   def bind[R](value: A)(f: => R): R = {
     val parent = Local.getContext()
     Local.setContext(parent.bind(key, Some(value)))
-    try f
-    finally Local.setContext(parent)
+
+    f match {
+      case future: Future[_] =>
+        future
+          .transform(result => {
```
oleg-py: Same as above

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -304,13 +350,28 @@ final class Local[A](default: () => A) {
   def bindClear[R](f: => R): R = {
     val parent = Local.getContext()
     Local.setContext(parent.bind(key, None))
-    try f
-    finally Local.setContext(parent)
+
+    f match {
+      case future: Future[_] =>
+        future
+          .transform(result => {
```
oleg-py: Same

monix-eval/jvm/src/test/scala/monix/eval/TaskLocalJVMSuite.scala
```diff
@@ -174,4 +177,87 @@ object TaskLocalJVMSuite extends SimpleTestSuite {
...
+    def runAssertion(run: Task[Unit] => Any, method: String): Future[Unit] = {
+      val p = Promise[Unit]
+      val local = Local(0)
+      val task = Task.evalAsync(local := 50).guarantee(Task(p.success(())).void)
+
+      run(task)
+
+      // Future could still carry isolated local
+      // because it was created inside isolated block
+      val f = Local.isolate(p.future)
```
oleg-py: Does `isolate` actually do anything here?

monix-eval/shared/src/main/scala/monix/eval/Task.scala
oleg-py: I don't think there is. But `bind..If` is somewhat of internal method, we can make a version for `CancelableFuture` and make it private.

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
Avasil: Good point, I'll change it although it is still ambigious reference to overloeaded definition if we prefer overload.

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
Avasil: I think it's not available in `2.11`

monix-eval/jvm/src/test/scala/monix/eval/TaskLocalJVMSuite.scala
Avasil: Yes, otherwise it will carry the state of `Local` from the place of `Promise` completion. Not sure if it's a bug or a feature

monix-eval/shared/src/main/scala/monix/eval/Task.scala
Avasil: Hmm, maybe it could make sense to make `bindCurrentIf` private as well? It's basically `if p isolate(x) else x` so if anyone needs it - he can add it easily

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
oleg-py: Oh, right

monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
oleg-py: I'd say it's a quirk, I don't think it can be solved.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
oleg-py: Yes, that what I was thinking of :)

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
Avasil: I guess we can always create `Future.transform` in version specific modules to optimize it for `2.12` and `2.13` but not sure if it matters right now

monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
Avasil: Yeah, I tried but failed miserably :D I guess we can live with it. I'm not sure if it's even a use case we should support

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
Avasil: Actually, I think we have `transform` in `FutureUtils`, I'll try it

Avasil:
@oleg-py Thank you for the review, I updated PR:
- Removed duplicated test
- used `FutureUtils.transform` 
- changed `bindCurrentif` to `private[monix]`

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -60,12 +64,31 @@ object Local {
...
+  /** Execute a synchronous block of code without propagating any `Local.Context`
     * changes outside.
     */
   def isolate[R](f: => R): R =
     macro Macros.isolate
 
+  /** Execute an asynchronous block of code without propagating any `Local.Context`
+    * changes outside.
+    */
+  def isolate[R](f: => Future[R]): Future[R] = {
```
alexandru: I am concerned about this overload.

The old one is a macro, this one is not. Also I know that the JVM cannot differentiate between `R` and `Future[R]`, especially here because we have a by-name parameter which translates into a `Function0`.

I'm thinking that we should no longer use a macro here. I'm not convinced that we gain anything by it. Macros are fine for internal functions, were we don't want the overhead of that closure, but for user code it just makes things complicated.

So I'm thinking that maybe we should introduce a type class, because why not:

```scala
trait CanIsolate[R] {
  def isolate(f: => R): R
}

//...

object Local {
  def isolate[R : CanIsolate](f: => R): R =
    CanIsolate[R].isolate(f)
}
```

Given that what we had was a macro, I don't think this would violate any binary compatibility.

Not sure about how this should behave in the context of `Task` though. Or if it's a good idea. Just a suggestion.

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -294,8 +338,21 @@ final class Local[A](default: () => A) {
   def bind[R](value: A)(f: => R): R = {
     val parent = Local.getContext()
     Local.setContext(parent.bind(key, Some(value)))
-    try f
-    finally Local.setContext(parent)
+
+    f match {
```
alexandru: Note that this does not have the semantic of the original code — `f` is a by-name parameter and you're evaluating it twice.

Also the first time you're evaluating `f`, it isn't in the context of a try/catch statement. This means that if `f` throws an error, you're not restoring the context.

Also this pattern matching here is not future proof. It isn't idiomatic Scala so to speak. If we really need to discriminate between data types like this, then we should use implicits.

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
Avasil: I agree with macro, I meant to rewrite it to macro later but only because of consistency. I think we could safely remove it (and possibly reintroduce if it turns out it's needed). I know Oleg kept it for similar reasons

> So I'm thinking that maybe we should introduce a type class, because why not:
> ...

Hmm, I think it could work, we could have an instance for `Future` and then a lower priority instance for everything else. I can't tell if it's an improvement or not, to me it seems similar to having overload

> Not sure about how this should behave in the context of Task though. Or if it's a good idea. Just a suggestion.

`Task` has a different function for isolate on `TaskLocal` which does `bracket`. We could add an instance for normal `Local` in `Task` object I guess if it will be higher priority than default one for `Local` (not sure, needs to check)

WDYT @oleg-py ?

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
Avasil: > Note that this does not have the semantic of the original code — f is a by-name parameter and you're evaluating it twice.
> Also the first time you're evaluating f, it isn't in the context of a try/catch statement. This means that if f throws an error, you're not restoring the context.

Right, I will fix it

> Also this pattern matching here is not future proof. It isn't idiomatic Scala so to speak. If we really need to discriminate between data types like this, then we should use implicits.

What do you mean by "future proof" ?

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
oleg-py: I have no preference here. Typeclasses will work as well as `asInstanceOf`, really, unless you push it to properly isolate `Option[Future[Option[Future[A]]]]`. You'll also have to make sure it works properly with our `CancelableFuture`, that it doesn't fall back to lower priority instance.

> maybe we should introduce a type class, because why not

🤣

PR #976, opened 2019-08-09
Expand Task.gatherN and wanderN scaladocs

Avasil:
Just added more docs there and used Single Producer - Multi Consumer Queue instead


monix-eval/shared/src/main/scala/monix/eval/internal/TaskGatherN.scala
```diff
@@ -37,7 +38,8 @@ private[eval] object TaskGatherN {
     } else {
       for {
         error <- Deferred[Task, Throwable]
-        queue <- ConcurrentQueue.bounded[Task, (Deferred[Task, A], Task[A])](itemSize)
+        queue <- ConcurrentQueue
+          .withConfig[Task, (Deferred[Task, A], Task[A])](BufferCapacity(itemSize), ChannelType.SPMC)
```
Avasil: Improvement probably won't be even noticable but won't hurt either :D

PR #986, opened 2019-08-23
RestoreLocals in Bracket

Avasil:
CI will fail because of https://github.com/monix/monix/pull/985 but all other Local tests should pass


alexandru:
Dude, thanks a lot for trying hard to get this right.

PR #989, opened 2019-08-23
Observable.unfoldEval and Observable.unfoldEvalF

### LINKED ISSUE #956: Add Observable.unfoldEval
```scala
def unfoldEval[S, A](seed: => S)(f: S => Task[Option[(A, S)]]): Observable[A]
```

Similar to [fromAsyncStateAction](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/AsyncStateActionObservable.scala) but stops as soon as `None` is generated. `unfoldEvalF` working for `F[_]` would be useful too

pk044:
Closes #956


monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/UnfoldEvalObservable.scala
```diff
@@ -0,0 +1,72 @@
...
+      { ex =>
+        subscriber.onError(ex)
+        Task.unit
+      }, {
+        case Some((a, newState)) =>
+          Task.fromFuture(subscriber.onNext(a)).flatMap {
+            case Continue =>
+              loop(subscriber, newState)
+            case Stop =>
+              subscriber.onComplete()
```
Avasil: We shouldn't call `onComplete()` here because downstream subscriber already returned `Stop` to us and most likely called `onComplete` on its own

See: https://monix.io/docs/3x/reactive/observers.html#contract

Avasil:
Thanks, I think it looks good other than one comment I left. Build should pass if you merge to latest master

PR #990, opened 2019-08-25
3.0.0-RC4 Release Notes

Avasil:


CHANGES.md
```diff
@@ -1,3 +1,144 @@
...
+Last release before `3.0.0` which will be released as soon as Cats-Effect 2.0.0 is available. At the time of writing release notes,
+it is only waiting on Cats 2.0.0 which is about to release next RC which will become stable version if no critical issues are discovered.
+
+Any other development for this release is now frozen except if we discover critical bugs like memory leaks. All other
+changes will target `3.1.0`. 
+
+`3.0.0` will be binary and source compatible with `3.0.0-RC4` for Scala 2.12. Monix itself will be also binary compatible
+for 2.11 but it will have a dependency on Cats-Effect which is not. [See Cats-Effect release notes](https://github.com/typelevel/cats-effect/releases/tag/v2.0.0-RC1) for more details.
+
+We're very sorry for such a long wait. We wish we could release `3.0.0` already but if we released now, we would have to
```
alexandru: Don't apologize so much. It is what it is. It's not like we're paid for it.

alexandru: Saying that we need to synchronize with Cats and Cats-Effect 2.0 and that we are in feature freeze ... is IMO enough.

CHANGES.md
```diff
@@ -1,3 +1,144 @@
...
+Any other development for this release is now frozen except if we discover critical bugs like memory leaks. All other
+changes will target `3.1.0`. 
+
+`3.0.0` will be binary and source compatible with `3.0.0-RC4` for Scala 2.12. Monix itself will be also binary compatible
+for 2.11 but it will have a dependency on Cats-Effect which is not. [See Cats-Effect release notes](https://github.com/typelevel/cats-effect/releases/tag/v2.0.0-RC1) for more details.
+
+We're very sorry for such a long wait. We wish we could release `3.0.0` already but if we released now, we would have to 
+release `4.0.0` for Cats-Effect 2.0 because of 2.11 incompatibility there so we decided to hold on. 
+We hope you can understand and wait a little longer. It was a rough road but currently we have 3 active maintainers instead
+of just Alex so I hope it can give you a confidence for the future! :)
```
alexandru: That we have multiple active maintainers that can do releases going forward ... is worth mentioning 👍

CHANGES.md
```diff
@@ -1,3 +1,144 @@
...
+- `Local.isolate` has a overload for isolating `Future` which is safer than regular `Local.isolate`.
+
+The `Local` model since `3.0.0-RC3` shares by default. To isolate modifications of `Local` by
+other concurrently running `Future`, you have to call `Local.isolate`. 
+
+In case of `Task`, there is a `TaskLocal.isolate` version. It is automatically called whenever you run `Task`
+so if your use case is setting `correlationId` or similar, it probably won't require any explicit isolation because
+your HTTP library will most likely run the `Task` per request.
+
+NOTE: We are still looking for improvements for `Local` and it is likely that it will see additional changes in subsequent releases.
```
alexandru: Well I hope the current design is solid enough that we shouldn't break binary compatibility for it. Otherwise we're screwed 🙂

But it's solid enough so I have my hopes up.

CHANGES.md
alexandru: So to expand on this point — we are not in under any contractual obligation, we are not getting paid for it and feeling bad about delaying a final release in an open source is a recipe for burnout, since it stops being fun.

I think we've got a solid release and that it will happen along with Cats-Effect 2.0, we worked hard on it, so that's nothing to apologize for.

This is me giving unsolicited advice, given that I experienced burnout before 🙂

CHANGES.md
Avasil: I hope so too but if we discover possible improvements, such as doing it without support in `Task` run-loop or figure out a sensible way to auto-isolate in parallel operators without breaking `fa.start.flatMap(_.join) <-> fa` then I think it could be worth considering

Do you think we should remove/reword this note?

CHANGES.md
Avasil: Thanks, that's a good advice.

Updated it

CHANGES.md
alexandru: Yes. Delete this paragraph too.

Avasil: Ok, removed :)

PR #997, opened 2019-08-29
Bump up dependencies.

takayahilton:


Avasil:
Thanks @takayahilton !
I think it also needs a Scala version bump in travis yaml

Issue #1002, opened 2019-09-02
Stack safety of Observable.combineLatestList


scubamaggo:
In contrast to the implementations in RxScala and RxJava, the Monix (3.0.0-RC5) implementation of `Observable.combineLatestList` is not stack safe.

```
locally {
 implicit val scheduler = monix.execution.schedulers.TestScheduler()

  val n = 1000 // works
  //val n = 10000 // Stack overflow!
  val sources = (1 to n).map(i => monix.reactive.Observable.fromIterable(List(i))).toList
  val combined = monix.reactive.Observable.combineLatestList(sources: _*)

  combined.subscribe(
     values => { println("monix: " + values.length); monix.execution.Ack.Continue },
     println
  )
}

locally {
  val n = 100000 // works
  val sources = (1 to n).map(i => rx.lang.scala.Observable.just(i)).toIterable
  val combined = rx.lang.scala.Observable.combineLatest(sources)(identity)
  
  combined.subscribe(
     values => println("rxscala: " + values.length)
  )
}
```

According to the documentation of  `combineLatestList` this seems inevitable since it is implemented via `combineLatestMap2`.

alexandru:
Indeed it looks like a bug, we'll have to fix it.

PR #1008, opened 2019-09-05
Fix for #1002

jbracker:
This is a fix for #1002. 

All it does it replace the default implementation of `Observable.combineLatestList` with the custom implementation `CombineLatestListObservable`. The custom implementation takes special care to not cause any stack issues and tries to produce as little overhead as possible.

`CombineLatestListObservable` is based on the template set by `CombineLatest2Observable` and just expands on it to handle any number of incoming observables.

Should `Observable.combineLatestList` use the specialized implementations for 1 to 6 incoming `Observable`s if possible (i.e. `CombineLatest2Observable`, `CombineLatest3Observable`, etc.)?


Avasil:
Thank you very much for PR @jbracker 
We are a bit busy with `3.0.0` release so we are slacking a bit with reviews but I will try to review it as soon as I can and this is something we can add in `bugfix` release.

> Should Observable.combineLatestList use the specialized implementations for 1 to 6 incoming Observables if possible (i.e. CombineLatest2Observable, CombineLatest3Observable, etc.)?

It will be hard to know unless we benchmark it so I think it's OK in the current state

alexandru:
A review for this ticket is on my TODO 🙂

The problem with these `Observable` operators is that they operate with concurrent side effects.

jbracker:
Glad to hear that this is making progress :)

> The problem with these Observable operators is that they operate with concurrent side effects.

That's why I based the implementation on `CombineLatest2Observable`.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -5859,14 +5859,10 @@ object Observable extends ObservableDeprecatedBuilders {
     * returning a new observable that generates sequences.
```
jakoschiko: The doc mentions that the implemenation is based on combineLatestMap2. That's not true anymore.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/CombineLatestListObservable.scala
```diff
@@ -0,0 +1,164 @@
...
+    val observables: Array[Observable[(A, Int)]] = new Array(numberOfObservables)
+    observables.indices.foreach { i =>
+      observables(i) = obss(i).map(x => (x, i))
+    }
+    // MUST BE synchronized by `lock`
+    var lastAck = Continue: Future[Ack]
+    // MUST BE synchronized by `lock`
+    val elems: mutable.ArraySeq[A] = new mutable.ArraySeq(numberOfObservables)
+    // MUST BE synchronized by `lock`
+    val hasElems: Array[Boolean] = new Array(numberOfObservables)
```
Avasil: Could be `Array.fill(numberOfObservables)(false)`

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/CombineLatestListObservable.scala
```diff
@@ -0,0 +1,164 @@
...
+
+    val numberOfObservables = obss.size
+
+    val lock = new AnyRef
+    var isDone = false
+
+    // NOTE: We use arrays and other mutable structures here to be as performant as possible.
+
+    // MUST BE synchronized by `lock`
+    val observables: Array[Observable[(A, Int)]] = new Array(numberOfObservables)
```
Avasil: I wonder if we could store `(Int, Observable[A])` instead to not create and unpack a tuple on each `onNext`?

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/CombineLatestListObservable.scala
```diff
@@ -0,0 +1,164 @@
...
+          val index: Int = elemAndIndex._2
+          if (isDone) {
+            Stop
+          } else {
+            elems(index) = elem
+            if (!hasElems(index)) {
+              hasElems(index) = true
+            }
+
+            if (hasElems.forall(identity)) {
```
Avasil: We could probably set a flag the first time it is true so we don't have to check full array each time

jakoschiko: Or we store the number of `true`s in an additional val, so we never have to check the full array.

jbracker:
@Avasil @jakoschiko I have addressed some of the comments:

- Comments have been updated. I did not see any harm leaving the class comment for `CombineLatestListObservable`.
- `toArray` is not an option because we don't have a `ClassTag` for `Observable[A]` and I don't think we wan't one there. I could make it an ArraySeq or a Vector though if you think that would be preferable.
- Using `Array.fill` now for initialization of `hasElems`.
- Yes, you are right I don't need to map the index into the observable. Having it on the outside is sufficient.
- I created the counter `hasElemCount` to allow fast checks.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/CombineLatestListObservable.scala
```diff
@@ -0,0 +1,163 @@
...
+        lastAck = Stop
+      }
+    }
+
+    val composite = CompositeCancelable()
+
+    observables.foreach { case (obs, index) =>
+      composite += obs.unsafeSubscribeFn(new Subscriber[A] {
+        implicit val scheduler: Scheduler = out.scheduler
+
```
oleg-py: Correct me if I wrong, but why do we need these tuples in `observables`? It seems that `index` is only used here, so one should be able to drop these allocations just by doing:
```var i = 0
observables.foreach { obs =>
  val index = i // close over proper index
  i += 1
```
and thus you don't need to reallocate collection at all (can just use original `obss` coming to you)

jakoschiko: Or
```
observables.iterator.zipWithIndex.map { case (obs, index) =>
    ???
}}}
```

oleg-py: @jakoschiko that is indeed an option, but it's only halfway - you skip the allocation of an extra collection, but not the tuples for each instance.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/CombineLatestListObservable.scala
```diff
@@ -0,0 +1,164 @@
...
+import monix.execution.{Ack, Cancelable, Scheduler}
+import monix.reactive.Observable
+import monix.reactive.observers.Subscriber
+
+import scala.collection.mutable
+import scala.concurrent.Future
+import scala.util.Success
+import scala.util.control.NonFatal
+
+/** Only used in [[Observable.combineLatestList()]]. Is tested by `CombineLatestListSuite`. */
```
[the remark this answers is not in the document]
oleg-py: Yeah, we don't really mention tests in comments, but we do say "implementation for `Observable.combineLatestList`"

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/CombineLatestListObservable.scala
```diff
@@ -0,0 +1,162 @@
...
+private[reactive] final class CombineLatestListObservable[A, +R](observables: Seq[Observable[A]])(f: Seq[A] => R)
```
oleg-py: I think we can get away without the `f` here.

`combineLatestMapX` helps avoid tuple allocations and induced primitive boxing, but we probably aren't buying much over `combineLatestList(...).map` as you do in your tests.

If you go with dropping it, you won't need to guard against user errors in f, so:
```scala
def rawOnNext(as: Seq[A]): Future[Ack] = if (isDone) Stop else out.onNext(as)
```

Avasil:
Many thanks @jbracker :)

Issue #995, opened 2019-08-28
Proposal: add Task.flatMapLoop method


alexandru:
I think this method would be cool to have on `Task`:

```scala
sealed abstract class Task[+A] {

  // ...

  def flatMapLoop[S](seed: S)(f: (A, S, S => Task[S]) => Task[S]): Task[S] =
    this.flatMap { a => f(a, seed, loop(_)(f)) }
}
```

Usage sample:

```scala
val random = Task(Random.nextInt())

random.flatMapLoop(Vector.empty[Int]) { (a, list, continue) =>
  val newList = list :+ a

  if (newList.length < 10)
    continue(newList)
  else
    Task.now(newList)
}
```

This saves us from describing a loop via `def`, which I find super annoying:

```scala
def loop(list: Vector[Int]): Task[Vector[Int]] =
  random.flatMap { a =>
    val newList = list :+ a

    if (newList.length < 10)
      loop(newList)
    else
      Task.now(newList)    
  }

loop(Vector.empty)
```

So it's a syntactic helper that's similar with `onErrorRestartLoop`.

---

TODO:

- [ ] check if there's already an alternative in [Cats](https://github.com/typelevel/cats) (when in doubt, ask on Cats's gitter :))
- [ ] get the signature right, one problem is that this looks like a `foldLeft`, however usually the order of params in `foldLeft` in Scala is `(S, A)` and not `(A, S)` (like we're doing here)
- [ ] also add it for `Coeval`

alexandru:
Btw, this helper is super obvious — is it described anywhere? In Cats maybe?

alexandru:
👍 let's call it flatMapLoop, has a nicer ring to it. Also search for an equivalent in Cats first.

alexandru:
One issue we need to solve is to get the signature right — this function looks like a `foldLeft`:

```scala
def foldLeft[A, B](fa: F[A], b: B)(f: (B, A) => B): B
```

This is consistent with the definition from Haskell:

```haskell
foldl :: Foldable t => (b -> a -> b) -> b -> t a -> b
```

In other words maybe the signature should be:

```scala
def flatMapLoop[S](seed: S)(f: (S, A, S => Task[S]) => Task[S]): Task[S]
```

alexandru:
Note that `foldRight` on the other hand is:

```haskell
foldr :: Foldable t => (a -> b -> b) -> b -> t a -> b
```

```scala
[...]

PR #1010, opened 2019-09-06
First stab at flatMapLoop for Task and Coeval

mudsam:
#995

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -57,3 +57,6 @@ https://github.com/lukestephenson
 
 Paweł Kiersznowski
 https://github.com/pk044
+
+Johan Mudsam
+https://github.com/mudsam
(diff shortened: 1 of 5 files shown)


alexandru:
Cool. Needs documentation I think.

mudsam:
Apologies for the delay, how about this as documentation?
`
  /** Enables flatMap loops over Tasks
    * 
    * Sample:
    * 
    * {{{
    *   val random = Task(Random.nextInt())
    *   val loop = random.flatMapLoop(Vector.empty[Int]) { (a, list, continue) =>
    *     val newList = list :+ a
    *     if (newList.length < 5)
    *       continue(newList)
    *     else
    *       Task.now(newList)
    *   }
    * }}}
    * 
    * @param seed initializes the output
    * @param f is a function that takes three inputs:
    *        1. the next value
    *        2. the accumulated result of the loop
    *        3. the continuation function of the loop.
    * @return a new [[Task]] that contains the result of the loop.
    */
`

alexandru:
@mudsam that looks good, not sure if that list will be formatted properly. You can check by generating the documentation with `unidoc`.

Also instead of this:

> Enables flatMap loops over Tasks

You might say something like this:

> Describes `flatMap`-driven loops, as an alternative to recursive functions.

PR #1012, opened 2019-09-07
added ConcurrentQueue.isEmpty

### LINKED ISSUE #978: add `isEmpty` to the ConcurrentQueue

pk044:
Closes #978


monix-catnap/shared/src/main/scala/monix/catnap/ConcurrentQueue.scala
```diff
@@ -277,6 +277,15 @@ final class ConcurrentQueue[F[_], A] private (
     notifyProducers()
   }
 
+  /** Checks if the queue is empty.
+    *
+    * '''UNSAFE PROTOCOL:''' Concurrent shared state changes very frequently, therefore this function might yield nondeterministic results.
+    */
+  @UnsafeProtocol
+  @UnsafeBecauseImpure
```
Avasil: It is actually pure because of `F[_]`

monix-catnap/shared/src/main/scala/monix/catnap/ConcurrentQueue.scala
```diff
@@ -277,6 +277,15 @@ final class ConcurrentQueue[F[_], A] private (
     notifyProducers()
   }
 
+  /** Checks if the queue is empty.
+    *
+    * '''UNSAFE PROTOCOL:''' Concurrent shared state changes very frequently, therefore this function might yield nondeterministic results.
```
Avasil: I think it could use a new line. Also maybe similar warning to `tryPoll` would be more accurate, i.e. it implies an understanding of concurrency for given use case. I don't have any good idea about wording off the top of my head :D

Any ideas @alexandru ?

monix-catnap/shared/src/main/scala/monix/catnap/ConcurrentQueue.scala
pk044: oh yeah, crap, you're totally right - i forgot about that fact when i was reading the description of that annotation haha

monix-catnap/shared/src/main/scala/monix/catnap/ConcurrentQueue.scala
pk044: ok, i wrote it in the manner you suggested, do you think it's correct more or less/should I add something more?

alexandru:
@pk044 this looks good, but can you add some tests? Nothing fancy.

Also, wouldn't this make sense for `ConcurrentChannel` too? Probably on the consumer side. If yes, that could be a different PR.

pk044:
could you guys see if the test I've just added is enough? 

I'll make another PR for ConcurrentChannel soon too.

Avasil:
Thanks @pk044

PR #1016, opened 2019-09-08
Add generic observable take until

### LINKED ISSUE #1005: Observable.takeUntil for F[_]
We have

```scala
final def takeUntil(trigger: Observable[Any]): Observable[A]
```
which takes elements from source `Observable` until `trigger` emits. I would like to see a variant which takes `Task` and/or `F[_]` so it could interop with pure code bases in more natural way.

For instance, we can cancel `Observable` from outside using `takeWhileNotCanceled` but if we want to be pure and use `Deferred` we have to:

```scala
source.takeUntil(Observable.fromTask(deferred.get))
```

Implementation could look like this but I'd look first if it is possible to avoid conversion in easy way. [Here's implementation of TakeUntilObservable](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/TakeUntilObservable.scala) and [Observable.fromTask](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/TaskAsObservable.scala) for reference.

Not sure if the name should be overload to `takeUntil` or something new like `takeUntilEval`

allantl:
Closes #1005

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -3125,6 +3125,27 @@ abstract class Observable[+A] extends Serializable { self =>
   final def takeUntil(trigger: Observable[Any]): Observable[A] =
     new TakeUntilObservable[A](self, trigger)
 
+  /** Version of [[takeUntil]] that can work with a trigger expressed by a [[monix.eval.Task]]
+    *
+    * @see [[takeUntil]] for version that works with Observable.
+    * @see [[takeUntilEvalF]] for version that works with generic `F[_]` powered by [[monix.eval.TaskLike]].
+    *
+    * @param trigger task that will cancel the stream as soon as it completes.
+    */
+  final def takeUntilEval(trigger: Task[_]): Observable[A] =
+    self.takeUntil(Observable.fromTask(trigger))
+
+  /** Version of [[takeUntil]] that can work with a trigger expressed by a generic `F[_]`
+    * provided an implicit [[monix.eval.TaskLike]] exists.
+    *
+    * @see [[takeUntil]] for version that works with Observable.
+    * @see [[takeUntilEval]] for version that works with [[monix.eval.Task]].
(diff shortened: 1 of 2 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -3125,6 +3125,27 @@ abstract class Observable[+A] extends Serializable { self =>
...
     new TakeUntilObservable[A](self, trigger)
 
+  /** Version of [[takeUntil]] that can work with a trigger expressed by a [[monix.eval.Task]]
+    *
+    * @see [[takeUntil]] for version that works with Observable.
+    * @see [[takeUntilEvalF]] for version that works with generic `F[_]` powered by [[monix.eval.TaskLike]].
+    *
+    * @param trigger task that will cancel the stream as soon as it completes.
+    */
+  final def takeUntilEval(trigger: Task[Any]): Observable[A] =
```
alexandru: Instead of `Task[Any]` we could use `Task[_]`. Same for the `Observable` and the `F[_]` versions.

The `F[_]` version isn't even correct because it isn't specified as a covariant type and so it doesn't really work.

So change `Any` to `_`.

allantl: Ah yes, you're right. I've updated the PR.

alexandru:
While you're at it, also change the `Observable` version. I don't see why not. Shouldn't break source or binary compatibility.

allantl:
Unfortunately, it breaks binary compatibility.
Screenshot 2019-09-16 at 21 27 31

alexandru:
That's weird but if it's so, revert the Observable version to usage of Any, since it's not an problen anyway and we might deal with it later.

allantl:
Yeap, its done.

allantl:
@Avasil My pleasure. I'm happy to help. Thanks for all the good work!

PR #1025, opened 2019-09-11
Add openjdk11 builds in CI

Avasil:
Let's see how CI handle it
Maybe we could also try testing different JDK only on one Scala version


alexandru:
Nice. Seems to work.

.travis.yml
```diff
@@ -17,6 +21,10 @@ matrix:
     - jdk: oraclejdk8
       scala: 2.12.9
       env: COMMAND=ci-jvm-all
+    # Scala 2.12, JVM, JDK 11
+    - jdk: openjdk11
+      scala: 2.12.9
```
mihaisoloi: 2.12.10 is out now

alexandru: That should probably come in a different PR — tried to upgrade last week (at work) and macro-paradise isn't published for it yet.

PR #1044, opened 2019-10-04
WIP: add option to isolate on fork

oleg-py:
Posting this quick experiment to get some feedback:

- new option `contextIsolationOnFork` in `Task.Options` (in an ugly, but bincompatible way)
- its handling in `task.start` that makes it isolate, but  `join` overrides whatever you had before - if you ever `join`

This still preserves `fa.start.flatMap(_.join) <-> fa` when `fa` is changing locals, just in a different way. It might prove less surprising for users of TaskLocal, but I don't know the extent of the impact for cats-effect based libraries like fs2 which might use `Fiber`s underneath. At least with `Options` you can switch things around as you please, and we can see if we should consider changing the default model.

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -31,10 +31,10 @@ import monix.execution.misc.Local
 import monix.execution.schedulers.{CanBlock, TracingScheduler, TrampolinedRunnable}
 import monix.execution.compat.BuildFrom
 import monix.execution.compat.internal.newBuilder
-
 import scala.annotation.unchecked.{uncheckedVariance => uV}
 import scala.concurrent.duration.{Duration, FiniteDuration, NANOSECONDS, TimeUnit}
 import scala.concurrent.{ExecutionContext, Future, TimeoutException}
+import scala.runtime.AbstractFunction2
 import scala.util.{Failure, Success, Try}
 
 /** `Task` represents a specification for a possibly lazy or
@@ -4191,8 +4191,11 @@ object Task extends TaskInstancesLevel1 {
     */
   final case class Options(
     autoCancelableRunLoops: Boolean,
-    localContextPropagation: Boolean
+    localContextPropagation: Boolean,
+    contextIsolationOnFork: Boolean
   ) {
+    def this(autoCancelableRunLoops: Boolean, localContextPropagation: Boolean) =
+      this(autoCancelableRunLoops, localContextPropagation, false)
     /** Creates a new set of options from the source, but with
       * the [[autoCancelableRunLoops]] value set to `true`.
       */
@@ -4217,6 +4220,9 @@ object Task extends TaskInstancesLevel1 {
     def disableLocalContextPropagation: Options =
       copy(localContextPropagation = false)
 
+    def enableContextIsolationOnFork: Options =
+      new Options(autoCancelableRunLoops, localContextPropagation, true)
+
     /**
       * Enhances the options set with the features of the underlying
       * [[monix.execution.Scheduler Scheduler]].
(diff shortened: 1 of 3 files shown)


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -4191,8 +4191,11 @@ object Task extends TaskInstancesLevel1 {
     */
   final case class Options(
     autoCancelableRunLoops: Boolean,
-    localContextPropagation: Boolean
+    localContextPropagation: Boolean,
+    contextIsolationOnFork: Boolean
```
Avasil: If we go through with it, we could also add env variable to https://github.com/monix/monix/blob/master/monix-execution/jvm/src/main/scala/monix/execution/internal/Platform.scala

monix-eval/shared/src/main/scala/monix/eval/internal/TaskStart.scala
```diff
@@ -50,10 +53,32 @@ private[eval] object TaskStart {
...
+        }
+      Task.unsafeStartEnsureAsync(fa2, ctx2, Callback.fromPromise(p))
       // Signal the created fiber
       val task = Task.fromCancelablePromise(p)
-      cb.onSuccess(Fiber(task, ctx2.connection.cancel))
+      val join =
+        if (pCtx eq null) task
+        else
+          task.flatMap { a =>
+            if (pCtx ne null)
```
Avasil: Can it be not null before `flatMap` and then change?

monix-eval/shared/src/main/scala/monix/eval/Task.scala
oleg-py: Yep, if we go with this approach, we should do it for consistency

monix-eval/shared/src/main/scala/monix/eval/internal/TaskStart.scala
oleg-py: Well spotted - I forgot to remove that as I was moving code around 😅

alexandru:
@oleg-py can you please explain why this is needed?

And why does it have to be an option in `Task.Options`?

oleg-py:
@alexandru there are at least two lawful models for `TaskLocal` behavior in fibers:
1. we share everything by default unless you explicitly use `isolate`
2. we isolate every fiber (it gets its own copy of locals), but switch to its context if/when you `join`

Currently we only support model 1. For some uses - specifically, those not under your control and/or using tagless final, it's desirable to have both. As a contrived example, I can have a library that simplifies producer-consumer pattern:

```scala
object ProCon {
  def forever[F[_]: Concurrent: Parallel, A](produce: F[A], consume: A => F[Unit], bufferSize: Int): F[Unit] =
    for {
      q <- Queue.bounded[F, A](bufferSize)
      pro = (produce >>= q.enqueue1).foreverM[Unit]
      con = (q.dequeue1 >>= consume).foreverM[Unit]
      _ <- pro &> con // [1]
    } yield ()
}
```

As a user of monix and such library, I might need to put `isolate` around `pro` and `con` on line [1] to have producer and consumer fibers have their own copy, but I cannot do this without modifying the library. Putting isolate on `produce` and `consume` has totally different semantics (per each value produced as opposed to one for all values produced). Having it as a task option would allow me to just do:
```
ProCon.forever(mkInt, printInt, 42).executeWithOptions(_.enableContextIsolationOnFork).attempt
```

or if it's what my app needs most of the time, I can have this option toggled on by default and switch to shared model when I need to.

Putting it on options since it'll affect the execution of sub-tasks like the other two options we have. In 4.0 we can probably drop the booleans and have a enum of how locals work with  { `Disabled`, `Shared`, `Isolated` } as options. I'm open to other ideas, but I can't think of one.

jvican:
@oleg-py Are you planning to continue work on this feature? I think I've just run into the underlying issue that this PR fixes, as well as #1146. My expectation was that `start` would appropriately isolate the context automatically without any user intervention.

oleg-py:
@jvican well, given it's been over a year, probably no. I'm watching related developments in cats-effect 3 [here](https://github.com/typelevel/cats-effect/pull/1393) though. I'm not sure what's the absolute best model would possibly be, but I'm leaning towards one where creator of local specifies the sharing strategy instead of user.

Avasil:
I've created an issue for it a while ago: https://github.com/monix/monix/issues/1302

Personally, I would like to support auto-isolation on fork, as a default.

> but I'm leaning towards one where creator of local specifies the sharing strategy instead of user.

I feel like that at best, we could have it as a `Task` option, not `Local` because I don't see a way to have this strategy with `Future`

oleg-py:
> I feel like that at best, we could have it as a `Task` option, not `Local` because I don't see a way to have this strategy with `Future`

I am talking about `TaskLocal` only since we have no control over how operations with `Local` are forked. I think that creator of `TaskLocal` knows best what it should be used _for_ and can therefore pick a proper strategy. Having an option on Task means that you, as a user, must know what bits of code uses TaskLocal so you can properly isolate (or not) on the outside.

PR #1054, opened 2019-10-20
Issue #1028 - Added Task.fromPublisher 

### LINKED ISSUE #1028: Create Task from Reactive Streams
Some libraries use Reactive Streams `Publisher` results for streaming a single event.

In my work project I ended up with this helper which might be useful in Monix:

```scala
  def fromPublisher[A](source: Publisher[A]): Task[Option[A]] =
    Task.cancelable0 { (sc, cb) =>
      val sub = SingleAssignSubscription()

      source.subscribe(new Subscriber[A] {
        private[this] var active = true

        def onSubscribe(s: Subscription): Unit = {
          sub := s
          sub.request(1)
        }

        def onNext(a: A): Unit = {
          if (active) {
            active = false
            sub.cancel()
            cb.onSuccess(Some(a))
          }
        }

        def onError(e: Throwable): Unit = {
          if (active) {
            active = false
            cb.onError(e)
          } else {
            sc.reportFailure(e)
          }
        }

        def onComplete(): Unit = {
          if (active) {
            active = false
            cb.onSuccess(None)
          }
        }
      })

      Task(sub.cancel())
    }
```

pk044:
Closes #1028


monix-eval/shared/src/test/scala/monix/eval/TaskConversionsSuite.scala
```diff
@@ -378,6 +379,45 @@ object TaskConversionsSuite extends BaseTestSuite {
     assertEquals(f2.value, Some(Success(99)))
   }
 
+  test("Task.fromPublisher protects against user error") { implicit s =>
```
oleg-py: Since we have `task.toReactivePublisher`, it's probably worth adding a property-based test about how they relate, e.g.:
```scala
check1 { task: Task[Thing] =>
  Task.fromPublisher(task.toReactivePublisher) <-> task.map(Some(_))
}
```

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -2675,6 +2676,21 @@ object Task extends TaskInstancesLevel1 {
...
+    * See the [[http://www.reactive-streams.org/ Reactive Streams]]
+    * protocol that Monix implements.
+    *
+    * @see [[Task.toReactivePublisher]] for converting a `Task` to
+    *      a reactive publisher.
+    *
+    * @param source is the `org.reactivestreams.Publisher` reference to
+    *        wrap into a [[Task]].
+    */
+  def fromPublisher[A](source: Publisher[A]): Task[Option[A]] =
```
oleg-py: We might have this named `fromReactivePublisher` for consistency with `toReactivePublisher` that we already have.
/cc @alexandru

monix-eval/shared/src/main/scala/monix/eval/internal/TaskConversions.scala
```diff
@@ -107,6 +110,49 @@ private[eval] object TaskConversions {
...
+    * Implementation for `Task.fromReactivePublisher`.
+    */
+  def fromReactivePublisher[A](source: Publisher[A]): Task[Option[A]] =
+    Task.cancelable0 { (scheduler, cb) =>
+      val sub = SingleAssignSubscription()
+
+      source.subscribe(new Subscriber[A] {
+        private[this] var isActive = true
+
+        def onSubscribe(s: org.reactivestreams.Subscription): Unit = {
```
TapanVaishnav: nitpick: can you write `org.reactivestream` in imports instead of writing it here.

monix-eval/shared/src/main/scala/monix/eval/internal/TaskConversions.scala
```diff
@@ -23,6 +23,9 @@ import monix.execution.Callback
 import monix.eval.Task
 import monix.execution.Scheduler
 import monix.execution.schedulers.TrampolinedRunnable
+import org.reactivestreams.{Publisher, Subscriber}
```
TapanVaishnav: formatting imports

monix-eval/shared/src/test/scala/monix/eval/TaskConversionsSuite.scala
```diff
@@ -378,6 +379,51 @@ object TaskConversionsSuite extends BaseTestSuite {
...
+
+    val pub = new Publisher[Int] {
+      def subscribe(s: Subscriber[_ >: Int]): Unit = {
+        s.onSubscribe(new Subscription {
+          var isActive = true
+          def request(n: Long): Unit = {
+            if (n > 0 && isActive) {
+              isActive = false
+              s.onNext(1)
+              s.onComplete()
```
TapanVaishnav: I think this call is irrelevant here, what do you think?

Avasil: I think it's fine, it doesn't change the test but it is something that would happen in real scenario so we might as well keep it

TapanVaishnav: Sounds reasonable 👍

PR #1057, opened 2019-10-22
Fix default scheduleOnce implementation

alexandru:
Unfortunately we've got a nasty bug:

```scala
scheduler.scheduleOnce(0, MILLISECONDS, () => {
  println(Thread.currentThread().getName)
})
```

This will print `monix-scheduler`, because the logic will execute on Monix's thread-pool meant for scheduling stuff. And unfortunately this extends to task as well...

```scala

Task.sleep(-1.millis) *> Task {
  println(Thread.currentThread().getName)
}
```

This too will print `monix-scheduler`.

### DIFF
--- a/monix-execution/jvm/src/main/scala/monix/execution/internal/ScheduledExecutors.scala
+++ b/monix-execution/jvm/src/main/scala/monix/execution/internal/ScheduledExecutors.scala
@@ -28,14 +28,17 @@ private[execution] object ScheduledExecutors {
     * an underlying Java `ScheduledExecutorService`.
     */
   def scheduleOnce(
-    self: Scheduler,
-    executor: ScheduledExecutorService)(initialDelay: Long, unit: TimeUnit, r: Runnable): Cancelable = {
+    executor: Scheduler,
+    scheduler: ScheduledExecutorService
+  )(
+    initialDelay: Long, unit: TimeUnit, r: Runnable
+  ): Cancelable = {
     if (initialDelay <= 0) {
       executor.execute(r)
       Cancelable.empty
     } else {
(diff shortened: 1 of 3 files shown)


monix-execution/jvm/src/main/scala/monix/execution/internal/ScheduledExecutors.scala
```diff
@@ -28,14 +28,17 @@ private[execution] object ScheduledExecutors {
...
   def scheduleOnce(
-    self: Scheduler,
-    executor: ScheduledExecutorService)(initialDelay: Long, unit: TimeUnit, r: Runnable): Cancelable = {
+    executor: Scheduler,
+    scheduler: ScheduledExecutorService
+  )(
+    initialDelay: Long, unit: TimeUnit, r: Runnable
+  ): Cancelable = {
     if (initialDelay <= 0) {
       executor.execute(r)
```
oleg-py: I didn't realize initially that bug is on this line, because the actual diff looks like just rename

oleg-py: Probably worth noting in commit message/release notes/etc. that it was triggered by using nonpositive delays

alexandru:
Released hash version: `3.0.0-bfc2b51`

monix-execution/jvm/src/main/scala/monix/execution/internal/ScheduledExecutors.scala
alexandru: Indeed. I think it was the naming that made this confusing, hence the rename.

PR #1059, opened 2019-10-22
isEmpty for ConcurrentChannel ConsumerF

pk044:
as was requested in this issue: https://github.com/monix/monix/pull/1012#issuecomment-531655051


monix-catnap/shared/src/main/scala/monix/catnap/ConcurrentChannel.scala
```diff
@@ -924,6 +924,10 @@ object ConcurrentChannel {
           }
       }
     }
+
+    override def isEmpty: F[Boolean] =
+      F.pure(queue.isEmpty)
```
oleg-py: Should be `F.delay`. Testing the state of a mutable queue is not referentially transparent.

monix-catnap/shared/src/test/scala/monix/catnap/ConcurrentChannelSuite.scala
```diff
@@ -113,6 +113,41 @@ abstract class BaseConcurrentChannelSuite[S <: Scheduler] extends TestSuite[S] {
...
   }
 
+  testIO("consumer queue is empty after pulling all messages", times = repeatForFastTests) { implicit ec =>
+    for {
+      chan <- ConcurrentChannel[IO].withConfig[Int, Int](boundedConfig)
+      consume = chan.consume.use { consumer =>
+        for {
+          _ <- consumer.pull
+        } yield {
+          assertEquals(consumer.isEmpty.unsafeRunSync(), true)
```
oleg-py: That's actually an odd test. I'd expect `isEmpty` to be false, since you are also able to `pull` a halt event (`Left`).
You might also want to synchronize your producer and consumer with a `Deferred` to know exactly when pushing/halting has completed.

oleg-py (changes requested):

monix-catnap/shared/src/test/scala/monix/catnap/ConcurrentChannelSuite.scala
pk044: i had a lot of doubts when writing this test, i'll try implementing it another way

pk044:
@alexandru I have to admit I'm not quite familiar with ConcurrentChannel so I'm not sure how to test it properly, do you perhaps have any thoughts on this PR?

Issue #1033, opened 2019-09-17
StackOverflowError when using parSequence


oleg-py:
Running this app:

```scala
import monix.eval.{Task, TaskApp}
import cats.effect.ExitCode
import cats.implicits._

object TestRunner extends TaskApp {
  override def run(args: List[String]): Task[ExitCode] = {
    List.fill(2000)(Task.unit).parSequence_
      .executeWithOptions(_.enableLocalContextPropagation)
      .as(ExitCode.Success)
  }
}
```

produces (shortened):

```
java.lang.StackOverflowError
        at java.security.AccessController.doPrivileged(Native Method)
        at java.lang.ClassLoader.checkPackageAccess(ClassLoader.java:500)
        at monix.execution.internal.Trampoline.monix$execution$internal$Trampoline$$immediateLoop(Trampoline.scala:70)

        // :loop
        at monix.execution.internal.Trampoline.startLoop(Trampoline.scala:32)
        at monix.execution.schedulers.TrampolineExecutionContext$JVMOptimalTrampoline.startLoop(TrampolineExecutionContext.scala:143)
        at monix.execution.internal.Trampoline.execute(Trampoline.scala:40)
        at monix.execution.schedulers.TrampolineExecutionContext.execute(TrampolineExecutionContext.scala:64)
        at monix.execution.schedulers.BatchingScheduler.execute(BatchingScheduler.scala:50)
        at monix.execution.schedulers.BatchingScheduler.execute$(BatchingScheduler.scala:47)
        at monix.execution.schedulers.TracingScheduler$Base.execute(TracingScheduler.scala:53)
        at monix.eval.internal.TaskRestartCallback.start(TaskRestartCallback.scala:56)
        at monix.eval.internal.TaskRunLoop$.executeAsyncTask(TaskRunLoop.scala:592)
        at monix.eval.internal.TaskRunLoop$.startFull(TaskRunLoop.scala:120)
        at monix.eval.Task$.unsafeStartNow(Task.scala:4582)
        at monix.eval.Task$.unsafeStartEnsureAsync(Task.scala:4563)
        at monix.eval.internal.TaskMapBoth$Register.apply(TaskMapBoth.scala:128)
        at monix.eval.internal.TaskMapBoth$Register.apply(TaskMapBoth.scala:43)
        at monix.eval.internal.TaskRestartCallback.run(TaskRestartCallback.scala:65)
[...]

Issue #1023, opened 2019-09-10
Possible leak with Local


alexandru:
Reported on Gitter, see [September 10, 2019 10:45 AM](https://gitter.im/monix/monix?at=5d7754a11882325229974e05).

Reproducible example: https://gist.github.com/hanny24/3096f1541e1f05716f8c477691c36f2f

To be investigated.

Avasil:
@hanny24 
Which version of http4s do you use? I'm running it for 30 mins on jdk1.8 with http4s 0.20.10 and so far I haven't noticed any indication of memory leak. I will leave it running for the remainder of the night. How long does it take to manifest?

hanny24:
I'm using monix RC5, http4s 0.20.10, cats effect 1.4.0 and OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.3+7). I'm running it JVM options `-Xmx64m -XX:+HeapDumpOnOutOfMemoryError`.  

I'm using `ab` to generate a lot of traffic: `ab -k -n 50000 -c 10 http://127.0.0.1:8080/hello` and it usually takes just a few seconds to run out of heap space. 

I can even provide you with a heap dump. Available on my [gdrive](https://drive.google.com/file/d/1z-GvVNlIIjdwjOHoYZb2VnWexp1UWIFt/view?usp=sharing).

Quick check in Memory Analyzer shows this:
```
The thread scala.concurrent.impl.ExecutionContextImpl$DefaultThreadFactory$$anon$1 @ 0x7fc4fd278 scala-execution-context-global-16 keeps local variables with total size 25 732 920 (39,43 %) bytes.

The memory is accumulated in one instance of "java.lang.ThreadLocal$ThreadLocalMap$Entry[]" loaded by "<system class loader>".
The stacktrace of this Thread is available. See stacktrace.

Keywords
java.lang.ThreadLocal$ThreadLocalMap$Entry[]
```

Let me know if I can provide any further assistance.

hanny24:
Example SBT project: https://github.com/hanny24/monix-bug

guymers:
Could this be because the thread pool is unbounded and your computer cannot process the volume of requests?
If you used a fixed size pool with a fixed queue size is memory usage constant?
Alternatively if you bench with a much smaller number of requests such that that your computer can process them without needing to queue is memory usage constant?

hanny24:
Both `Scheduler.traced` and `Scheduler.global` are using the same thread pool.

Avasil:
[...]

PR #1062, opened 2019-10-25
Fix #870, #1023, #1033: fix scheduler decorators leaking ThreadLocalEntry's

### LINKED ISSUE #870: Memory leak with TrampolineExecutionContext$JVMOptimalTrampoline
We are currently experience a memory leak when using Monix Task 3.0.0-RC2. The application is mainly using `Task`, along with functions like `delayExecution`/`fromFuture`/`executeAsync` (along with using the default `Scheduler.global`)

Originally we thought it was coming from Doobie, but we narrowed this down to Monix Task. The leak appears to be happening at `monix.execution.schedulers.TrampolineExecutionContext$JVMOptimalTrampoline$$anon$3`. Here is a screenshot of VisualVM showing what the leak looks like

2019-04-25-185654_1920x1015_scrot

We have around 160k of these instances after running a few hours in production. Not sure if this is related to https://github.com/monix/monix/issues/791 since we are not using `Observable`.

Unfortunately this is company code however I will try and minimize the example

oleg-py:
Changed Trampoline to accept ECs as a parameter.
Changed TrampolineExecutionContext to use single ThreadLocal[Trampoline] on companion instead of one per instance.

See comment here for more details:
https://github.com/monix/monix/issues/870#issuecomment-545342047

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -319,7 +319,8 @@ lazy val cmdlineProfile =
 
 def mimaSettings(projectName: String) = Seq(
   mimaPreviousArtifacts := Set("io.monix" %% projectName % monixSeries),
-  mimaBinaryIssueFilters ++= MimaFilters.changesFor_3_0_0__RC5
+  mimaBinaryIssueFilters ++= MimaFilters.changesFor_3_0_0__RC5,
+  mimaBinaryIssueFilters ++= MimaFilters.changesFor_3_0_1
 )
 // https://github.com/lightbend/mima/pull/289
 mimaFailOnNoPrevious in ThisBuild := false
--- a/monix-eval/jvm/src/test/scala/monix/eval/TaskExecuteWithLocalContextSuite.scala
+++ b/monix-eval/jvm/src/test/scala/monix/eval/TaskExecuteWithLocalContextSuite.scala
@@ -1,0 +1,17 @@
+package monix.eval
+
+import scala.util.Success
+import cats.implicits._
+
+
+object TaskExecuteWithLocalContextSuite extends BaseTestSuite {
+  test("cats' parSequence with LCP is stack safe") { implicit sc =>
+    val f = List.fill(2000)(Task.unit).parSequence_
+      .executeWithOptions(_.enableLocalContextPropagation)
+      .runToFuture
+
+    sc.tick()
+
+    assertEquals(f.value, Some(Success(())))
+  }
+}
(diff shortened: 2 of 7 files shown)


oleg-py:
I haven't looked as to why #1033 is fixed. Possible that the code was generating a deep chain of decorators, each with its own trampoline, and so trampolining didn't actually work as intended. We'd still have some inefficiency if that's the case.

I also don't know of a sane way to test for the leak since it requires app to be running with low memory or for indeterminate time to crash. I just used publishLocal and ran all repro snippets I had.

mdedetrich:
@alexandru @Avasil Is there anything stopping this PR? Would be great to get new monix published with this fix since its a pretty serious memory leak.

Avasil:
@mdedetrich 

I'm trying to fully understand the issue and the fix and I'm hoping for @alexandru input
Although to me the PR looks good

I released a snapshot if you want to try it out: `3.1.0-6bf632d-SNAPSHOT`
I had some issues with Scala 2.11 but it should be available for 2.12 and 2.13

Avasil:
> I also don't know of a sane way to test for the leak since it requires app to be running with low memory or for indeterminate time to crash. I just used publishLocal and ran all repro snippets I had.

Can we test for stack overflow? It might not be the core issue but it is something we don't want to return. :D

oleg-py:
Interestingly, the test doesn't work on JS - it doesn't throw an error, but doesn't give output via ticking `TestScheduler` either.

mdedetrich:
Maybe JS is doing some optimization? Also check `"undefined behaviours" in https://www.scala-js.org/doc/semantics.html, i.e.

> The JVM is a very well specified environment, which even specifies how some bugs are reported as exceptions. Currently known exhaustive list of exceptions are:
> StackOverflowError and other VirtualMachineErrors

Unless you specify a specific flag to Scala.js, this behavior is classified as undefined (which can mean it may not even throw it)

mdedetrich:
@Avasil 

> I'm trying to fully understand the issue and the fix and I'm hoping for @alexandru input
Although to me the PR looks good

So I tried it out and it looks good on my end. @oleg-py What are your thoughts on merging this now? To me this looks safe and it would be ideal to make a release that doesn't have such a leak. I am not sure how busy @alexandru is right now to give a proper look into this

Avasil:
Thanks @mdedetrich - we agreed with Oleg to make a release this week, I will probably do it on Thursday or Wednesday

mdedetrich:
Many thanks!

PR #1063, opened 2019-10-25
Fix #1056: ensure boundary in TaskCreate if LCP enabled

### LINKED ISSUE #1056: ConcurrentChannel can break TaskLocal isolation
If `ConcurrentChannel.push` results in an async boundary, the producer fiber might lose it's own local context and switch to one the consumer has

Right now I don't know if this is caused by implementation of ConcurrentChannel or something like `Task.asyncF` not having proper isolation.

This example reproduces the issue on my machine. The key to triggering the condition seems to be filling the internal buffer of ConcurrentChannel (1024 by default, hence having 2100 in `produceLoop`) with producer faster than the consumer processes them (for which 1ms delay was enough for me)

```scala
import cats.effect.ExitCode
import monix.eval.{Task, TaskApp, TaskLocal}

import cats.effect.concurrent.Deferred
import cats.implicits._
import monix.catnap.ConcurrentChannel


object MonixFiberLoad extends TaskApp {

  override protected def options: Task.Options = super.options.enableLocalContextPropagation
  override protected def scheduler: Scheduler = Scheduler.traced

  class Test(
    l: TaskLocal[String],
    ch: ConcurrentChannel[Task, Unit, Int],
    sync: Deferred[Task, Unit]
  ) {
    private[this] def produceLoop(n: Int): Task[Unit] =
      if (n == 0) Task.unit else ch.push(n) >> l.read.flatMap { s =>
        Task(println(s"Local leak: expected 'producer', got '$s'")).whenA(s != "producer")
      } >> produceLoop(n - 1)

    def produce: Task[Unit] =
      for {
        _ <- sync.get
        _ <- l.write("producer")
        _ <- produceLoop(2100)
        _ <- ch.halt(())
        s <- l.read
        _ <- Task(println(s"Producer done, local: $s"))
      } yield ()

    def consume: Task[Unit] = ch.consume.use { c =>
      sync.complete(()).attempt >>
      c.pull.delayResult(1.milli).flatMap { x =>
        l.write(x.toString).as(x)
      }.iterateWhile(_.isRight).void >> l.read.flatMap { s =>
        Task(println(s"Consumer done, local: $s"))
      }
    }
  }

  override def run(args: List[String]): Task[ExitCode] =
    (
      TaskLocal("undefined"),
      ConcurrentChannel[Task].of[Unit, Int],
      Deferred[Task, Unit]
    )
      .mapN(new Test(_, _, _))
      .flatMap(t => TaskLocal.isolate(t.produce) &> TaskLocal.isolate(t.consume))
      .as(ExitCode.Success)
}
```
@alexandru do you have any quick ideas about this?

oleg-py:
The optimization bypassing the async boundary can break Local isolation.

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/internal/TaskCreate.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/internal/TaskCreate.scala
@@ -226,10 +226,11 @@ private[eval] object TaskCreate {
       if (shouldPop) ctx.connection.pop()
       // Optimization — if the callback was called on the same thread
       // where it was created, then we are not going to fork
+      // This is not safe to do when localContextPropagation enabled
       isSameThread = Platform.currentThreadId() == threadId
       try {
         ctx.scheduler.execute(
-          if (isSameThread)
+          if (isSameThread && !ctx.options.localContextPropagation)
             this
           else
             StartAsyncBatchRunnable(this, ctx.scheduler)
--- a/monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
+++ b/monix-eval/shared/src/test/scala/monix/eval/TaskLocalSuite.scala
@@ -19,11 +19,14 @@ package monix.eval
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
+
 import minitest.SimpleTestSuite
-import monix.execution.Scheduler
+import monix.execution.{BufferCapacity, CancelablePromise, Scheduler}
 import monix.execution.exceptions.DummyException
 import monix.execution.misc.Local
 import cats.implicits._
+import monix.catnap.{ConcurrentChannel, ConsumerF}
+import monix.execution.ExecutionModel.SynchronousExecution
 
 object TaskLocalSuite extends SimpleTestSuite {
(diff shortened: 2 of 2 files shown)


monix-eval/shared/src/main/scala/monix/eval/internal/TaskCreate.scala
```diff
@@ -226,10 +226,11 @@ private[eval] object TaskCreate {
       if (shouldPop) ctx.connection.pop()
       // Optimization — if the callback was called on the same thread
       // where it was created, then we are not going to fork
+      // This is not safe to do when localContextPropagation enabled
```
Avasil: Why is it not safe?

oleg-py: Because `this` extends a `TrampolinedRunnable`, and `TracingScheduler` extends `BatchingScheduler`, doing `execute` bypasses the `TracingRunnable` creation and context isolation that it has.

Without that isolation, the continuation of that async/asyncF/etc. node therefore will have the context from whoever called the `cb`, not from where the `async` was actually created.

That only happened in case `cb` was called synchronously, but it does happen deep in ConcurrentChannel impl in certain concurrent usage patterns with fast producer and slow-ish consumer.

PR #1064, opened 2019-10-25
Fix #1045: ensure early cancellation of Observable#mapParallel on error

### LINKED ISSUE #1045: Observable#mapParallel could cancel earlier
If any concurrent `Task` fails we don't cancel immediately:

https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelUnorderedObservable.scala#L101

We send the error downstream so it will reply with `Stop` on the next completed `Task`:
https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelUnorderedObservable.scala#L108

And then it cancels. I think we could do it earlier, during the first error. The same applies to `mapParallelOrdered` .

cc @alexandru in case it is desired for some reason

TapanVaishnav:
closes: https://github.com/monix/monix/issues/1045

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
@@ -157,8 +157,8 @@ private[reactive] final class MapParallelOrderedObservable[A, B](
 
           case Failure(error) =>
             lastAck = Stop
-            composite.cancel()
             composite -= future.cancelable
+            composite.cancel()
             self.onError(error)
         }
       } catch {
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
@@ -103,7 +103,6 @@ private[reactive] final class MapParallelOrderedObservable[A, B](
                     composite -= head.cancelable
                   case Failure(ex) =>
                     lastAck = Stop
-                    composite.cancel() // cancel the whole downstream on error
                     self.onError(ex)
                 }
 
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
@@ -160,6 +160,7 @@ private[reactive] final class MapParallelOrderedObservable[A, B](
 
           case Failure(error) =>
             lastAck = Stop
+            composite.cancel()
             composite -= future.cancelable
             self.onError(error)
         }
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
@@ -105,7 +105,7 @@ private[reactive] final class MapParallelOrderedObservable[A, B](
                     composite -= head.cancelable
                   case Failure(ex) =>
                     lastAck = Stop
-                    composite -= head.cancelable
+                    composite.cancel() // cancel the whole downstream on error
                     self.onError(ex)
                 }
 
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
@@ -60,10 +60,8 @@ private[reactive] final class MapParallelOrderedObservable[A, B](
     implicit val scheduler = out.scheduler
     // Ensures we don't execute more than a maximum number of tasks in parallel
     private[this] val semaphore = AsyncSemaphore(parallelism)
-    // Reusable instance for releasing permits on cancel, but
-    // it's debatable whether this is needed, since on cancel
     // everything gets canceled at once
-    private[this] val releaseTask = Task.eval(semaphore.release())
+    private[this] val cancelComposite = Task.eval(composite.cancel())
     // Buffer with the supplied  overflow strategy.
     private[this] val buffer = BufferedSubscriber[B](out, overflowStrategy, MultiProducer)
 
@@ -145,7 +143,7 @@ private[reactive] final class MapParallelOrderedObservable[A, B](
       // we can no longer stream errors downstream
       var streamErrors = true
       try {
-        val task = f(elem).doOnCancel(releaseTask)
+        val task = f(elem).doOnCancel(cancelComposite)
         // No longer allowed to stream errors downstream
         streamErrors = false
         // Start execution (forcing an async boundary)
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelUnorderedObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelUnorderedObservable.scala
@@ -101,6 +101,7 @@ private[reactive] final class MapParallelUnorderedObservable[A, B](
             error => {
               lastAck = Stop
               composite -= subscription
+              composite.cancel()
               self.onError(error)
             },
             value =>
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelUnorderedObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelUnorderedObservable.scala
@@ -71,6 +71,9 @@ private[reactive] final class MapParallelUnorderedObservable[A, B](
     private[this] val semaphore = AsyncSemaphore(parallelism)
     // Buffer with the supplied  overflow strategy.
     private[this] val buffer = BufferedSubscriber[B](out, overflowStrategy, MultiProducer)
+    // everything gets canceled at once
+    private[this] val cancelComposite = Task.eval(composite.cancel())
+
     // Flag indicating whether a final event was called, after which
     // nothing else can happen. It's a very light protection, as
     // access to it is concurrent and not synchronized
@@ -79,8 +82,6 @@ private[reactive] final class MapParallelUnorderedObservable[A, B](
     // coming from the `buffer` - this indicates that the downstream
     // no longer wants any events, so we must cancel
     private[this] var lastAck: Ack = Continue
-    // everything gets canceled at once
-    private[this] val cancelComposite = Task.eval(composite.cancel())
 
     private def process(elem: A) = {
       // For protecting against user code, without violating the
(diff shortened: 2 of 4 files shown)


Avasil:
Thanks @TapanVaishnav Could you add a test for it? Something that if `n` tasks are running in parallel and any of them fails then all the others will be canceled as well. It should be quite simple to do with `TestScheduler` which can be used to manually tick time. e.g.

```scala
val failedTask = Task.raiseError(ex).delayExecution(1.second)
val otherTask = Task.sleep(2.second).doOnCancel(incrementCounter)

Observable(0, 1, 2, 3, 4, 5, 6).mapParallelOrdered(n)(if (i == 0) failedTask else otherTask)
```

and then we can do `s.tick(1.second)` to instantly complete `failedTask` and see if others were canceled

TapanVaishnav:
> Thanks @TapanVaishnav Could you add a test for it? Something that if `n` tasks are running in parallel and any of them fails then all the others will be canceled as well. It should be quite simple to do with `TestScheduler` which can be used to manually tick time. e.g.
> 
> ```scala
> val failedTask = Task.raiseError(ex).delayExecution(1.second)
> val otherTask = Task.sleep(2.second).doOnCancel(incrementCounter)
> 
> Observable(0, 1, 2, 3, 4, 5, 6).mapParallelOrdered(n)(if (i == 0) failedTask else otherTask)
> ```
> 
> and then we can do `s.tick(1.second)` to instantly complete `failedTask` and see if others were canceled

Got it, will add a test. Thanks

TapanVaishnav:
@Avasil Hi, I've added test for MapParallelOrderedSuite and was planning to add one for mapParallelUnorderedSuite but coz the map is unordered(in the latter case) we can't ensure the value of the counter. For example
```
    test("should cancel the whole stream when if one fails") { implicit s =>
    val wasThrown: Throwable = null
    var received = 0

    val failedTask = Task.raiseError(wasThrown).delayExecution(1.second)
    val otherTask = Task.sleep(2.second).doOnCancel(Task { println("otherTask")})

    Observable(0,1,2,3,4,5,6).mapParallelUnordered(4)(i => if (i == 0) failedTask else otherTask)
      .toListL.runToFuture

    s.tick(1.second)

    assertEquals(received, 3)
  }
```
in the above case, coz the map is unordered test might pass/fail, is that right?

Avasil:
I feel like it should still work because it will preserve the order of the input (so `failedTask` should start as the first one and then 1,2,3)

Did you experience non-deterministic behavior here?

TapanVaishnav:
> Did you experience non-deterministic behavior here?

`Observable(0,1,2,3,4,5,6).mapParallelUnordered(4)(i => Task.now(println(i)))
      .toListL.runToFuture`
Yes, I changed my task inside the map to print the order and it's varying
like in one case the order was `1->3->4->2->0->5->6` and in another was `1->4->3->2->0->6->5`

Thought not sure if is this the input order or the order in which Task is getting completed in the respective threads.

TapanVaishnav:
Also, I tried running the https://github.com/monix/monix/pull/1064#issuecomment-547859152 test and out of three cases, it passed in one and failed in others. So it seems like the input order isn't preserved.

Avasil:
In your example the tasks are "instant" so I can see the output happening in this way, I think it would be a little bit more consistent with a small delay

AFAIK `mapParallelUnordered` will start input tasks in the correct order but they can finish in different order and will send any result downstream as soon as it is done

> Also, I tried running the #1064 (comment) test and out of three cases, it passed in one and failed in others. So it seems like the input order isn't preserved.

I wonder if we have a race condition, i.e. https://github.com/monix/monix/blob/befbac7f84452da200af5681dc260c1bfc27334b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelUnorderedObservable.scala#L104 cancels other tasks, they release semaphore and the tasks which were waiting for the permit are starting

Do you see any tasks successfully completing or just more cancels?
I think we could try `guaranteeCase` instead of `doOnCancel` to track all exit cases and see what's going on

TapanVaishnav:
> I think it would be a little bit more consistent with a small delay
> I think we could try guaranteeCase instead of doOnCancel to track all exit cases and see what's going on

I see, will try both, thanks for your input. 👍

TapanVaishnav:
Hi @Avasil, I changed my observable to this `Observable(0,1,2,3,4,5,6).mapParallelUnordered(4)(i => Task(println(i)).delayExecution(1.second)).toListL.runToFuture`
and following are the outputs for two different runs
`0->3->2->1->5->6->4`
`1->2->0->3->6->4->5`

Am I executing my tasks right in the right manner??

Avasil:
> Hi @Avasil, I changed my observable to this `Observable(0,1,2,3,4,5,6).mapParallelUnordered(4)(i => Task(println(i)).delayExecution(1.second)).toListL.runToFuture`
> and following are the outputs for two different runs
> `0->3->2->1->5->6->4`
> `1->2->0->3->6->4->5`
> 
> Am I executing my tasks right in the right manner??

This is correct, note that the first 4 tasks are always from 0-3 range.
0, 1, 2, 3 start and wait 1 second. They run concurrently so they can print in a different order

TapanVaishnav:
> This is correct, note that the first 4 tasks are always from 0-3 range.
0, 1, 2, 3 start and wait 1 second. They run concurrently so they can print in a different order

I see but if that is the case then the below code should produce the same output(=3) as `failedTask` would always be there in the first call, right? But in some cases, it's producing the `received` value as 4 and in some as 3.

```

    test("should cancel the whole stream when if one fails") { implicit s =>
    val wasThrown: Throwable = null
    var received = 0

    val failedTask = Task.raiseError(wasThrown).delayExecution(1.second)
    val otherTask = Task.sleep(2.second).doOnCancel(Task { received+= 1; println("otherTask")})

    Observable(0,1,2,3,4,5,6).mapParallelUnordered(4)(i => if (i == 0) failedTask else otherTask).toListL.runToFuture

    s.tick(1.second)
    assertEquals(received, 6)
  }
```

TapanVaishnav:
Ok, so the output is not the same on each run but at most only the 5th Task is getting called i.e. 
`Observable(0,1,2,3,4,5,6).mapParallelUnordered(4)(i => if (i == 0) failedTask else { Task.now(println(i)) ; otherTask})
    .toListL.runToFuture`
running this produces two kind of outputs
```
3
1
2
// the above order can be different as you mentioned earlier
otherTask
otherTask
otherTask
4
otherTask
```
and 
```
3 
1
2
// the above order can be different as you mentioned earlier
otherTask
otherTask
otherTask
```
will check more on the code.

TapanVaishnav:
@Avasil I think it was as you mentioned about the race condition,
`failedTask` was releasing the semaphore on cancellation and nextTask (element 4) was getting triggered (not sure if it making much sense), so I changed the doOnCancel function 
https://github.com/monix/monix/blob/6dc9a95b110f144892edc4a79d43972ad5344222/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelUnorderedObservable.scala#L120
to cancel the composite instead of releasing the semaphore and it's working as it should have.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
```diff
@@ -105,7 +105,7 @@ private[reactive] final class MapParallelOrderedObservable[A, B](
                     composite -= head.cancelable
                   case Failure(ex) =>
                     lastAck = Stop
-                    composite -= head.cancelable
+                    composite.cancel() // cancel the whole downstream on error
```
Avasil: I think it will be better to `composite.cancel()` in `process` method so as soon as the task is completed with failure

Avasil:
Thank you @TapanVaishnav 
This fix should be fine, I've left one suggestion (it might cause ordered version to need a similar treatment but I'm not sure)

TapanVaishnav:
@Avasil Thank you for your review. I've added `composite.cancel()` in both `process` and `sendDownstreamOrdered` method. Also, created a cancellable task instead of releasing semaphore.
https://github.com/monix/monix/blob/0818687c9e56d182ba1ab42874773a23cd4ba655/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala#L64

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
Avasil: @TapanVaishnav Sorry for nitpicking but this line should be redundant now :D

TapanVaishnav: Ahhh, because `process` calls `sendDownstreamOrdered` which then calls `onNext` and which then calls the `process` again. And if we stop the stream in the `process` itself then need not worry about `sendDownstreamOrdered` method.

TapanVaishnav: @Avasil I have removed the above line and believe that we needn't to add `composite -= head.cancelable` again, right?

TapanVaishnav:
All the tests are passing on my local machine and I didn't even touch the tests which are failing on CI(and don't have any dependency on MapParallel(Un)orderedSuite), 
can you please have a look?

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
```diff
@@ -160,6 +157,7 @@ private[reactive] final class MapParallelOrderedObservable[A, B](
 
           case Failure(error) =>
             lastAck = Stop
+            composite.cancel()
```
Avasil: the order should be different, no gain in `composite -= future.cancelable` after it is canceled

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
Avasil: Yes, I think there are some redundant operators there but it's OK, I have to sit one day and refactor entire operator, it's a bit of a mess :D

Avasil:
> All the tests are passing on my local machine and I didn't even touch the tests which are failing on CI(and don't have any dependency on MapParallel(Un)orderedSuite),
> can you please have a look?

Looks like a random error, I restarted CI

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala
TapanVaishnav: > the order should be different,
Do we even need this line now as we can just cancel the composite?

Avasil: I'm not 100% so I'd prefer to stay with status quo

TapanVaishnav: Hmm, Ok. Done

TapanVaishnav:
@Avasil hey, no worries.
I totally understand your concern, if you don’t nitpick now, the code will become cumbersome to understand later. 
And I’m also in a learning phase so totally appreciate all the comments.

PR #1065, opened 2019-10-31
Add mapAccumulate to Observable

busti:
This Pull Request adds support for the following operator:  
```scala
def mapAccumulate[S, R](seed: => S)(op: (S, A) => (S, R)): Observable[R]
```  

It is similar to `scan` with the difference being that it allows for a separate accumulator and result type.  
It is analogous to the `mapAccumulate` operator found in **fs2**.

I am still uncertain about the naming and the placement of the operator in context to the other operators. The naming in fs2 implies that it is more closely related to `map`, however I personally think that it is more similar to `scan`. I initially called it `scanTo` which is a bit shorter and maybe clearer.  
Also, `scanAccumulate` might be redundant, since `scan` already implies that an accumulation is happening.

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -2875,6 +2875,19 @@ abstract class Observable[+A] extends Serializable { self =>
   final def scan[S](seed: => S)(op: (S, A) => S): Observable[S] =
     new ScanObservable[A, S](self, seed _, op)
 
+  /**
+    * Applies a binary operator to a start value and all elements of
+    * this Observable, going left to right and returns a new
+    * Observable that emits on each step the result element of
+    * the applied function.
+    *
+    * Similar to [[scan]], but the supplied function returns a tuple
+    * of the next accumulator state and the result type emitted by
+    * the returned observable.
+    */
+  final def scanAccumulate[S, R](seed: => S)(op: (S, A) => (S, R)): Observable[R] =
+    new ScanAccumulateObservable[A, S, R](self, seed _, op)
+
   /** Applies a binary operator to a start value and all elements of
     * this Observable, going left to right and returns a new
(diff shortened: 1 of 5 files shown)


Avasil:
Thank you @busti , the implementation looks good to me

>  The naming in fs2 implies that it is more closely related to map, however I personally think that it is more similar to scan. I initially called it scanTo which is a bit shorter and maybe clearer.
Also, scanAccumulate might be redundant, since scan already implies that an accumulation is happening.

I agree and I think we should go with either `scanTo` or `mapAccumulate`. 
I know that ZIO also use `mapAccum` so maybe the method has a bigger history. I asked on fs2 channel if they know the origins of the name.

Avasil:
@busti It seems like the name is derived from Haskell's `mapAccumL` with similar signature:

https://github.com/functional-streams-for-scala/fs2/issues/309

http://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Traversable.html

busti:
Thanks a lot for looking further into that. Then I see no reason why we should not name it like that either. I am personally in favor of `mapAccum` since it is a bit shorter. Thoughts?
I can make these changes this evening once I get back from work.

Avasil:
I'm leaning towards `mapAccumulate` because I don't see any abbreviated names in `Observable` but I don't have strong opinion either way so go with what you prefer :)

cc @oleg-py if he wants to join the bikeshedding :D

busti:
> because I don't see any abbreviated names in `Observable`  

That makes perfect sense. Sorry for the unnecessary bikeshedding.

busti:
I changed the name, but it is still placed underneath `scan`. 
(more bikeshedding incoming) Since the functionality **and** implementation is almost exactly as the one of `scan`, I would personally leave it there.  
When I had to order these functions in the collection library that I am working on at the moment, I simply ordered them alphabetically.

Avasil:
I think the placement is alright. In any case it is something that can be easily changed :)

Thank you for the contribution!

PR #1069, opened 2019-11-05
Introduce ICircuitBreaker interface for CircuitBreaker

### LINKED ISSUE #1068: CircuitBreaker Trait
Hello,
is there a reason why `CircuitBreaker` is not a `trait`? It would be very useful to have it as an interface with multiple possible implementation but mostly for the ability to wrap it and enrich it with extra features. What do you think? I could give it a try and implement it if this is something that could be done.

Thank you.

jakubjanecek:
It is good design to have an interface and implementation for any component.
Trait allows us to pass different circuit breaking implementation in tests
or extend the functionality by wrapping another circuit breaker.

The name was chosen because we cannot introduce CircuitBreaker trait in
backwards-compatible way.

Closes #1068

### DIFF
--- a/monix-catnap/shared/src/main/scala/monix/catnap/ICircuitBreaker.scala
+++ b/monix-catnap/shared/src/main/scala/monix/catnap/ICircuitBreaker.scala
@@ -61,25 +61,6 @@ trait ICircuitBreaker[F[_]] {
     */
   def state: F[CircuitBreaker.State]
 
-  /**
-    * Awaits for this `CircuitBreaker` to be [[CircuitBreaker.Closed closed]].
-    *
-    * This only works if the type class instance used is implementing
-    * [[https://typelevel.org/cats-effect/typeclasses/async.html cats.effect.Async]].
-    *
-    * If this `CircuitBreaker` is already in a closed state, then
-    * it returns immediately, otherwise it will wait (asynchronously) until
-    * the `CircuitBreaker` switches to the [[CircuitBreaker.Closed Closed]]
-    * state again.
-    *
-    * @param F is a restriction for `F[_]` to implement
-    *        `Concurrent[F]` or `Async[F]` (from Cats-Effect). If it
-    *        implements `Concurrent`, then the resulting instance will
-    *        be cancelable, to properly dispose of the registered
-    *        listener in case of cancellation.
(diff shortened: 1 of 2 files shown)


monix-catnap/shared/src/main/scala/monix/catnap/ICircuitBreaker.scala
```diff
@@ -0,0 +1,143 @@
...
+    * the `CircuitBreaker` switches to the [[CircuitBreaker.Closed Closed]]
+    * state again.
+    *
+    * @param F is a restriction for `F[_]` to implement
+    *        `Concurrent[F]` or `Async[F]` (from Cats-Effect). If it
+    *        implements `Concurrent`, then the resulting instance will
+    *        be cancelable, to properly dispose of the registered
+    *        listener in case of cancellation.
+    */
+  def awaitClose(implicit F: Concurrent[F] OrElse Async[F]): F[Unit]
```
Avasil: I wonder about it in the trait because it's not relevant to all implementations but in reality we will probably have just 1 + user wrappers so I don't have a strong opinion on this

jakubjanecek: I took all the public methods from CircuitBreaker and put them into the trait. However you are right that this method does not seem generic enough so I decided to remove it. It can always be added if necessary.

alexandru:
@jakubjanecek it's not necessarily good design to expose OOP interface.

Are you thinking of a use-case? Do you want to implement your own CircuitBreaker?

Avasil:
@alexandru see motivation in https://github.com/monix/monix/issues/1068

jakubjanecek:
@alexandru Could you give me an example why not? As described in #1068 I want to enrich the circuit breaker with logging and metrics and having a trait is very good for that. Also testing piece of code with circuit breaker would be much easier if you can mock it easily. Or is there another way I am missing?

alexandru:
Hey, I gave a reply to that on the issue:

https://github.com/monix/monix/issues/1068#issuecomment-550979390

alexandru:
Ticket at #1068 was closed, I think we can close this for now, but please feel free to reopen if the case.

PR #1070, opened 2019-11-06
Remove task.doOnCancel(cancelComposite) in Observable.mapParallel

Avasil:
@TapanVaishnav  I noticed that `doOnCancel(cancelComposite)` is redundant because a `Task` is a part of composite - so if it gets canceled then the other `Task` will as well. 

I didn't want to bother you because of my oversight :D


TapanVaishnav:
> @TapanVaishnav I noticed that `doOnCancel(cancelComposite)` is redundant because a `Task` is a part of composite - so if it gets canceled then the other `Task` will as well.

I see
> I didn't want to bother you because of my oversight :D

No trouble, I'm ready to help anytime.

Issue #927, opened 2019-06-27
Revise Observable.mapParallelOrdered


Avasil:
[Implementation](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/MapParallelOrderedObservable.scala) is quite complicated and perhaps could be improved.

I would like to see benchmark vs `Observable.mapParallelUnordered` and then if replacing `java.ConcurrentQueue` with the one from `jctools` has any performance benefit.

Ideally we could also simplify implementation but I don't have anything on my mind yet

Avasil:
In both `mapParallelUnordered` and `mapParallelOrdered` we do `f(elem).redeem(...).doOnCancel(...)` and return `Task[Unit]`. 

I wonder if there is a performance difference if we just replace it with `guaranteeCase`.

TapanVaishnav:
@Avasil Will have a look at this one, thank you for the mention and sharing about the issue. You are the best

TapanVaishnav:
Now that we have created the benchmark, I am starting working on different queues.
Right now, I am working with `SpmcArrayQueue`.

Avasil:
@TapanVaishnav I would suggest `LowLevelConcurrentQueue` because available queue can be different depending on the platform. See `LowLevelConcurrentQueueBuilders`

TapanVaishnav:
> I would suggest LowLevelConcurrentQueue because available queue can be different depending on the platform. See LowLevelConcurrentQueueBuilders

Yes, I checked this one but because this is a trait, I thought I'll start with someone already implemented(to get a hold of the problem)
> See LowLevelConcurrentQueueBuilders

Okay, I'll have a look again on `LowLevelConcurrentQueue` family.
Thanks for the suggestion :)

Avasil:
It is used internally in `AsyncQueue` and `ConcurrentQueue` if you are looking for example

PR #1073, opened 2019-11-07
Upgrade to Scala 2.13.1 and 2.12.10

### LINKED ISSUE #1071: Upgrading Scala-2.12.X
Scala2.12.10 is available and fixes the big pattern matching, So would like to upgrade if okay.

Avasil:
Fixes #1071

I'm preparing for a release tonight so I'm bumping it now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,27 +19,27 @@ matrix:
       env: COMMAND=ci-js
     # Scala 2.12, JVM
     - jdk: oraclejdk8
-      scala: 2.12.9
+      scala: 2.12.10
       env: COMMAND=ci-jvm-all
     # Scala 2.12, JVM, JDK 11
     - jdk: openjdk11
-      scala: 2.12.9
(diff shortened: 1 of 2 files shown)


TapanVaishnav:
> I'm preparing for a release tonight so I'm bumping it now

Oh, Okay.

Avasil:
@TapanVaishnav Sorry for taking it if you wanted to upgrade yourself!

If you're looking for something then I was wondering about https://github.com/monix/monix/issues/927 for a while.

For starters, I'd like to have a simple benchmark. We could take a look if there is one for one of Akka Streams/fs2/zio streams to see if they have a good one. Or we could do something like https://github.com/monix/monix/blob/master/benchmarks/shared/src/main/scala/monix/benchmarks/ObservableConcatMapBenchmark.scala
but with `Observable.range(0, size).mapParallelXXX(parallelism)(i => Task.eval(i))`

Issue #1072, opened 2019-11-07
make Observable's drop take a long instead of int


gclaramunt:
I would really love to have
```
final def drop(n: Long): Observable[A] =
    self.liftByOperator(new DropFirstOperator(n))
```
instead of
```
final def drop(n: Int): Observable[A] =
    self.liftByOperator(new DropFirstOperator(n))
```
AFAICT, DropFirstOperator takes a long actually ... I'm more than happy to submit a PR (I can alternative add a dropLong method if you want to avoid changing the interface ... maybe we can have both `final def drop(n: Int): Observable[A]` and `final def drop(n: Long): Observable[A]` but I seem to recall a vague corner case where it was problematic, but I might be wrong, parameter overloading is used a lot in java APIs :) )
If we just replace int with long, current callers might get a "implicit numeric widening" warning, depending on how strict are the compiler options

A less intrusive approach would be 
```
final def dropLong(n: Long): Observable[A] =
    self.liftByOperator(new DropFirstOperator(n))
```
(tho is a tad noisier :) )

PR #1098, opened 2019-12-15
Observable drop long

gclaramunt:
See https://github.com/monix/monix/issues/1072

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -63,3 +63,6 @@ https://github.com/jbracker
 
 Johan Mudsam
 https://github.com/mudsam
+
+Gabriel Claramunt
+https://github.com/gclaramunt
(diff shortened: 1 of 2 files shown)


Avasil:
Thanks for PR @gclaramunt sorry for not following up on it earlier

Unfortunately the change is not binary compatible so we can't replace it until `4.x`

I don't remember the issue about the overloads, I will try to dig it up

I suppose we are stuck with separate method name unless we can find a way to make it compatible

gclaramunt:
Ah, sorry I missed the binary check... I'll try to check the issue too, it might not be a big deal.
Otherwise, I'll change it to dropLong

gclaramunt:
Here are the issues with overloading: https://stackoverflow.com/questions/2510108/why-avoid-method-overloading but I've tested and it doesn't happen with Int and Long (it picks int).
Still, it might be risky ... let me know if you want to got with dropLong or something else

gclaramunt:
cool, thanks! :)

PR #1101, opened 2019-12-29
Canceled tasks in Half-Open state should return to Open state in Circ…

### LINKED ISSUE #1080: CircuitBreaker shoud change state to Open when reset attempt was canceled
https://github.com/monix/monix/blob/b10210bd30c43b95265c818537c24098cbbcab05/monix-catnap/shared/src/main/scala/monix/catnap/CircuitBreaker.scala#L361

In previous step CB's state was set to `HalfOpen`. When task was canceled the state wasn't changed (see CircuitBreaker.scala#L361). In the HalfOpen state CB doesn't try to check F status.

It's look like a bug

Avasil:
…uitBreaker

Fixes #1080


Avasil:
I've published a snapshot: `3.2.0-eb5e5c7-SNAPSHOT`

PR #1106, opened 2020-01-09
Add mapConcat to Observable

fchaillou:
Hello this PR adds
```scala
def [B] mapConcat(f: A: => Seq[B]) : Observable[B]
```
on Observable
which is a shorter version of
```scala
obs.flatMap(a => Observable.fromIterable(f(a))
```

I had to wait on the ack in the `onComplete` to make sure all elements from the last upstream `onNext` are properly propagated before completing our downstream.

Thank you

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -1759,7 +1759,7 @@ abstract class Observable[+A] extends Serializable { self =>
     *
     * ==Example==
     * {{{
-    *   Observable(1, 2, 3).mapConcat( x => Seq(x, x * 10, x * 100))
+    *   Observable(1, 2, 3).mapConcat( x => List(x, x * 10, x * 100))
     * }}}
     *
     * == Visual Example ==
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -55,6 +55,7 @@ import monix.reactive.subjects._
 import org.reactivestreams.{Publisher => RPublisher, Subscriber => RSubscriber}
 
 import scala.collection.mutable
+import scala.collection.immutable
 import scala.concurrent.duration.{Duration, FiniteDuration}
 import scala.concurrent.{Future, Promise}
 import scala.util.{Failure, Success, Try}
@@ -1750,6 +1751,28 @@ abstract class Observable[+A] extends Serializable { self =>
     */
   final def map[B](f: A => B): Observable[B] =
     self.liftByOperator(new MapOperator(f))
+
+  /** Applies a function that you supply to each item emitted by the
+    * source observable, where that function returns a sequence of elements, and
+    * then concatenating those resulting sequences and emitting the
+    * results of this concatenation.
+    *
+    * ==Example==
+    * {{{
+    *   Observable(1, 2, 3).mapConcat( x => Seq(x, x * 10, x * 100))
+    * }}}
+    *
+    * == Visual Example ==
+    *
+    * <pre>
+    * stream: 1 -- -- 2 -- -- 3 -- --
+    * result: 1, 10, 100, 2, 20, 200, 3, 30, 300
+    * </pre>
+    *
+    * @param f is a generator for the sequences being concatenated
+    */
+  final def mapConcat[B](f: A => immutable.Iterable[B]): Observable[B] =
+    self.liftByOperator(new MapConcatOperator(f))
 
   /** Alias for [[concatMap]].
     *
--- a/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/MapConcatSuite.scala
+++ b/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/MapConcatSuite.scala
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2014-2019 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.reactive.internal.operators
+
+import cats.laws._
+import cats.laws.discipline._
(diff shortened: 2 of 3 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -1751,6 +1751,28 @@ abstract class Observable[+A] extends Serializable { self =>
...
+    * {{{
+    *   Observable(1, 2, 3).mapConcat( x => Seq(x, x * 10, x * 100))
+    * }}}
+    *
+    * == Visual Example ==
+    *
+    * <pre>
+    * stream: 1 -- -- 2 -- -- 3 -- --
+    * result: 1, 10, 100, 2, 20, 200, 3, 30, 300
+    * * </pre>
```
Avasil:
```suggestion
    * </pre>
```

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -1751,6 +1751,28 @@ abstract class Observable[+A] extends Serializable { self =>
...
+    *
+    * @param f is a generator for the sequences being concatenated
+    */
+  final def mapConcat[B](f: A => Seq[B]): Observable[B] =
```
Avasil: I'm wondering about two things:
1. Should we use `Iterable` instead of `Seq`? `Seq extends Iterable` but it also introduce some dangerous operations like `size` or `def apply(i: Int): A` and is usually discouraged. On the other hand it is used in many other existing operators...

2. We also have `concatMap` which is an alias for `flatMap`. `concatMap` comes from RX but they don't seem to have anything named `mapConcat` which is consistent with Akka. I guess it's fine then, I'm only slightly worried about potential confusion.

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/MapConcatSuite.scala
```diff
@@ -0,0 +1,110 @@
...
+import monix.execution.Ack
+import monix.execution.Ack.{Continue, Stop}
+import monix.execution.exceptions.DummyException
+import monix.reactive.{Observable, Observer}
+
+import scala.concurrent.duration.Duration.Zero
+import scala.concurrent.duration._
+import scala.concurrent.{Future, Promise}
+
+object MapConcatSuite extends BaseOperatorSuite {
```
Avasil: It feels like we could add property based test `mapConcat(f) <-> flatMap(x => fromIterable(f(x))`

I'm also curious about performance difference but I can check myself later :D

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
fchaillou: 1. I think we should `Iterable` as well, I will go for `scala.collection.immutable.Iterable` that's what makes more sense and after checking that is also exactly what akka-streams is using.

2. For `concatMap` vs `mapConcat` i have to admit it might be a bit confusing but hopefully the signature and doc should help to allievate the confusion

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/MapConcatSuite.scala
fchaillou: Good idea.
i can take care of that next week if  you don't mind waiting

fchaillou: Do you have any pointers on doing that with minitest ?
thanks

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
fchaillou: Pushed the changes for using scala.collection.immutable, only remaining is the property based test

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/MapConcatSuite.scala
Avasil: Take a look at [ReduceSuite](https://github.com/monix/monix/blob/master/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/ReduceSuite.scala#L63)

fchaillou: Thanks, just pushed a test, let me know if that makes sense for you ?

Avasil: I think the test looks great :)

Avasil:
Thank you a lot, there are some compile errors after changing `Seq` to `Iterable` - once you fix it it's good to merge :)

fchaillou:
Oh, sbt kept the cache versions locally so i missed those.
This should hopefully be good now :)

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -1751,6 +1752,28 @@ abstract class Observable[+A] extends Serializable { self =>
...
     self.liftByOperator(new MapOperator(f))
 
+  /** Applies a function that you supply to each item emitted by the
+    * source observable, where that function returns a sequence of elements, and
+    * then concatenating those resulting sequences and emitting the
+    * results of this concatenation.
+    *
+    * ==Example==
+    * {{{
+    *   Observable(1, 2, 3).mapConcat( x => Seq(x, x * 10, x * 100))
```
Avasil:
```suggestion
    *   Observable(1, 2, 3).mapConcat( x => List(x, x * 10, x * 100))
```

fchaillou: Oh i didn't realize the doc were tested !
That is pretty great actually :)
Do you have any pointers on what you're using to achieve that ?
I would be interested to reuse it
Thanks !

allantl:
@Avasil @fchaillou FYI, `flatMap` is also aliased as `concatMap`

fchaillou:
@allantl yesm @Avasil mentionned that earlier in the PR, hopefully the type signature and the documentation will prevent confusion between the two !

Avasil:
@allantl We discussed it here: https://github.com/monix/monix/pull/1106#discussion_r365449263

Nothing has been released yet so it's not too late if you have an idea for a better name but my head is empty

allantl:
In RxJava, its named as `concatMapIterable` [link](http://reactivex.io/RxJava/3.x/javadoc/snapshot/io/reactivex/rxjava3/core/Observable.html#concatMapIterable-io.reactivex.rxjava3.functions.Function-). There's also `flatMapIterable`. What do you guys think about that?

Avasil:
Ah, I missed it. Thank you

If it already has a name in RX then we should be consistent and have an alias at the very least
I kinda like `mapConcat` more than longer names but the confusion is a concern

fchaillou:
Hey @Avasil,
I feel like i forgot about the eval end evalF version !
I will look into them sometime next week as well.
Fabien

PR #1117, opened 2020-01-31
Upgrade cats-effect to 2.1.0, add SyncEffect instance for Coeval

oleg-py:
Since cats-effect 2.1.0 drops Scala 2.11, I have to cross-build against different versions of it.

So, SyncEffect[Coeval] is only available in 2.12+.
Also, Resource is now covariant, and it wrecks the inference in pattern-matching that we use to make Iterant and Observable out of Resource. I've changed it to use unchecked matches, but we can also just delegate to `resource.allocated` instead of recursively decomposing it.


build.sbt
```diff
@@ -27,8 +27,14 @@ addCommandAlias("ci-jvm-all",  s";ci-jvm-mima ;unidoc")
 addCommandAlias("release",     ";project monix ;+clean ;+package ;+publishSigned")
 
 val catsVersion = "2.0.0"
-val catsEffectVersion = "2.0.0"
-val catsEffectLawsVersion = catsEffectVersion
+lazy val catsEffectVersion = settingKey[String]("cats-effect version")
+ThisBuild/catsEffectVersion := {
+  CrossVersion.partialVersion(scalaVersion.value) match {
+    case Some((2, 11)) => "2.0.0"
+    case _ => "2.1.0"
```
Avasil:
```suggestion
    case _ => "2.1.1"
```

Avasil:
Seems like there is something wrong with `Observable` doc, I'm not sure what has changed though

```
[error] /home/travis/build/monix/monix/monix-reactive/shared/src/main/scala/monix/reactive/observables/CombineObservable.scala:24:1: Could not find any member to link for "cats.NonEmptyParallel".
[error] /** Newtype encoding for an [[Observable]] datatype that has a [[cats.Apply]]
[error] ^
[error] /home/travis/build/monix/monix/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala:6116:3: Could not find any member to link for "cats.NonEmptyParallel".
[error]   /** [[cats.NonEmptyParallel]] instance for [[Observable]]. */
[error]   ^
[error] two errors found
[error] (Scalaunidoc / doc) Scaladoc generation failed
[error] Total time: 40 s, completed Jan 31, 2020 2:07:43 PM
```

Avasil:
I allowed myself to push to your branch, I don't know why `unidoc` can't resolve few links on Scala 2.12 but it's fine on 2.13. It should be good enough

PR #1120, opened 2020-02-03
Cancel should always wait for the finalizer

### LINKED ISSUE #1086: TaskApp can exit prematurely without releasing all resources
Reported on gitter: https://gitter.im/monix/monix?at=5ddbe5e4d2c92c2275f3d0da

I think the gist of it is:

```scala

Task.race(
  Task.sleep(1.seconds).bracket(_ => Task.unit)(_ => Task { println("release") }),
  Task.sleep(100.millis)
).flatMap(_ => Task(println("end"))).runSyncUnsafe()
```
Outputs:

```
end

Process finished with exit code 0
```

Which can lead to a resource leak and is not consistent with `cats.effect.IO`:

```scala
implicit val timer = IO.timer(global)
implicit val cs = IO.contextShift(global)

IO.race(
  IO.sleep(1.seconds).bracket(_ => IO.unit)(_ => IO { println("release") }),
  IO.sleep(100.millis)
).flatMap(_ => IO(println("end"))).unsafeRunSync()
```

```
release
end

Process finished with exit code 0
```

### LINKED ISSUE #1124: Resource with Task works differently than with IO when cancelled
Hello,
I've recently encountered a problem with `Resource` using `Task`. When cancelling the fibre, resource cancellation mechanics doesn't seem to obey the bracket laws. 

Please take a look at code examples with results below:

Lib versions:
monix-eval: 3.1.0
cats-core: 2.1.0
cats-effect: 2.1.1

Example code using `Task` (just replace TaskApp and Task with IOApp and IO for IO usage):
```scala
object ResourceTask extends TaskApp {

  override def run(args: List[String]): Task[ExitCode] = for {
    _ <- runProgram[Task]
    _ <- sleep[Task](10.seconds)
  } yield ExitCode.Success

  def runProgram[F[_] : Sync : Concurrent : Timer]: F[Unit] = {
    for {
      fiber <- resourceProgram[F].start
      _ <- sleep[F](5.seconds)
      _ <- fiber.cancel // Cancel the program during step 2 acquisition
      _ <- log[F]("Fiber has been cancelled")
    } yield ()
  }

  def resourceProgram[F[_] : Sync : Timer]: F[Unit] = {
    (for {
      _ <- Resource.make(log[F]("Step 1 acquire start") *> sleep[F](2.seconds) *> log[F]("Step 1 acquire stop"))(
        _ => log[F]("Step 1 release") *> sleep[F](2.seconds))
      _ <- Resource.make(log[F]("Step 2 acquire start") *> sleep[F](5.seconds) *> log[F]("Step 2 acquire stop"))(
        _ => log[F]("Step 2 release") *> sleep[F](5.seconds))
      _ <- Resource.make(log[F]("Step 3 acquire"))(_ => log[F]("Step 3 release"))
      _ <- Resource.make(log[F]("Step 4 acquire"))(_ => log[F]("Step 4 release"))
    } yield ()).allocated.flatMap(_._2)
  }

  def log[F[_] : Sync](message: String): F[Unit] = Sync[F].delay(println(message))

  def sleep[F[_] : Timer](duration: FiniteDuration): F[Unit] = Timer[F].sleep(duration)
}
```

Results using `Task`:
```
Step 1 acquire start
Step 1 acquire stop
Step 2 acquire start
Step 1 release
Fiber has been cancelled
Step 2 acquire stop
Step 2 release
Step 3 acquire
Step 3 release
Step 4 acquire
Step 4 release
```

Results using `IO`:
```
Step 1 acquire start
Step 1 acquire stop
Step 2 acquire start
Step 2 acquire stop
Step 2 release
Step 1 release
Fiber has been cancelled
```

In both examples, I'm cancelling the fibre during the acquisition phase of `Step 2`. According to `Bracket` laws, `acquire` in uncancellable which seems to work fine. But when using `Task` the `release` phase of `Step 1` is executed when `acquire` of step 2 is not finished yet. It should not happen because resource acquired during `Step 1` can still be used during `Step 2` acquisition.

The second issue I have is that after cancellation, `Step 3` and `Step 4` are still executed. It seems to work just fine when using `IO` instead of `Task`.

Avasil:
Port of my change in https://github.com/typelevel/cats-effect/pull/727 but we also support `tryReactivate` to implement `onCancelRaiseError`. Prior to the PR we backpressured finalizers only for the first action

I'm not sure how to handle `def tryReactivate(): Boolean`, I did something that might work but any suggestions welcome. It doesn't wait for finalizers BUT the method is private and only exists for `onCancelRaiseError` in `conn.cancel.map(_ => ...)` so it shouldn't be an issue.

Fixes #1086 and fixes #1124

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/internal/ForwardCancelable.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/internal/ForwardCancelable.scala
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2014-2020 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.eval.internal
+
+import java.util.concurrent.atomic.AtomicReference
+
+import cats.effect.CancelToken
+import monix.eval.Task
+import monix.execution.schedulers.TrampolineExecutionContext
+import monix.execution.{Callback, Scheduler}
+
+import scala.annotation.tailrec
+import scala.concurrent.ExecutionContext
+import scala.util.control.NonFatal
+
+/**
+  * A placeholder for a [[cats.effect.CancelToken]] that will be set at a later time,
+  * the equivalent of a `Deferred[Task, CancelToken]`.
+  *
+  * Used in the implementation of `bracket`, see [[TaskBracket]].
+  */
+final private[internal] class ForwardCancelable private () {
+  import ForwardCancelable._
+
+  private[this] val state = new AtomicReference[State](init)
+
+  val cancel: CancelToken[Task] = {
+    @tailrec def loop(ctx: Task.Context, cb: Callback[Throwable, Unit]): Unit =
+      state.get() match {
+        case current @ Empty(list) =>
+          if (!state.compareAndSet(current, Empty(cb :: list)))
+            loop(ctx, cb)
+
+        case Active(token) =>
+          state.lazySet(finished) // GC purposes
+          context.execute(new Runnable {
+            def run() =
+              Task.unsafeStartNow(token, ctx, cb)
+          })
+      }
+
+    Task.Async(loop)
+  }
+
+  def complete(value: CancelToken[Task])(implicit s: Scheduler): Unit =
+    state.get() match {
+      case current @ Active(_) =>
+        value.runAsyncAndForget
+        throw new IllegalStateException(current.toString)
+
+      case current @ Empty(stack) =>
+        if (current eq init) {
+          // If `init`, then `cancel` was not triggered yet
+          if (!state.compareAndSet(current, Active(value)))
+            complete(value)
+        } else {
+          if (!state.compareAndSet(current, finished))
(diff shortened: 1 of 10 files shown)


Avasil:
> Ok, I have very little understanding of what it actually does, but the tests look reasonable :)

My understanding of finalizers implementation goes like this:

1. If it's a success or a failure then finalizers goes after flatMap/errorHandle like a normal `Task`
2. Finalizers are added to the `TaskConnection` stack so if `Task` is canceled then they are invoked and backpressured via `TaskConnection#cancel`
3. Whatever causes finalizers to run, it marks `Atomic Reference` that the task is no longer waiting for any results and finalizers were executed

The problem was that after setting `Atomic Reference` any next cancel would become a no-op even if finalizers are still in progress so I just added a `Promise` there to wait for it. Now that I think about it, we can probably remove `Atomic` and just check if `Promise` is completed, I'll check it in a moment.

I added `ForwardCancelable` from `cats.effect.IO` implementation because I think there was an issue that a `Task` could be canceled before a finalizer was installed

Sorry for chaotic explanation but I hope it adds a bit more context :D
My hope is that @alexandru could find some time to review it. 
I already went through review in `cats-effect` but it would give me more confidence

alexandru:
Would like to do a review — please allow me until Monday, or if I go silent again, then proceed without me 😀

Avasil:
Awesome @alexandru I will really appreciate it! During my work on `monix-bio` I became more confident with internals but it's hard to find anyone to do proper review and confirm my understanding :D

I'm about to update PR because I also recognized an issue with our parallel operators.
We currently do:

```scala
conn.cancel.runAsyncAndForget
cb.onError(ex)
```
which causes some problems, e.g.
```scala

Task.race(
  Task.sleep(1.seconds).bracket(_ => Task.unit)(_ => Task { println("release") }),
  Task.sleep(100.millis)
).flatMap(_ => Task(println("end"))).runSyncUnsafe()
```

will exit the program without waiting for `release`. 
cats.effect.IO waits for `release` here so the change will be consistent

Avasil:
@alexandru I am merging it but if you find time to go through it then we could just do a follow up PR :D

PR #1125, opened 2020-02-14
Rename mapConcat to concatMapIterable

Avasil:
cc @fchaillou @allantl 

I'm still wondering about `mapConcat` alias for consistency with Akka Streams but we should follow rx as the first priority


fchaillou:
It looks good to me, and after sleeping on it a few days, I think it makes more sense to keep consistency with the rest of the Monix.

PR #1132, opened 2020-02-18
Add withFilter to Iterant

mantasindrasius:
Cousin of https://github.com/monix/monix/pull/1129

### DIFF
--- a/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
+++ b/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
@@ -816,6 +816,25 @@ sealed abstract class Iterant[F[_], A] extends Product with Serializable {
   final def filter(p: A => Boolean)(implicit F: Sync[F]): Iterant[F, A] =
     IterantFilter(this, p)(F)
 
+  /** Filters the iterant by the given predicate function, returning
+    * only those elements that match.
+    *
+    * Example: {{{
+    *   import monix.eval.Task
+    *
+    *   // Yields 2, 4, 6
+    *   Iterant[Task].of(1, 2, 3, 4, 5, 6).withFilter(_ % 2 == 0)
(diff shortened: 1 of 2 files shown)


monix-tail/shared/src/main/scala/monix/tail/Iterant.scala
```diff
@@ -816,6 +816,25 @@ sealed abstract class Iterant[F[_], A] extends Product with Serializable {
   final def filter(p: A => Boolean)(implicit F: Sync[F]): Iterant[F, A] =
     IterantFilter(this, p)(F)
 
+  /** Filters the iterant by the given predicate function, returning
+    * only those elements that match.
+    *
+    * Example: {{{
```
alexandru: In this ScalaDoc we could simply mention that it's an alias of `filter`, which is better because people need to know that the main function is. 

If we mention anything in this documentation, we should mention the reason for its existence (i.e. to make it compatible with for-comprehensions).

Avasil: In `Observable` PR: https://github.com/monix/monix/pull/1129
I did
```
  /** Alias to [[filter]] to support syntax in for comprehension, i.e.
    *
    * {{{
    *   case class Person(age: Long)
    *
    *   val peopleObservable: Observable[Person] =
    *     Observable.range(1, 100).map(Person.apply)
    *
    *   for {
    *     adult <- peopleObservable if adult.age >= 18
    *   } yield adult
    * }}}
    */
```

alexandru: Yeah, that's good.

Avasil: Okay, I'll just update that PR since I have already merged this one

PR #1136, opened 2020-02-21
Flatmap iterable eval

fchaillou:
Hello,
Following the addition of concatMapIterable, I realized that their async version would be really useful as well, mostly reduce the code when using it which feels makes it more readable.

I'm reusing the other operators here, not sure if we have to create one specifically for that use case.
If you think this is better, I could do with a bit of guidelines.
Thank  you


Avasil:
Thanks for PR, the docs are really good!

I'm just wondering if it's worth it to add four extra methods which are easy to write for an operator that is (probably) not very common. What do you think?

fchaillou:
Maybe the F variant are not needed and we can go with only one like `flatmapEvalIterable` ?
I would use it quite often but I agree that this is not vital :)

Issue #1130, opened 2020-02-18
Add Task.parSequence and Task.parTraverse aliases


Avasil:
Task.parSequence = Task.gather
Task.parTraverse = Task.wander

alexandru:
I'm starting to dislike the proliferation of aliases in Monix.

WDYT? Should we deprecate the old names? We'll keep them for binary compatibility of course.

Avasil:
I'm fine with deprecation in this example.
I suppose we should add `parXXXUnordered` and `parXXXN` as well?
I think in Cats it's called `parUnorderedTraverse` but we use `wanderUnordered` and `wanderN`

1. If we are consistent with ourselves:
`parTraverse`, `parTraverseN`, `parTraverseUnordered`

2. If we are consistent with Cats:
`parTraverse`, `parTraverseN`, `parUnorderedTraverse`

It's unfortunate but I would go with 1 because discoverability seems to be better

jozic:
+1 for `parTraverseUnordered`

jozic:
I'll take this if no objections

PR #1145, opened 2020-03-15
Deprecate gather, gatherN, gatherUnordered in favor of parSequence, parSequenceN, parSequenceUnordered

jozic:
addresses #1130 

- [x] gatherX to parSequenceX
- [x] wanderX to parTraverseX
- [ ] update https://github.com/monix/monix.io

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -169,7 +169,7 @@ import scala.util.{Failure, Success, Try}
   * it does for `Future.sequence`, the given `Task` values being
   * evaluated one after another, in ''sequence'', not in ''parallel''.
   * If you want parallelism, then you need to use
-  * [[monix.eval.Task.gather Task.gather]] and thus be explicit about it.
+  * [[monix.eval.Task.parSequence Task.parSequence]] and thus be explicit about it.
   *
   * This is great because it gives you the possibility of fine tuning the
   * execution. For example, say you want to execute things in parallel,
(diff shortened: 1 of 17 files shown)


jozic:
@Avasil @oleg-py @alexandru 
please check this out, if this looks good, i can continue with the rest

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -3517,36 +3517,46 @@ object Task extends TaskInstancesLevel1 {
     implicit bf: BuildFrom[M[A], B, M[B]]): Task[M[B]] =
     TaskSequence.traverse(in, f)(bf)
 
+  /** Alias for [[parSequence]] */
+  @deprecated("Use parSequence", "3.2.0")
+  def gather[A, M[X] <: Iterable[X]](in: M[Task[A]])(implicit bf: BuildFrom[M[Task[A]], A, M[A]]): Task[M[A]] =
```
Avasil: We should put deprecated methods to `TaskDeprecated.Companion` to clean normal API

There is also slightly different convention for description with
```
/**
   * DEPRECATED — please use XXX
   */
```

jozic:
@Avasil thanks for feedback!
 I've moved deprecated methods to `TaskDeprecated.Companion` and also added renamed classes to the mima filter
I've also tried to revert changes to benchmarks, but current version wasn't compilable due to deprecation errors, so I've disabled `benchmarksPrev`
CI seems happy now

Please check it out again

Avasil:
@jozic Looks good, we can add the rest :+1:

I think we can wait with `monix.io` docs update until `3.2.0` is released

jozic:
@Avasil I've added second commit with changes for `wander` and friends
lmk if anything need to be changed

PR #1146, opened 2020-03-21
Add option to allow disabling isolating Local context on run

mdedetrich:
This PR adds an options to `Task.Options` which allows you to keep the current Local context when running to a Future, this solves the issue described here https://github.com/monix/monix/issues/848#issuecomment-568891708.

The new option is called `isolateLocalContextOnRun` and by default it is `true` (which keeps current behavior). If you set it to `false` then when running a `Task` to create a `Future`, the `Local` context propagation is not isolated so you can continue getting/setting values from the `Local` afterwards.

Note that a simple test was added to verify its working however some more tests may be needed? @Avasil @alexandru @oleg-py

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -36,6 +36,7 @@ import org.reactivestreams.Publisher
 import scala.annotation.unchecked.{uncheckedVariance => uV}
 import scala.concurrent.duration.{Duration, FiniteDuration, NANOSECONDS, TimeUnit}
 import scala.concurrent.{ExecutionContext, Future, TimeoutException}
+import scala.runtime.AbstractFunction2
 import scala.util.{Failure, Success, Try}
 
 /** `Task` represents a specification for a possibly lazy or
@@ -582,7 +583,7 @@ sealed abstract class Task[+A] extends Serializable with TaskDeprecated.BinCompa
   def runToFutureOpt(implicit s: Scheduler, opts: Options): CancelableFuture[A] = {
     val opts2 = opts.withSchedulerFeatures
     Local
-      .bindCurrentIf(opts2.localContextPropagation) {
+      .bindCurrentIf(opts2.localContextPropagation && opts2.localContextIsolateOnRun) {
         TaskRunLoop.startFuture(this, s, opts2)
       }
   }
@@ -706,7 +707,7 @@ sealed abstract class Task[+A] extends Serializable with TaskDeprecated.BinCompa
   @UnsafeBecauseImpure
   def runAsyncOpt(cb: Either[Throwable, A] => Unit)(implicit s: Scheduler, opts: Options): Cancelable = {
     val opts2 = opts.withSchedulerFeatures
-    Local.bindCurrentIf(opts2.localContextPropagation) {
+    Local.bindCurrentIf(opts2.localContextPropagation && opts2.localContextIsolateOnRun) {
       UnsafeCancelUtils.taskToCancelable(runAsyncOptF(cb)(s, opts2))
     }
   }
@@ -808,7 +809,7 @@ sealed abstract class Task[+A] extends Serializable with TaskDeprecated.BinCompa
   @UnsafeBecauseImpure
   def runAsyncOptF(cb: Either[Throwable, A] => Unit)(implicit s: Scheduler, opts: Options): CancelToken[Task] = {
     val opts2 = opts.withSchedulerFeatures
-    Local.bindCurrentIf(opts2.localContextPropagation) {
+    Local.bindCurrentIf(opts2.localContextPropagation && opts2.localContextIsolateOnRun) {
       TaskRunLoop.startLight(this, s, opts2, Callback.fromAttempt(cb))
     }
   }
@@ -930,7 +931,7 @@ sealed abstract class Task[+A] extends Serializable with TaskDeprecated.BinCompa
   @UnsafeBecauseImpure
   def runAsyncUncancelableOpt(cb: Either[Throwable, A] => Unit)(implicit s: Scheduler, opts: Task.Options): Unit = {
     val opts2 = opts.withSchedulerFeatures
-    Local.bindCurrentIf(opts2.localContextPropagation) {
+    Local.bindCurrentIf(opts2.localContextPropagation && opts2.localContextIsolateOnRun) {
       TaskRunLoop.startLight(this, s, opts2, Callback.fromAttempt(cb), isCancelable = false)
     }
   }
@@ -995,7 +996,7 @@ sealed abstract class Task[+A] extends Serializable with TaskDeprecated.BinCompa
   @UnsafeBecauseImpure
   final def runSyncStepOpt(implicit s: Scheduler, opts: Options): Either[Task[A], A] = {
     val opts2 = opts.withSchedulerFeatures
-    Local.bindCurrentIf(opts2.localContextPropagation) {
+    Local.bindCurrentIf(opts2.localContextPropagation && opts2.localContextIsolateOnRun) {
       TaskRunLoop.startStep(this, s, opts2)
     }
   }
@@ -1093,7 +1094,7 @@ sealed abstract class Task[+A] extends Serializable with TaskDeprecated.BinCompa
   ): A = {
     /*_*/
     val opts2 = opts.withSchedulerFeatures
-    Local.bindCurrentIf(opts2.localContextPropagation) {
+    Local.bindCurrentIf(opts2.localContextPropagation && opts2.localContextIsolateOnRun) {
       TaskRunSyncUnsafe(this, timeout, s, opts2)
     }
     /*_*/
@@ -4247,11 +4248,23 @@ object Task extends TaskInstancesLevel1 {
     *        case you want the [[monix.execution.misc.Local Local]]
     *        variables to be propagated on async boundaries.
     *        Defaults to `false`.
+    *
+    * @param localContextIsolateOnRun should be set to
+    *        `true` if when using `localContextPropagation` you
+    *        want the [[monix.execution.misc.Local Local]] to be
+    *        isolated when running the task otherwise the
+    *        [[monix.execution.misc.Local Local]] will be visible after
+    *        running to a Future. Defaults to `true`.
     */
   final case class Options(
     autoCancelableRunLoops: Boolean,
-    localContextPropagation: Boolean
+    localContextPropagation: Boolean,
+    localContextIsolateOnRun: Boolean
   ) {
+    def this(autoCancelableRunLoops: Boolean,
+             localContextPropagation: Boolean) =
+      this(autoCancelableRunLoops, localContextPropagation, true)
+
     /** Creates a new set of options from the source, but with
       * the [[autoCancelableRunLoops]] value set to `true`.
       */
@@ -4276,6 +4289,18 @@ object Task extends TaskInstancesLevel1 {
     def disableLocalContextPropagation: Options =
       copy(localContextPropagation = false)
 
+    /** Creates a new set of options from the source, but with
+      * the [[localContextIsolateOnRun]] value set to `true`.
+      */
+    def enableLocalContextIsolateOnRun: Options =
+      new Options(autoCancelableRunLoops, localContextPropagation, localContextIsolateOnRun = true)
+
+    /** Creates a new set of options from the source, but with
+      * the [[localContextIsolateOnRun]] value set to `false`.
+      */
+    def disableLocalContextIsolateOnRun: Options =
+      new Options(autoCancelableRunLoops, localContextPropagation, localContextIsolateOnRun = false)
+
     /**
       * Enhances the options set with the features of the underlying
       * [[monix.execution.Scheduler Scheduler]].
@@ -4291,6 +4316,16 @@ object Task extends TaskInstancesLevel1 {
       else
         copy(localContextPropagation = wLocals || localContextPropagation)
     }
+
+    def copy(autoCancelableRunLoops: Boolean = this.autoCancelableRunLoops,
+             localContextPropagation: Boolean = this.localContextPropagation): Options =
+      new Options(autoCancelableRunLoops, localContextPropagation, this.localContextIsolateOnRun)
+  }
+
+  object Options extends AbstractFunction2[Boolean, Boolean, Options] {
+    def apply(autoCancelableRunLoops: Boolean,
+              localContextPropagation: Boolean): Options =
+      new Options(autoCancelableRunLoops, localContextPropagation)
   }
 
   /** Default [[Options]] to use for [[Task]] evaluation,
@@ -4308,12 +4343,15 @@ object Task extends TaskInstancesLevel1 {
     *  - `monix.environment.localContextPropagation`
     *    (`true`, `yes` or `1` for enabling)
     *
+    *  - `monix.environment.localContextIsolateOnRun`
+    *    (`false`, `no` or `0` for disabling)
     * @see [[Task.Options]]
     */
   val defaultOptions: Options =
     Options(
       autoCancelableRunLoops = Platform.autoCancelableRunLoops,
-      localContextPropagation = Platform.localContextPropagation
+      localContextPropagation = Platform.localContextPropagation,
+      localContextIsolateOnRun = Platform.localContextIsolateOnRun
     )
 
   /** The `AsyncBuilder` is a type used by the [[Task.create]] builder,
--- a/monix-eval/jvm/src/test/scala/monix/eval/TaskLocalJVMSuite.scala
+++ b/monix-eval/jvm/src/test/scala/monix/eval/TaskLocalJVMSuite.scala
@@ -277,4 +277,18 @@ object TaskLocalJVMSuite extends SimpleTestSuite {
 
     test.runToFuture
   }
+
+    testAsync("With disableLocalContextIsolateOnRun and enableLocalContextPropagation a Task local should retain its context when running the Task") {
+      import monix.execution.Scheduler.Implicits.global
+      implicit val opts = Task.defaultOptions.enableLocalContextPropagation.disableLocalContextIsolateOnRun
+
+      val local = Local(0)
+
+      val f = for {
+        _ <- Task { local := 50}.asyncBoundary.runToFutureOpt
+        v <- Future { local() }
+      } yield v
+
+      for (v <- f) yield assertEquals(v, 50)
+  }
 }
--- a/monix-eval/shared/src/test/scala/monix/eval/TaskEffectInstanceSuite.scala
+++ b/monix-eval/shared/src/test/scala/monix/eval/TaskEffectInstanceSuite.scala
@@ -31,7 +31,8 @@ object TaskEffectInstanceSuite extends BaseTestSuite {
   test("Effect instance should make use of implicit TaskOptions") { implicit sc =>
     implicit val customOptions: Task.Options = Task.Options(
       autoCancelableRunLoops = true,
-      localContextPropagation = true
+      localContextPropagation = true,
+      localContextIsolateOnRun = true
     )
 
     var received: Task.Options = null
--- a/monix-execution/js/src/main/scala/monix/execution/internal/Platform.scala
+++ b/monix-execution/js/src/main/scala/monix/execution/internal/Platform.scala
@@ -75,6 +75,12 @@ private[monix] object Platform {
   final val localContextPropagation: Boolean = false
 
   /**
+    * Isolating local context when running to Future is
+    * set to `true` by default.
+    */
+  final val localContextIsolateOnRun: Boolean = true
+
+  /**
     * Establishes the maximum stack depth for fused `.map` operations
     * for JavaScript.
     *
--- a/monix-execution/jvm/src/main/scala/monix/execution/internal/Platform.scala
+++ b/monix-execution/jvm/src/main/scala/monix/execution/internal/Platform.scala
@@ -132,6 +132,19 @@ private[monix] object Platform {
       .exists(v => v == "yes" || v == "true" || v == "1")
 
   /**
+    * Default value for isolating local context when running to a Future
+    * is set to true. On top of the JVM the default can be overridden by
+    * setting the following system property:
+    *
+    *  - `monix.environment.localContextIsolateOnRun`
+    *    (`true`, `yes` or `1` for enabling)
+    */
+  val localContextIsolateOnRun: Boolean =
+    Option(System.getProperty("monix.environment.localContextIsolateOnRun", ""))
+      .map(_.toLowerCase)
+      .forall(v => v != "no" && v != "false" && v != "0")
+
+  /**
     * Establishes the maximum stack depth for fused `.map` operations.
     *
     * The default is `128`, from which we subtract one as an


monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -4249,10 +4249,18 @@ object Task extends TaskInstancesLevel1 {
...
     *        Defaults to `false`.
+    * 
+    * @param isolateLocalContextOnRunToFuture should be set to
+    *        `true` if when using `localContextPropagation` you
+    *        want the [[monix.execution.misc.Local Local]] to be
+    *        isolated when running the task otherwise the
+    *        [[monix.execution.misc.Local Local]] will be visible after
+    *        running to a Future. Defaults to `true`.
     */
   final case class Options(
```
Avasil: If you follow Oleg's PR: https://github.com/monix/monix/pull/1044/files
then we should be able to preserve binary compatibility for at least Scala 2.12 and 2.13.

You can check if it passes with something like:
```scala
++ 2.12.10
evalJVM/mimaReportBinaryIssues
```

I feel like we can get away with only the following in companion object (+ similar changes in case class):
```scala
  object Options extends AbstractFunction2[Boolean, Boolean, Options] {
    def apply(autoCancelableRunLoops: Boolean, localContextPropagation: Boolean) =
      new Options(autoCancelableRunLoops, localContextPropagation)
  }
```

But I'm not 100% sure yet.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -4249,10 +4249,18 @@ object Task extends TaskInstancesLevel1 {
     *        case you want the [[monix.execution.misc.Local Local]]
     *        variables to be propagated on async boundaries.
     *        Defaults to `false`.
+    * 
+    * @param isolateLocalContextOnRunToFuture should be set to
```
Avasil:
```suggestion
    * @param isolateLocalContextOnRun should be set to
```

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -4249,10 +4249,18 @@ object Task extends TaskInstancesLevel1 {
...
+    *        want the [[monix.execution.misc.Local Local]] to be
+    *        isolated when running the task otherwise the
+    *        [[monix.execution.misc.Local Local]] will be visible after
+    *        running to a Future. Defaults to `true`.
     */
   final case class Options(
     autoCancelableRunLoops: Boolean,
-    localContextPropagation: Boolean
+    localContextPropagation: Boolean,
+    isolateLocalContextOnRun: Boolean
```
Avasil: A little bikeshedding: 
Should we call it `localContextIsolateOnRun` to be prefix everything with `localContext`? In other PR, Oleg suggested `isolateLocalContextOnFork` for auto-isolation during forking fibers (so `start`, `parMapN` etc.). It's not merged though

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -148,14 +150,17 @@ object Local extends LocalCompanionDeprecated {
...
       val CanBindLocals = symbolOf[CanBindLocals[_]].companion
 
       resetTree(
         q"""if (!$b) { $f } else {
            import $CanBindLocals.Implicits.synchronousAsDefault
-           $Local.isolate($f)
+           if ($isolate)
+             $Local.isolate($f)
+           else
+             $Local.bind($Local.getContext())($f)
```
Avasil: is it necessary to `bind` if we don't isolate?

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -132,9 +132,11 @@ object Local extends LocalCompanionDeprecated {
     }
 
   /** If `b` evaluates to `true`, execute a block of code using a current
-    * state of `Local.Context` and restore the current state when complete.
+    * state of `Local.Context` and if `isolate` evalutes to true then
+    * restore the current state when complete otherwise keep the current
+    * state.
     */
-  private[monix] def bindCurrentIf[R](b: Boolean)(f: => R): R =
+  private[monix] def bindCurrentIf[R](b: Boolean, isolate: Boolean)(f: => R): R =
```
Avasil: method is private so we can just add
```
exclude[DirectMissingMethodProblem]("monix.execution.misc.Local#Macros.localLetCurrentIf")
```

to `MimaFilters.scala`

monix-execution/jvm/src/test/scala/monix/execution/misc/LocalJVMSuite.scala
```diff
@@ -94,6 +94,22 @@ object LocalJVMSuite extends SimpleTestSuite {
     for (v <- f) yield assertEquals(v, 50)
   }
 
+    testAsync("Local.bindCurrentIf(CancelableFuture) should properly keep current context during async boundaries on isolate=false") {
```
oleg-py: I think this is a bad test.
- It's important that we get the use-case working (`Task.runToFuture.flatMap { ... future that reads local }`), not the internal package-private method
- Shouldn't be using single thread scheduler. If it does not work on multiple threads, then the use-case is broken which is something better seen in tests.

mdedetrich:
Thanks for the comments, will look into this today!

monix-execution/jvm/src/test/scala/monix/execution/misc/LocalJVMSuite.scala
mdedetrich: Sure thing, do you have a recommendation of where to place such a test? Also should I keep this package private test as well (I personally think small isolated unit test is a good idea just so people don't accidentally break it in the future)

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
mdedetrich: Its the only method I could find, is there another one I should be using?

monix-eval/shared/src/main/scala/monix/eval/Task.scala
mdedetrich: Sure I can change it to `localContextIsolateOnRun`, it also sounds better.

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
Avasil: I'm wondering if it can be old implementation of this method - did you encounter any problems?

I think it will be already current context (`Local.getContext()`) and the runloop should take care of preserving it across async boundaries

monix-eval/shared/src/main/scala/monix/eval/Task.scala
mdedetrich: So I tried replicating @oleg-py PR as much as possible however I am still getting

```
[error] monix-eval: Failed binary compatibility check against io.monix:monix-eval_2.12:3.0.0! Found 1 potential problems
[error]  * the type hierarchy of object monix.eval.Task#Options is different in current version. Missing types {scala.runtime.AbstractFunction2}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("monix.eval.Task$Options$")
[error] stack trace is suppressed; run last evalJVM / mimaReportBinaryIssues for the full output
[error] (evalJVM / mimaReportBinaryIssues) Failed binary compatibility check against io.monix:monix-eval_2.12:3.0.0! Found 1 potential problems
[error] Total time: 2 s, completed Mar 28, 2020 12:23:41 PM
```

monix-execution/jvm/src/test/scala/monix/execution/misc/LocalJVMSuite.scala
oleg-py: LocalJVMSuite is good enough, since it's testing both JVM-specific stuff (threading) and Local interop.

mdedetrich: Thanks, will do this tonight

mdedetrich:
I have just updated the PR, I did the binary compatible changes as well as changing the option to `localContextIsolateOnRun`.

I still need to do the additional test which I will do it tonight.

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
mdedetrich: So I am just doing `$f` now and the unit test seems to pass so I am using this now.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
Avasil: Yes, I feel like we only need `def apply` there, at least according to mima and our tests :D

monix-execution/jvm/src/test/scala/monix/execution/misc/LocalJVMSuite.scala
mdedetrich: I ended up adding a test to `TaskLocalJVMSuite` since you don't have access to `Task` in `LocalJVMSuite`. Let me know if you need anything else.

mdedetrich:
So I added the requested test, the PR should be done.

project/MimaFilters.scala
```diff
@@ -5,7 +5,7 @@ object MimaFilters {
   lazy val changesFor_3_2_0: Seq[ProblemFilter] = Seq(
     // Signature change in internal instance
     exclude[IncompatibleResultTypeProblem]("monix.catnap.internal.ParallelApplicative.apply"),
-    exclude[MissingClassProblem]("monix.eval.internal.TaskGather*")
```
Avasil: Something went wrong during the merge, this line shouldn't be removed

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
```diff
@@ -148,14 +150,17 @@ object Local extends LocalCompanionDeprecated {
     def isolate(f: Tree): Tree =
       c.abort(c.macroApplication.pos, "Macro no longer implemented!")
 
-    def localLetCurrentIf(b: Tree)(f: Tree): Tree = {
+    def localLetCurrentIf(b: Tree, isolate: Tree)(f: Tree): Tree = {
```
Avasil: now it can be just `def localLetCurrentIf(b: Tree)`  because the behavior is the same for `b && isolate`
If we do it, we can also remove mima exclusion

project/MimaFilters.scala
mdedetrich: Yeah I just noticed the merge screwed up, just fixed it

monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala
mdedetrich: Maybe I am missing something about how macros work, but aren't the signatures meant to match, i.e.

```scala
private[monix] def bindCurrentIf[R](b: Boolean, isolate: Boolean)(f: => R): R
```

matches with

```scala
 def localLetCurrentIf(b: Tree, isolate: Tree)(f: Tree): Tree
```

I mean I need to check the `isolate` condition somehow?

mdedetrich: Or do you mean doing the isolate check inside of
```private[monix] def bindCurrentIf[R](b: Boolean, isolate: Boolean)(f: => R): R =```?

Comment (Avasil) [reply monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala]: I mean we could keep the old signature, that is:

`private[monix] def bindCurrentIf[R](b: Boolean)(f: => R): R`

And call it with `&&` in `Task`

Comment (mdedetrich) [reply monix-execution/shared/src/main/scala/monix/execution/misc/Local.scala]: DfKL_yPX4AAAaPG

Fixed

**Comment by mdedetrich** [comment]
@Avasil Okay, everything is resolved now and the tests are passing

**Comment by Avasil** [comment]
@mdedetrich 
Not sure if it helps you with anything but I released a snapshot: `3.2.0-4524fe2-SNAPSHOT`

Nothing else is blocking `3.2.0` release, I will release after we merge this PR but I want to try it to do sanity checks with HTTP libraries. If everything is good then we could release even on Monday

**Comment by mdedetrich** [comment]
Many thanks for pushing this through @Avasil , I just did one additional change where I in the test I also check to see if `enableLocalContextPropagation` is working correctly.

I will try the snapshot in my projects to see if the tests pass now!

**Comment by mdedetrich** [comment]
Actually I am having issues downloading the snapshot, I am getting 

```
[error] (core / update) sbt.librarymanagement.ResolveException: Error downloading io.monix:monix_2.12:3.2.0-4524fe2-SNAPSHOT
[error]   Not found
[error]   Not found
[error]   not found: /home/mdedetrich/.ivy2/local/io.monix/monix_2.12/3.2.0-4524fe2-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/io/monix/monix_2.12/3.2.0-4524fe2-SNAPSHOT/monix_2.12-3.2.0-4524fe2-SNAPSHOT.pom
[error] Error downloading io.monix:monix-execution_2.12:3.2.0-4524fe2-SNAPSHOT
[error]   Not found
[error]   Not found
[error]   not found: /home/mdedetrich/.ivy2/local/io.monix/monix-execution_2.12/3.2.0-4524fe2-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/io/monix/monix-execution_2.12/3.2.0-4524fe2-SNAPSHOT/monix-execution_2.12-3.2.0-4524fe2-SNAPSHOT.pom
[error] (task / ssExtractDependencies) sbt.librarymanagement.ResolveException: Error downloading io.monix:monix_2.12:3.2.0-4524fe2-SNAPSHOT
[error]   Not found
[error]   Not found
[error]   not found: /home/mdedetrich/.ivy2/local/io.monix/monix_2.12/3.2.0-4524fe2-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/io/monix/monix_2.12/3.2.0-4524fe2-SNAPSHOT/monix_2.12-3.2.0-4524fe2-SNAPSHOT.pom
[error] Error downloading io.monix:monix-execution_2.12:3.2.0-4524fe2-SNAPSHOT
[error]   Not found
[error]   Not found
[error]   not found: /home/mdedetrich/.ivy2/local/io.monix/monix-execution_2.12/3.2.0-4524fe2-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/io/monix/monix-execution_2.12/3.2.0-4524fe2-SNAPSHOT/monix-execution_2.12-3.2.0-4524fe2-SNAPSHOT.pom
[error] (core / ssExtractDependencies) sbt.librarymanagement.ResolveException: Error downloading io.monix:monix_2.12:3.2.0-4524fe2-SNAPSHOT
[error]   Not found
[error]   Not found
[error]   not found: /home/mdedetrich/.ivy2/local/io.monix/monix_2.12/3.2.0-4524fe2-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/io/monix/monix_2.12/3.2.0-4524fe2-SNAPSHOT/monix_2.12-3.2.0-4524fe2-SNAPSHOT.pom
[error] Error downloading io.monix:monix-execution_2.12:3.2.0-4524fe2-SNAPSHOT
[error]   Not found
[error]   Not found
[error]   not found: /home/mdedetrich/.ivy2/local/io.monix/monix-execution_2.12/3.2.0-4524fe2-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/io/monix/monix-execution_2.12/3.2.0-4524fe2-SNAPSHOT/monix-execution_2.12-3.2.0-4524fe2-SNAPSHOT.pom
[error] Total time: 1 s, completed Mar 29, 2020 11:32:00 AM)
```

**Comment by mdedetrich** [comment]
@Avasil I just realized that I have some unintended behavior now (although I think you were hinting at it earlier). Basically if you mix `enableLocalContextPropagation` with `disableLocalContextIsolateOnRun` then setting/getting values concurrently stops working. I have updated both of my projects (https://github.com/mdedetrich/monix-mdc/ and https://github.com/mdedetrich/monix-opentracing) and tests which involve concurrently setting/getting values have stopped working.

Is there a conflict theory wise between having `enableLocalContextPropagation` and `disableLocalContextIsolateOnRun`?

Also Is it possible to hold off on the merge until I figure this out.

File: monix-eval/shared/src/main/scala/monix/eval/Task.scala [hunk 9]
```diff
@@ -995,7 +996,7 @@ sealed abstract class Task[+A] extends Serializable with TaskDeprecated.BinCompa
   @UnsafeBecauseImpure
   final def runSyncStepOpt(implicit s: Scheduler, opts: Options): Either[Task[A], A] = {
     val opts2 = opts.withSchedulerFeatures
-    Local.bindCurrentIf(opts2.localContextPropagation) {
+    Local.bindCurrentIf(opts2.localContextPropagation && opts2.localContextIsolateOnRun) {
```
mdedetrich: @Avasil So I think that this is causing the issue where concurrent read/writes of `Local` stopped working if you have `disableLocalContextIsolateOnRun`

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -808,7 +809,7 @@ sealed abstract class Task[+A] extends Serializable with TaskDeprecated.BinCompa
   @UnsafeBecauseImpure
   def runAsyncOptF(cb: Either[Throwable, A] => Unit)(implicit s: Scheduler, opts: Options): CancelToken[Task] = {
     val opts2 = opts.withSchedulerFeatures
-    Local.bindCurrentIf(opts2.localContextPropagation) {
+    Local.bindCurrentIf(opts2.localContextPropagation && opts2.localContextIsolateOnRun) {
```
mdedetrich: @Avasil Do you think that this is causing the issue where concurrent read/writes of `Local` stopped working if you have `disableLocalContextIsolateOnRun`?

Its the only thing that comes to mind, the rest of the functions just deal with running it to a `Future`?

Avasil:
@mdedetrich 

I feel like these tests:
```scala
  "Using Future only" can {
    "Write and get different values concurrently" in {
      val keyValues = MultipleKeysMultipleValues.multipleKeyValueGenerator.sample.get

      val futures = keyValues.keysAndValues.map { keyValue =>
        getAndPut(keyValue.key, keyValue.value)
      }

      val future = Future.sequence(futures)

      future.map { retrievedKeyValues =>
        retrievedKeyValues.size shouldBe keyValues.keysAndValues.size
        retrievedKeyValues shouldBe keyValues.keysAndValues.map(_.value)
      }
    }
  }
```

and

```
    "Write and get different values concurrently" in {
      val keyValues = MultipleKeysMultipleValues.multipleKeyValueGenerator.sample.get

      val tasks = keyValues.keysAndValues.map { keyValue =>
        getAndPut(keyValue.key, keyValue.value).executeAsync
      }

      val task = Task.parSequence(tasks)

      task.runToFutureOpt.map { retrievedKeyValues =>
        retrievedKeyValues.size shouldBe keyValues.keysAndValues.size
        retrievedKeyValues shouldBe keyValues.keysAndValues.map(_.value)
      }
    }
```
won't work regardless of the change because `Local` is shared if it's not isolated. 
I feel like we should change it for `Task.parSequence` and other parallel methods (basically @oleg-py PR: #1044 ) but it won't automatically work for `Future.sequence`. We would have to call `TaskLocal.isolate` and `Local.isolate` on each `Task` and `Future` respectively.

Currently we auto-isolate on running `Task` and each `Task` that was separately run will work on a different copy of `Local`:

```scala
val local = Local(0)
Task(local := 10).runToFuture // will work with 10
Task(local := 20).runToFuture // will work with 20

println(local.get) // 0, each Task had different copy 
```

This is pretty good, libraries will run the `Task` with `runAsync` or something so each request will get different copy.

The problem with what we have on master is if something maps the result (like `Akka HTTP` directives) of running the `Task` to access the `Local`:

```scala
val local = Local(0)

Task(local := 10).runToFuture.map(_ => println(local.get))
```

We might expect `10` but it is `0` because `local := 10` was isolated and after `runToFuture` it went back to working on previous `Local`. I feel like what we really want is to allow `Future` returned in `runToFuture` to keep the reference to isolated `Local` to achieve:

```scala
val local = Local(0)

Task(local := 10).runToFuture.map(_ => println(local.get)) // 10
Task(local := 20).runToFuture.map(_ => println(local.get)) // 20

println(local.get) // 0
```

In this PR: 

```scala
val local = Local(0)

Task(local := 10).disableLocalContextIsolateOnRun.runToFuture.map(_ => println(local.get)) // 10 or 20
Task(local := 20).disableLocalContextIsolateOnRun.runToFuture.map(_ => println(local.get)) // 10 or 20

println(local.get) // 10 or 20
```

mdedetrich:
Indeed you are right, I just verified that the `disableLocalContextIsolateOnRun` is a red herring, furthermore and more importantly I actually realized what was causing my tests in monix-mdc to be flaky.

SBT by default will run tests in parallel and sync MDC is using a global context the tests was clearing the context when another test was running at the same time. Doing 

```scala
parallelExecution in Test in ThisBuild := false
```

Solved the issue.

Would it be difficult to get Oleg's PR https://github.com/monix/monix/pull/1044 in before the 3.2.0 release. I guess executing Future's asynchronously with automatic isolation might be too hard for now but I would prefer for it to be automatic with `Task` since its definitely possible there.

Avasil:
> Would it be difficult to get Oleg's PR #1044 in before the 3.2.0 release? I guess executing Future's asynchronously with automatic isolation might be too hard for now but I would prefer for it to be automatic with Task since its definitely possible there.

I don't think it's hard, I could add other parallel operators there if Oleg is busy. I think it's mostly a matter of semantics:

What should happen with locals if we run n tasks concurrently with `parMap`, `parSequence` etc.?
- all of them share Local unless isolated (current behavior)
- all of them are automatically isolated and returned task keeps previous `Local`
- all of them are automatically isolated and returned task gets `Local` from one of the tasks

> Should be checking for localContextIsolateOnRun? They are not really running the Task into a Future. Otherwise if you think the PR is fine as is then it can be merged after you test it with the http clients.

I feel like we have to ask ourselves a question if we need an option to remove auto-isolation or not. And if we do want it, then should it be separate from parallel operators or not?

mdedetrich:
> What should happen with locals if we run n tasks concurrently with parMap, parSequence etc.?

I can't comment that much in regards to laws because I am not that familiar with them (I am also not that familiar with Monix's internals). All I can say is that the behavior depends on the circumstance, however the second one that you mentioned, i.e.

For monix-mdc, using `isolate` by default tends to make sense because its rare that you want to share the same HashMap between different runnings tasks. Often you put some value by some key (i.e. `userId`) and you want this to be isolated because if this is happening concurrently its usually due to multiple requests (and you don't want conflicts between those requests)

However for monix-opentracing, you generally have one SpanContext that is backed by a local that is shared by everyone and your application logic just thows span's onto the span context.

I like the point that @oleg-py made here https://github.com/monix/monix/pull/1044#issuecomment-544953640, i.e. 

> Putting it on options since it'll affect the execution of sub-tasks like the other two options we have. In 4.0 we can probably drop the booleans and have a enum of how locals work with { Disabled, Shared, Isolated } as options. I'm open to other ideas, but I can't think of one.

Is a saner idea, because how you want the local's to behave by default is often defined by the datastructure/context that is backing the local and not the computations that you are doing on `Task`, i.e. if your `Local` is storing some sort of hashmap based on a key (mdc case) then isolating the local by default is almost always the right default because then you have collisions on your hashmap, however in the `OpenTracing` case where your Local is basically storing a queue of span's, sharing makes most sense because there aren't any collisions, you are just appending spans onto some collection and ordering doesn't really matter here.

> I feel like we have to ask ourselves a question if we need an option to remove auto-isolation or not. And if we do want it, then should it be separate from parallel operators or not?

I don't see how this is related. The idea I had in mind for the `disableLocalContextIsolateOnRun` is not anything explicit about isolation but rather interopting between `Future` and `Task`. The way I envisaged it is that if you are running a `Task, i.e. going from `Task` => `Future` then this is where this option should be used otherwise it should be ignored.

This is the reason why I am contemplating removing the check from those mentioned methods because they don't actually convert a `Task` to a `Future` and hence it shouldn't even be reading those options (maybe I should rename the option to make it more clear?)

mdedetrich:
Also I have changed my mind regarding putting @oleg-py changes before 3.2.0 release as I realized that whatever case you want depends on the situation at hand and luckily in my case the default is fine.

There is still the open question of checking the `localContextIsolateOnRun` on the two methods in `Task` which don't go from `Task => Future`.

mdedetrich:
@Avasil Do you have an update on this?

Avasil:
@mdedetrich 

What's the difference between using "global" `AtomicAny` and `Local` for `monix-opentracing`?

mdedetrich:
Well you can have different `SpanContext`'s available at the same runtime in an app (albeit its not typical).

Avasil:
BTW what was your initial issue with auto-isolation?

I suppose we could have something like that in Akka HTTP now:

```scala
// based on onComplete from Akka
def onCompleteLocal[T](task: ⇒ Task[T]]): Directive1[Try[T]] = {
  import akka.http.scaladsl.util.FastFuture._
  Directive { inner ⇒
    ctx ⇒
      Local.isolate {
        local := id // extract from header
        task.executeWithOptions(_.disableLocalContextIsolateOnRun).runToFuture.fast.transformWith(t ⇒ inner(Tuple1(t))(ctx))
      }
  }
}
```

but I'm not sure how much difference does it make

mdedetrich:
Something like this, lets assume we are using monix-mdc

```scala
  AddResponseHeadersDirective {  ⇒
    onCompleteLocal(...)
  }
```

Where `AddResponseHeadersDirective` would also do some logging after the future is complete, lets say transform the response by adding some headers, so it might be defined something like

```scala
val AddResponseHeadersDirective(inner: _ => Future[HttpResponse]): Future[HttpResponse] =
   inner.map{ response =>
     logger.info("Adding middleware header")
     response.addHeader(BasicHeader("a" -> "b"))
   }
}
```

The logging in `task` would invariable add some logs to the mapped diagnostic context (lets say add a `user_id`) and then `AddResponseHeadersDirective` would also do some logging, it would need to access the `Local` that the `task` modified so that the `logger.info("Middleware added header")` would see the `user_id` added in `task`.

Now it may be possible to do this if you have control over the definition of `AddResponseHeadersDirective` because you could just theoretically work with `Task[HttpResponse]` instead of `Future[HttpResponse]` however if you don't such control (i.e. some other library defines `AddResponseHeadersDirective` like some middleware that works over `Future` which is typical) then your stuck. Akka-http can also do some of its own logging well after the `onCompleteLocal` has completed and none of that logging will be able to access the `Local` from the task unless you use the `disableLocalContextIsolateOnRun`

So actually your example of

```
task.executeWithOptions(_.disableLocalContextIsolateOnRun).runToFuture.fast.transformWith(t ⇒ inner(Tuple1(t))(ctx))
```

The point is not about `task` having an isolated local in your custom directive (this I can do right now), its actually making sure that the `Local` being created by your task inside your custom directive can be accessed by akka-http and any other libraries which you have no control over.

Avasil:
@mdedetrich I'm wondering where `Local.isolate` should be called in this scenario, let's say we have:

```scala
import akka.http.scaladsl.model.StatusCodes
import akka.http.scaladsl.server._
import monix.eval.Task

import scala.util.{Random, Try}

object AkkaHttpTest extends HttpApp with App {
  implicit val s: Scheduler = Scheduler.traced
  val local = Local(0)

  val addResponseHeader: Directive0 =
    mapInnerRoute(_.andThen(_.map { result =>
      println(s"Add Header: ${local.get}")
      result
    }))

  def onCompleteLocal[T](task: ⇒ Task[T], id: Int): Directive1[Try[T]] = {
    import akka.http.scaladsl.util.FastFuture._
    Directive { inner ⇒
      ctx ⇒
        Local.isolate {
          local := id
          task.executeWithOptions(_.disableLocalContextIsolateOnRun).runToFuture.fast.transformWith(t ⇒ inner(Tuple1(t))(ctx))
        }
    }
  }

  override def routes: Route =
    path("hello" / IntNumber) { id =>
      get {
        addResponseHeader {
          onCompleteLocal(Task {
            Thread.sleep(Random.nextInt(1000))
            println(s"Processing request: ${local.get} ")
          }, id) { _ => complete(StatusCodes.OK) }
        }
      }
    }

  AkkaHttpTest.startServer("localhost", 8080)
}
```

and send requests concurrently with numbers 1-100
 
Part of the output:
```
Processing request: 72 
Add Header: 0
Processing request: 50 
Add Header: 0
Processing request: 65 
Add Header: 0
Processing request: 99 
Add Header: 0
Processing request: 51 
Add Header: 0
Processing request: 82 
Add Header: 0
```

If we remove `Local.isolate` then we would have:

```
Processing request: 88 
Add Header: 62
Processing request: 98 
Add Header: 62
Processing request: 62 
Add Header: 62
Processing request: 60 
Add Header: 62
Processing request: 84 
Add Header: 62
```

So it seems like neither way is correct. Although I'm not proficient with Akka HTTP and I could be misinterpreting something

mdedetrich:
Let me look at this today/tomorrow, will come back to you

monix-eval/shared/src/main/scala/monix/eval/Task.scala
```diff
@@ -4276,6 +4289,18 @@ object Task extends TaskInstancesLevel1 {
     def disableLocalContextPropagation: Options =
       copy(localContextPropagation = false)
 
+    /** Creates a new set of options from the source, but with
+      * the [[localContextIsolateOnRun]] value set to `true`.
+      */
+    def enableLocalContextIsolateOnRun: Options =
+      new Options(autoCancelableRunLoops, localContextPropagation, localContextIsolateOnRun = true)
```
alexandru, line 4296: The other methods are using `copy`. Any reason for going against it?

Just a superficial observation, no biggie.

mdedetrich:
@Avasil This is the problem I am trying to solve, try this

```scala
import akka.http.scaladsl.model.StatusCodes
import akka.http.scaladsl.server._
import monix.eval.Task

import scala.util.{Random, Try}

object AkkaHttpTest extends HttpApp with App {
  implicit val s: Scheduler = Scheduler.traced
  implicit val opts: Task.Options = Task.defaultOptions.enableLocalContextPropagation.disableLocalContextIsolateOnRun
  val local = Local(0)

  val addResponseHeader: Directive0 =
    mapInnerRoute(_.andThen(_.map { result =>
      println(s"Add Header: ${local.get}")
      result
    }))

  def onCompleteLocal[T](task: ⇒ Task[T]): Directive1[Try[T]] = {
    import akka.http.scaladsl.util.FastFuture._
    Directive { inner =>
      ctx =>
        task.runToFutureOpt.fast.transformWith(t ⇒ inner(Tuple1(t))(ctx))
    }
  }

  override def routes: Route =
    path("hello" / IntNumber) { id =>
      get {
        addResponseHeader {
          onCompleteLocal(Task {
            local := id
          }) { _ => complete(StatusCodes.OK) }
        }
      }
    }

  AkkaHttpTest.startServer("localhost", 8080)
}
```

If you have `Task.defaultOptions.enableLocalContextPropagation.disableLocalContextIsolateOnRun` and run `curl localhost:8080/hello/3` it prints out 3 (which is what you expect) however if you don't have `disableLocalContextIsolateOnRun`, i.e. `Task.defaultOptions.enableLocalContextPropagation` then it just prints out 0.

monix-eval/shared/src/main/scala/monix/eval/Task.scala
mdedetrich: Here the `copy` method only allows you to set `autoCancelableRunLoops` and `localContextPropagation` (due to bincompat we can only have copy for the original 2 parameters).

Avasil:
@mdedetrich The issue with the snippet is that if there are concurrent requests then the `Local` value could be mixed up

> My intention is to simply allow Local to propagate from Task to Future

Yes but we would like to have isolation too, right? Ideally we could propagate isolated `Local` to `Future` instead of restoring the original one

As I described in the second part of my comment:
https://github.com/monix/monix/pull/1146#issuecomment-605634230

mdedetrich:
> @mdedetrich The issue with the snippet is that if there are concurrent requests then the Local value could be mixed up

Ah I see what you mean but this is because you are using a global `Local right? I believe this is up to the implementation of how `Local` is being used by your library.

For example with monix-mdc, the `Local` is actually storing a `Map` (see https://github.com/mdedetrich/monix-mdc/blob/master/src/main/scala/org/mdedetrich/monix/mdc/MonixMDCAdapter.scala#L11) and theoretically if it used a `ConcurrentHashMap` rather than a standard one it would even work in your case (I actually have an unpublished open PR on this). Let me clarify this in a few hours.

mdedetrich:
@Avasil 

Okay so now I understand the issue, its basically when you mix concurrency on the boundaries of `runToFuture`/`runToFutureOpt` (which is actually very common if you run a http webserver).

I made a sample repo to better help test this stuff, you can check it out here https://github.com/mdedetrich/akka-monix-test

I am not exactly sure what to do here though, it seems like we kinda want a redesign of how `Local` works in other for this use case to work? (either that or we have to say its impossible to support).

Avasil:
@mdedetrich I'm experimenting with naive solution like:

```scala
  def runToFutureOpt(implicit s: Scheduler, opts: Options): CancelableFuture[A] = {
    val opts2 = opts.withSchedulerFeatures
    if (opts2.localContextPropagation) {
      val prev = Local.getContext()
      Local.setContext(Local.getContext().isolate())
      try TaskRunLoop.startFuture(this.executeAsync, s, opts2)
      finally Local.setContext(prev)
    }
    else {
      TaskRunLoop.startFuture(this, s, opts2)
    }
  }
```
or

```scala
  def runToFutureOpt(implicit s: Scheduler, opts: Options): CancelableFuture[A] = {
    val opts2 = opts.withSchedulerFeatures
    if (opts2.localContextPropagation) {
      CancelableFuture.unit.flatMap { _ =>
        Local.setContext(Local.getContext().isolate())
        TaskRunLoop.startFuture(this.executeAsync, s, opts2)
      }
    }
    else {
      TaskRunLoop.startFuture(this, s, opts2)
    }
  }
```

And it often works but there is a race condition, sometimes the following test fails:

```scala
  testAsync("Task.runToFuture isolates but preserves context for Future continuation") {
    implicit val s: Scheduler = Scheduler.Implicits.traced
    val local = Local(0)

    def test(i: Int): Future[Local.Context] = {
      for {
        _ <- {
          val prev = local.get
          Task {
            local.update(prev + i)
          }
      }.runToFuture
        next <- Future {
          local.get // sometimes gets restored local instead of isolated one
        }
      } yield {
        assertEquals(next, i)
        Local.getContext()
      }
    }

    val futures = List.range(0, 6).map(test)

    Future.sequence(futures).map { listOfLocals =>
      assert(listOfLocals.toSet.size == listOfLocals.size, "All locals need to be unique")
    }
  }
```

The idea is we run `Task + Future` asynchronously with isolated Local but restore it on a current thread

I will be looking into it further but if I don't figure it out today then I will probably release  `3.2.0` anyway. It has important bug fixes and a lot of people are waiting for ScalaJS 1.0.0. We can always release fix for local in `3.2.1` or `3.3.0` as soon as it is available

mdedetrich:
@Avasil Sounds good for me, I have plenty of other work to do in the meantime. Should I close this PR or leave it open?

Avasil:
I guess we can close, seems like the solution will look differently

mdedetrich:
Okay closing, can also re-open later if we need it for some reason.

jvican:
@Avasil @mdedetrich I'm also interested in having a capability such as this one, I believe I might need it for several libraries I'm building. Do you know if there has been some progress on this? The  test you shared Piotr looks exactly like the semantics we want.

Avasil:
@jvican I've temporarily switched to other things but I'm hoping to resume working on it this week

We definitely want to have this case working

jvican:
Thanks! I’m looking forward to it 😄

Avasil:
@jvican @mdedetrich 

I just wanted to let you know that I didn't forget about it. 😅 

I think I got pretty close the last time I was working on it, for a short moment I even thought I had it. :D
I want to release `monix-bio` 1.0.0 in the next week so I'm focusing on it right now. 
My hope is that once it's done, I will be able to multitask a lot less and focus on `Local` to have the feature in July

mdedetrich:
@Avasil Note to throw a spanner in your plans but wouldn't it make sense to figure this behaviour out in monix before releasing monix-bio on the assumption that you want to have the same behaviour in monix-bio which may have binary compatibility issues?

Avasil:
Well, I aim to do it in a binary compatible way; otherwise, we have a bigger problem with `monix/monix` 😅

Issue #1151, opened 2020-03-29
New consumer builder `.seq` for `Consumer` companion object


paualarco:
Would be nice to have the `.seq` signature as a builder for `Consumer`.
I guess there are some reasons why this is has not being implemented yet but  in my opinion would be worth having it. (I find very handy the equivalent in akka streams `Sink.seq`)

Avasil:
> I guess there are some reasons why this is has not being implemented yet but in my opinion would be worth having it.

I don't think there is a specific reason other than no one bringing it up before. 
The use case is covered (`toListL`) but I wouldn't mind seeing it on `Consumer`

paualarco:
@Avasil how does it seem to you? I have reused the already existing `FoldLeftConsumer`

PR #1152, opened 2020-03-30
Added collect signature for Consumer companion object (#1151)

paualarco:


monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
```diff
@@ -227,6 +227,17 @@ object Consumer {
...
+  /** A consumer that will return a collection of all the streamed
+   * values as a Scala `List`.
+   *
+   * Note that this method is functionally the same as the `.toListL` from [[Observable]]
+   *
+   * WARNING: For infinite streams the process will eventually
+   * blow up with an out of memory error.
+   */
+  def collect[A]: Consumer.Sync[A, List[A]] =
+    new FoldLeftConsumer[A, List[A]](() => List.empty[A], (l: List[A], elem: A) => l :+ elem)
```
Avasil: I feel like we could follow `toListL` implementation:
```
foldLeftL(mutable.ListBuffer.empty[A])(_ += _).map(_.toList)
```

I expect `map(_.toList)` to be much cheaper than appending to the end of list each time

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
```diff
@@ -227,6 +227,17 @@ object Consumer {
   def raiseError[In, R](ex: Throwable): Consumer.Sync[In, R] =
     new RaiseErrorConsumer(ex)
 
+  /** A consumer that will return a collection of all the streamed
+   * values as a Scala `List`.
+   *
+   * Note that this method is functionally the same as the `.toListL` from [[Observable]]
```
Avasil: A link would be nice, i.e. `[[Observable.toListL]]`

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
```diff
@@ -227,6 +227,17 @@ object Consumer {
...
+   *
+   * WARNING: For infinite streams the process will eventually
+   * blow up with an out of memory error.
+   */
+  def collect[A]: Consumer.Sync[A, List[A]] =
```
Avasil: `collect` usually uses partial function - what do you think about a name like `toList` (consistent with `toListL`) ?

paualarco: sure, btw, what does represent the final `L` from the `.toListL` ?

paualarco: amended all your comments, thank you for the quick feedback! :)

monix-reactive/shared/src/main/scala/monix/reactive/Consumer.scala
```diff
@@ -227,6 +229,17 @@ object Consumer {
...
 
+  /** A consumer that will return a collection of all the streamed
+   * values as a Scala `List`.
+   *
+   * Note that this method is functionally the same as the [[Observable.toListL]]
+   *
+   * WARNING: For infinite streams the process will eventually
+   * blow up with an out of memory error.
+   */
+  def toListL[A]: Consumer[A, List[A]] =
```
Avasil, line 240: `L` is only used in `Observable` to differentiate between methods returning `Task` and `Observable`, I would leave just `toList` in `Consumer`. Sorry for nitpick :D

paualarco: No worries, I have changed that already :)

monix-reactive/shared/src/test/scala/monix/reactive/consumers/ListConsumerSuite.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import scala.util.{Failure, Success}
+
+object ListConsumerSuite extends TestSuite[TestScheduler] {
+  def setup(): TestScheduler = TestScheduler()
+  def tearDown(s: TestScheduler): Unit = {
+    assert(s.state.tasks.isEmpty, "TestScheduler should have no pending tasks")
+  }
+
+  test("should return the same all consumed elements as a sequence") { implicit s =>
```
Avasil, line 34: I think it needs `s.tick()` for tests to pass and it's good to merge :)

paualarco: thought it was passing already... 
is there a better way of running minitests than "sbt project reactiveTests" andThen "sbt test"?

Avasil: `ReactiveTests` is something else but I can see how it is confusing :D 

```scala
reactiveJVM/test
```

or

```scala
reactiveJVM/testOnly *NameOfTheTestClass
```

PR #1157, opened 2020-04-05
Scalajs 1.0.0 support

Avasil:


project/plugins.sbt
```diff
@@ -1,4 +1,6 @@
-addSbtPlugin("org.scala-js"         % "sbt-scalajs"     % "0.6.32")
+val scalaJSVersion =
+  Option(System.getenv("SCALAJS_VERSION")).filter(_.nonEmpty).getOrElse("1.0.0")
```
alexandru, line 2: Good to see that you've made `1.0.0` the default version, as it should be.

Avasil:
Note there is one issue I will leave unsolved:

ScalaJS is more strict about linking errors since 1.0.0 and it fails for `java.lang.Thread.UncaughtExceptionHandler` which is used in `UncaughtExceptionReporter.asJava`.

The method is only used in JVM internals and arguably shouldn't be used in ScalaJS code.
It was only used in test so I moved it to JVM only.

I will make a PR to ScalaJS which will add the missing interface so the problem will disappear in the future.

To sum up:
- If anyone is using `java.lang.Thread.UncaughtExceptionHandler` in ScalaJS code and tries to upgrade from `0.6.32` to `1.0.0` then it will fail at compile time but it would fail without Monix as well

PR #1172, opened 2020-04-27
Fix #1170: Activate GitHub Actions (2)

### LINKED ISSUE #1170: Switch to GitHub Actions
Travis is sometimes too slow and I think I have enough experience to attempt a switch to GitHub Actions.

The problem is we need to test with multiple Java / Scala versions, so we need to setup a test matrix. We have to look for inspiration at what other projects have done.

Also it would be awesome if we configured automated releases.

alexandru:


alexandru:
Please mute this PR for now because it's going to be noisy. 

Configuring GitHub Actions is hard for a project like Monix and I need to do experiments.

alexandru:
This is the wrong approach. Closing the PR, as it might be misleading.

paualarco:
@alexandru what does wrong means here? we used it as a a base in `connect` and `bio`.
There also is a PR pointing to `github-actions-2` that brings automated releases too.

alexandru:
It's not wrong per se, but too simplistic, we need a complicated build matrix for this to work.

#1189 seems more evolved, closed this one as to not mislead as if it's somehow in the works.

PR #1176, opened 2020-05-02
WIP: Added RetryWithBackoff Operator

Executioner1939:
@Avasil this is my attempt at the operator you mentioned.

I used the tests from the `OnErrorRetryCountedObservableSuite`, I don't quite understand how they work exactly, so some pointers would be nice :-).

I could add in some more Backoff Strategies, like Linear and Fibonacci as well, which could be useful to have.

PS: It seems Scalafmt changed some things I didn't intend for it to change.


Avasil:
> I used the tests from the OnErrorRetryCountedObservableSuite, I don't quite understand how they work exactly, so some pointers would be nice :-).

I'm also often confused by `Sample` but it helps us get quite a few of tests for free.

```scala
case class Sample(
  // Tested Observable
  observable: Observable[Long],
  // Number of total elements it should emit
  count: Int,
  // Sum of all elements it should emit
  sum: Long,
 // How much time should we wait for the first element
  waitFirst: FiniteDuration,
 // How much time should we wait for each element after the first 
  waitNext: FiniteDuration
)
```

```scala
def createObservable(sourceCount: Int) = Some {
  val ex = DummyException("expected")
  val o = Observable
    .range(0, sourceCount)
    .endWithError(ex)
    .onErrorRestart(3)
    .onErrorHandle { case _ => 10L }

  val count = sourceCount * 4 + 1
  val sum = 1L * sourceCount * (sourceCount - 1) / 2 * 4 + 10
  Sample(o, count, sum, Duration.Zero, Duration.Zero)
}
```
`o` will emit an error after `sourceCount` elements. After each error, it will restart 3 times (so 4 runs in total) and when it will give up and emit an error downstream then it will be handled with `onErrorHandle { case _ => 10L }` (hence + 1 in count and + 10 in sum). 

There isn't any delay between elements so `waitFirst` and `waitNext` is `Duration.Zero`. 

I think we will need few extra tests because `BaseOperatorSuite` doesn't really handle exponential delays between elements. In case of retrying `Observable` it would be nice to have resetting backoff, i.e.:

```
next elem 
next elem
fail // waits 1 s
fail // waits 2 s
fail // waits 4 s
next elem
next elem
fail // waits 1 s (instead of 8 s)
```

If the application is long running then the delay between retries could become ridiculously big without it if we sometimes lose source connection. An example test for it:

```scala
  test("Observable.onErrorRestartWithBackoff resets the delay after success") { implicit s =>
    val dummy = DummyException("dummy")

    val source = Observable
      .range(0, 100)
      .flatMap(n => if (n == 2) Observable.raiseError(dummy) else Observable.now(n))
      .onErrorRetryWithBackoff(10, 1.second)
      .takeByTimespan(5.second)
      .toListL

    val f = source.runToFuture

    s.tick(5.second)
    assertEquals(f.value, Some(Success(List(0, 1, 0, 1, 0, 1, 0, 1, 0, 1))))
  }
```

I suppose it's not quite consistent with `onErrorRestart` so maybe we could hide it behind a flag or something

> I could add in some more Backoff Strategies, like Linear and Fibonacci as well, which could be useful to have.

It would be nice to allow users to configure it, maybe let them pass a function with few predefined ones?

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -2703,6 +2701,26 @@ abstract class Observable[+A] extends Serializable { self =>
...
+    * parameter, while also implementing an exponential backoff strategy meaning
+    * each successive retry will wait twice as long as the previous before attempting
+    * to subscribe to the source again.
+    *
+    * So for an Observable that always ends in error the total number of subscriptions that will eventually
+    * happen is `maxRetries + 1`, and the total time taken before the source fails will be
+    * `(initialDelay ^ maxAttempts - 1) / 2`, so if the `initialDelay` is 2 seconds with a `maxAttempts`
+    * value of 10, the the total time taken, assuming all retries fail, will be 512 seconds.
+    */
+  final def onErrorRetryWithBackoff(maxRetries: Long, initialDelay: FiniteDuration): Observable[A] = {
```
Avasil, line 2719: I suggest `onErrorRestartWithBackoff` for consistency with other `onErrorRestartXXX` operators

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/OnErrorRetryWithBackoffObservable.scala
```diff
@@ -0,0 +1,93 @@
...
+import monix.execution.cancelables.OrderedCancelable
+import monix.execution.{Ack, Cancelable, Scheduler}
+import monix.reactive.Observable
+import monix.reactive.observers.Subscriber
+
+import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
+
+private[reactive] final class OnErrorRetryWithBackoffObservable[+A](
+                                                                     source: Observable[A],
```
Avasil, line 31: The formatting seems off, it should be on the left - how did you run `scalafmt`?

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/OnErrorRetryWithBackoffObservable.scala
```diff
@@ -0,0 +1,93 @@
...
+          isDone = true
+          subscriber.onComplete()
+        }
+
+      def onError(ex: Throwable): Unit =
+        if (!isDone) {
+          isDone = true
+
+          if (maxRetries < 0 || retryIdx < maxRetries) {
+            timeout(ack, retryAfter).map {
```
Avasil, line 62: We don't want to timeout the next elements, we want to delay them specified amount. We could achieve it with:

```scala
scheduler.scheduleOnce(retryAfter){
  ack.syncOnContinue(loop(subscriber, task, retryIdx2 + 1, initialDelay * 2))
}
```

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/OnErrorRetryWithBackoffObservable.scala
```diff
@@ -0,0 +1,93 @@
...
+              case Continue =>
+                loop(subscriber, task, retryIdx + 1, retryAfter * 2)
```
Avasil, line 64: It seems like all delays will start with `retryAfter * 2` instead of `initialDelay`

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
Executioner1939: Will update accordingly :-).

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/OnErrorRetryWithBackoffObservable.scala
Executioner1939: I ran scalafmt as instructed in the CONTRIBUTING.md, I am a little concerned that it might have used IntelliJ's scalafmt config and not the projects. I will fix this.

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
```diff
@@ -0,0 +1,91 @@
...
+    * Example Usage:
+    *
+    * {{{
+    *   import monix.reactive.Observable
+    *
+    *   Observable.from(1).onErrorRestartWithBackoff()
+    * }}}
+    *
+    */
+  final case class Fibonacci private[reactive](var prev0: Int = 0, var prev1: Int = 1) extends BackoffStrategy {
```
Executioner1939, line 83: I am very unsure about this, Gitter says that this won't work in any sort of Concurrent Scenario, a better way to abstract something like this is most likely needed.

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
```diff
@@ -0,0 +1,91 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.reactive
+
+import scala.concurrent.duration.FiniteDuration
+
+abstract class BackoffStrategy extends ((Long, FiniteDuration, FiniteDuration) => FiniteDuration)
```
Executioner1939, line 22: I wasn't sure where the best place would be to locate this was, let me know if it should be somewhere else or removed entirely an implemented in an alternative way.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/OnErrorRetryWithBackoffObservable.scala
```diff
@@ -0,0 +1,93 @@
...
+
+    val cancelable = source.unsafeSubscribeFn(new Subscriber[A] {
+      implicit val scheduler: Scheduler = subscriber.scheduler
+
+      private[this] var isDone = false
+      private[this] var isBackoff = true
+      private[this] var ack: Future[Ack] = Continue
+
+      def onNext(elem: A): Future[Ack] = {
+        isBackoff = false
```
Executioner1939, line 42: The logic with this flag isn't clear but does seem to work, any call to `onNext` sets `isBackoff` to false, indicating a successful operation which means the loop, in the event of an error, is reset with the initial delay and attempt values,

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
Avasil: Is there any discussion somewhere?

Executioner1939: It was just one dude who replied: 

> @Executioner1939 Main concern there is that the state might be tough to manage. This will fail in any sort of concurrent configuration

I don't think it will be a problem though, I don't believe we will have any concurrency errors?

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
```diff
@@ -0,0 +1,88 @@
...
+import scala.annotation.tailrec
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
+
+abstract class BackoffStrategy extends ((Long, FiniteDuration, FiniteDuration) => FiniteDuration)
+
+object BackoffStrategy {
+
+  /**
+    * Implements a Linear Backoff Strategy, for example, and initial delay of 1 second plus a max attempts
```
Avasil, line 29:
```suggestion
    * Implements a Linear Backoff Strategy. For example, an initial delay of 1 second plus a max attempts
```

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
```diff
@@ -0,0 +1,88 @@
...
+    * | 5       | 4 seconds |
+    */
+  case object Linear extends BackoffStrategy {
+    override def apply(attempt: Long, initialDelay: FiniteDuration, currentDelay: FiniteDuration): FiniteDuration = {
+      initialDelay * attempt
+    }
+  }
+
+  /**
+    * Implements an Exponential Backoff Strategy, for example, and initial delay of 1 second plus a max attempts
```
Avasil, line 47:
```suggestion
    * Implements an Exponential Backoff Strategy. For example, an initial delay of 1 second plus a max attempts
```

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -2703,6 +2688,21 @@ abstract class Observable[+A] extends Serializable { self =>
     new OnErrorRetryCountedObservable(self, maxRetries)
   }
 
+  /** Returns an Observable that mirrors the behavior of the source,
```
Avasil, line 2691: I would mention "resetting" aspect of the retry function and add one example from backoff strategy illustrations

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
```diff
@@ -0,0 +1,88 @@
...
+
+abstract class BackoffStrategy extends ((Long, FiniteDuration, FiniteDuration) => FiniteDuration)
+
+object BackoffStrategy {
+
+  /**
+    * Implements a Linear Backoff Strategy, for example, and initial delay of 1 second plus a max attempts
+    * value of 5 would result in the following:
+    *
+    * | Attempt | Delay     |
```
Avasil, line 32: These are awesome!

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
```diff
@@ -0,0 +1,88 @@
...
+    * | ---     | ---       |
+    * | 1       | 0 seconds |
```
Avasil, line 34: Should it be `0 seconds` or `initialDelay`?

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
```diff
@@ -0,0 +1,88 @@
...
+    * | 4       | 8 seconds  |
+    * | 5       | 16 seconds |
+    */
+  case object Exponential extends BackoffStrategy {
+    override def apply(attempt: Long, initialDelay: FiniteDuration, currentDelay: FiniteDuration): FiniteDuration =
+      currentDelay * 2
+  }
+
+  /**
+    * Implements a Fibonacci Backoff Strategy, for example, and initial delay of 1 second plus a max attempts
```
Avasil, line 64:
```suggestion
    * Implements a Fibonacci Backoff Strategy. For example, an initial delay of 1 second plus a max attempts
```

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
```diff
@@ -0,0 +1,88 @@
...
+    private def getNthFibonacciNumber(attempt: Long): Long = {
+      @tailrec def loop(round: Long, a: Long, b: Long): Long = {
+        if (round == attempt) b else loop(round + 1, b, a + b)
+      }
+
+      loop(1, 0, 1)
+    }
+
+    override def apply(attempt: Long, initialDelay: FiniteDuration, currentDelay: FiniteDuration): FiniteDuration =
+      1.second * getNthFibonacciNumber(attempt)
```
Avasil, line 86: Should be `initialDelay` instead of `1.second`, right?

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/OnErrorRetryWithBackoffObservable.scala
```diff
@@ -0,0 +1,93 @@
...
+
+    val cancelable = source.unsafeSubscribeFn(new Subscriber[A] {
+      implicit val scheduler: Scheduler = subscriber.scheduler
+
+      private[this] var isDone = false
+      private[this] var inBackoff = true
+      private[this] var ack: Future[Ack] = Continue
+
+      def onNext(elem: A): Future[Ack] = {
+        inBackoff = false
```
Avasil, line 42: We could do `if (inBackoff) inBackoff = false` - I have a hunch it's a bit more optimal and saw it in other operators 

@alexandru Do you know how to check if JVM can get rid of `if` vs constantly setting `variable` other than benchmark?

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/OnErrorRetryWithBackoffObservable.scala
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+      def onError(ex: Throwable): Unit = {
+        if (!isDone) {
+          isDone = true
+
+          if (maxRetries < 0 || currAttempt < maxRetries) {
+            if (inBackoff) {
+              scheduler.scheduleOnce(currDelay) {
+                ack.syncOnContinue(loop(subscriber, task, currAttempt + 1, strategy(currAttempt + 1, initialDelay, currDelay)))
```
Avasil, line 60: We need to do `try catch`  in case `strategy` throws
Check tests related to "protects against user error" for inspiration

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/OnErrorRestartWithBackoffSuite.scala
```diff
@@ -0,0 +1,42 @@
...
+
+package monix.reactive.internal.operators
+
+import monix.execution.exceptions.DummyException
+import monix.reactive.{BaseTestSuite, Observable}
+
+import scala.concurrent.duration._
+import scala.util.Success
+
+object OnErrorRestartWithBackoffSuite extends BaseTestSuite {
```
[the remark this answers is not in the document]
Avasil: Let's try to use constant backoff for shared tests - if it doesn't work then we'll settle on copy paste

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/OnErrorRestartWithBackoffSuite.scala
```diff
@@ -0,0 +1,70 @@
...
+      .takeByTimespan(5.second)
+      .toListL
+
+    val f = source.runToFuture
+
+    s.tick(5.seconds)
+    assertEquals(f.value, Some(Success(List(0, 1, 0, 1, 0, 1, 0, 1, 0, 1))))
+  }
+
+  private def genBackoffSequence(strategy: BackoffStrategy,
```
Avasil, line 44: I would move it either to the top or the bottom so all the tests are next to each other

Executioner1939: Done

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/OnErrorRetryWithBackoffObservable.scala
Executioner1939: Done

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/OnErrorRestartWithBackoffSuite.scala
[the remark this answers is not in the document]
Executioner1939: @Avasil I couldn't get the shared tests to work, even with a Constant Strategy, I think this is because the attempts also reset after a successful subscription.

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/OnErrorRestartWithBackoffSuite.scala
```diff
@@ -0,0 +1,69 @@
...
+      .range(0, 100)
+      .flatMap(n => if (n == 2) Observable.raiseError(dummy) else Observable.now(n))
+      .onErrorRestartWithBackoff(10, 1.second)
+      .takeByTimespan(5.second)
+      .toListL
+
+    val f = source.runToFuture
+
+    s.tick(5.seconds)
+    assertEquals(f.value, Some(Success(List(0, 1, 0, 1, 0, 1, 0, 1, 0, 1))))
```
Executioner1939, line 52: @Avasil this test is sometimes flaky on my end, it sometimes returns 5 pairs or 6 pairs.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -2703,6 +2688,46 @@ abstract class Observable[+A] extends Serializable { self =>
...
```
Executioner1939, line 2691: @Avasil I have tried to make the docs a little more detailed :-).

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
Executioner1939: 😄 😄

monix-reactive/shared/src/main/scala/monix/reactive/BackoffStrategy.scala
Executioner1939: Fixed 😄 😄  Thanks for the catch!

lukestephenson:
I'm curious about the use case for this.

Issue #1180, opened 2020-05-12
Iterant.fromReactivePublisher throws MatchError: null if completed before started


aakoshh:
We're transforming an `Observable` to an `Iterant` using `fromReactivePublisher`. Occasionally we got `MatchError: null` from [this](https://github.com/monix/monix/blob/3.0.0/monix-tail/shared/src/main/scala/monix/tail/internal/IterantFromReactivePublisher.scala#L119) line, meaning that the `state` of the `IterantSubscriber` is still null, I'm guessing because it ends before it has a chance to be started [here](https://github.com/monix/monix/blob/3.0.0/monix-tail/shared/src/main/scala/monix/tail/internal/IterantFromReactivePublisher.scala#L60). 

I tried to reproduce it in a test. Not sure if my test case is legit, but here goes:
```scala
package test.monix

import monix.eval.Task
import monix.execution.{Cancelable, Scheduler}
import monix.reactive.Observable
import monix.reactive.observers.Subscriber
import monix.tail.Iterant
import org.scalatest._
import scala.concurrent.duration._

class ObservableIterantSpec extends FlatSpec {

  import Scheduler.Implicits.global

  "Iterant.fromReactivePublisher" should "not have a race condition" in {
    val obs = new Observable[Nothing] {
      override def unsafeSubscribeFn(subscriber: Subscriber[Nothing]): Cancelable = {
        subscriber.onComplete()
        Cancelable.empty
      }
    }

    val it = Iterant.fromReactivePublisher[Task, Nothing](obs.toReactivePublisher)

    it.toListL.runSyncUnsafe(1.second)
  }
}
```

And the result:
```
scala.MatchError: null
	at monix.tail.internal.IterantFromReactivePublisher$IterantSubscriber.finish(IterantFromReactivePublisher.scala:119)
	at monix.tail.internal.IterantFromReactivePublisher$IterantSubscriber.onComplete(IterantFromReactivePublisher.scala:152)
	at monix.reactive.internal.rstreams.ReactiveSubscriberAsMonixSubscriber.$anonfun$onComplete$1(ReactiveSubscriberAsMonixSubscriber.scala:91)
	at monix.execution.Ack$AckExtensions$.syncOnContinue$extension(Ack.scala:117)
	at monix.reactive.internal.rstreams.ReactiveSubscriberAsMonixSubscriber.onComplete(ReactiveSubscriberAsMonixSubscriber.scala:91)
[...]

PR #1181, opened 2020-05-13
Fix 'MatchError: null' in Iterant.fromReactivePublisher

aakoshh:
`Iterant.fromReactivePublisher` could encounter `MatchError: null` if the publisher completed the subscription (or I supposed raised an error) before the subscriber could have been initialised.

The PR introduces an `Uninitialized` state to make sure the individual methods matching on it handle the situation.

Fixes https://github.com/monix/monix/issues/1180


Avasil:
Oh, I have just noticed after the fact that the PR targets `monix:3.0.0` branch instead of `master` - would you mind submitting PR again?

alexandru:
Do we need a `monix:3.0.0` branch btw?

It's a little confusing, we should probably delete it.

PR #1184, opened 2020-05-15
Remove `blocking` around reading each element

vasily-kirichenko:
Removed `blocking` around reading each element in CharsReaderObservable, InputStreamObservable and LinesReaderObservable because it has a big performance overhead.

In a test that reads 3GB, 95M lines file `LinesReaderObservable` did ~400K lines per seconds before this fix, and ~7M lines per second after the fix (~17 times faster).


alexandru:
Oh wow, thanks @vasily-kirichenko for spotting this!

PR #1187, opened 2020-05-22
Issue #1177: Fix doAfterSubscribe popping cancelables in the wrong order

### LINKED ISSUE #1177: The doOnSubscriptionCancel side-effect not executed in combination with doAfterSubscribe
_Tested on both 3.1.0 and 3.2.1_

As expected, in following example the `OK` is printed once, as a subscription was cancelled.
```scala
Observable(1, 2, 3)
  .doOnSubscriptionCancel(Task(println("OK")))
  .subscribe().cancel()
```

However, if you modify the source observable with `doAfterSubscribe`, the on-subscription-cancel side effect is never executed. I.e. the `yay` is printed in this example, but `OK` isn't:
```scala
Observable(1, 2, 3)
  .doOnSubscriptionCancel(Task(println("OK")))
  .doAfterSubscribe(Task(println("yay")))
  .subscribe().cancel()
```

Maybe I am not getting the semantics of the `doAfterSubscribe`, but still it seems to me that it should not prevent the original Observable's side-effects to happen.

Attached ScalaTest test case where I even tried to give it a half-second before judging that it didn't happen:
[MonixSubscribeBugSpec.scala.zip](https://github.com/monix/monix/files/4570641/MonixSubscribeBugSpec.scala.zip)

Avasil:
Fixes #1177

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DoOnSubscribeObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DoOnSubscribeObservable.scala
@@ -103,7 +103,7 @@ private[reactive] object DoOnSubscribeObservable {
       )
 
       val ref = SingleAssignCancelable()
-      val conn = StackedCancelable(List(cancelable, ref))
+      val conn = StackedCancelable(List(ref, cancelable))
 
       ref := task.runAsync(new Callback[Throwable, Unit] {
         def onSuccess(value: Unit): Unit = {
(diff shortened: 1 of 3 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DoOnSubscribeObservable.scala
```diff
@@ -103,7 +103,7 @@ private[reactive] object DoOnSubscribeObservable {
       )
 
       val ref = SingleAssignCancelable()
-      val conn = StackedCancelable(List(cancelable, ref))
+      val conn = StackedCancelable(List(ref, cancelable))
```
Avasil, line 106: It is popped when `task` finishes running but `StackedCancelable` works in FIFO order so it was removing `cancelable`

Issue #1105, opened 2020-01-08
Observable getting stuck when consumed with failed Task


TVerse:
I've noticed some strange behaviour when consuming Observables with failing Tasks. The first snippet will will print "1", then the stack trace through the scheduler's uncaught exception handler, and then get stuck until `runSyncUnsafe`'s timeout, and print the `TimeoutException` in the catch block:
```scala
try {
  Observable(1, 2, 3)
    .consumeWith {
      Consumer.foreachTask { i =>
        if (i == 2) Task.raiseError(new RuntimeException("ERROR: " + i.toString))
        else Task(println(i))
      }
    }
    .runSyncUnsafe(5.seconds)
} catch {
  case NonFatal(e) => println("CAUGHT: " + e.getMessage)
}
```
What I would expect to happen instead, is the consumer completing early with the raised exception. The following two snippets do have the correct behaviour, but are not substitutes for the original snippet. This one `flatMap`'s an observable instead:
```scala
try {
  Observable(1, 2, 3)
    .flatMap { i =>
      if (i == 2) Observable.raiseError(new RuntimeException("ERROR: " + i.toString))
      else Observable.fromTask(Task(println(i)))
    }
    .consumeWith(Consumer.complete)
    .runSyncUnsafe(5.seconds)
} catch {
  case NonFatal(e) => println("CAUGHT: " + e.getMessage)
}
```
And this one throws an exception instead of returning a failed Task:
```scala
try {
  Observable(1, 2, 3)
    .consumeWith {
      Consumer.foreachTask { i =>
        if (i == 2) throw new RuntimeException("ERROR: " + i.toString)
        else Task(println(i))
      }
    }
    .runSyncUnsafe(5.seconds)
} catch {
  case NonFatal(e) => println("CAUGHT: " + e.getMessage)
}
```

It seems to be related the the `ForeachAsyncConsumer`. There is an `onError` method, which will signal `Stop` if an error is thrown from `onNext`. However, if the error is not thrown but instead returned in the `Future[Ack]` from `onNext`, the observable never completes and gets stuck until timed out.

Avasil:
Thank you for detailed report!

I think you are right, it is also not consistent with `mapEval`

I think we should use `redeem` instead of `map` here: 
https://github.com/monix/monix/blob/master/monix-reactive/shared/src/main/scala/monix/reactive/internal/consumers/ForeachAsyncConsumer.scala#L44

Are you interested in making a PR?

TVerse:
[...]

PR #1188, opened 2020-05-23
ForeachAsyncConsumer correctly suspending effects (#1105)

paualarco:


paualarco:
@Avasil only ScalaJS tests are failing... since it does not support blocking operations. Shall I move it under `jvm`?

monix-reactive/shared/src/test/scala/monix/reactive/consumers/ForeachAsyncConsumerSuite.scala
```diff
@@ -101,4 +101,21 @@ object ForeachAsyncConsumerSuite extends TestSuite[TestScheduler] {
...
+
+  test("should suspend effects encountered during the stream") { implicit s =>
+    val exceptionMessage: String = "Boom!"
+    val t: Try[Unit] = Try {
+      Observable(1)
+        .consumeWith(
+          Consumer.foreachTask(_ =>
+            Task.raiseError(new RuntimeException(exceptionMessage))
+          )
+        ).runSyncUnsafe()
```
Avasil, line 113: We could use `runToFuture` instead of `runSyncUnsafe` and it will work on ScalaJS too
After `s.tick()` we can check `f.value` to see if it's `Some(Failure(ex))`

paualarco: fixed that :)

monix-reactive/shared/src/test/scala/monix/reactive/consumers/ForeachAsyncConsumerSuite.scala
```diff
@@ -101,4 +102,21 @@ object ForeachAsyncConsumerSuite extends TestSuite[TestScheduler] {
...
+        .consumeWith(
+          Consumer.foreachTask(_ =>
+            Task.raiseError(new RuntimeException(exceptionMessage))
+          )
+        ).runToFuture
+    }
+
+    s.tick()
+    assert(f.value.get.isFailure)
+    assertEquals(f.value.get.failed.get.getMessage, exceptionMessage)
```
Avasil, line 119: We usually use `DummyException` in tests, it can be compared:

```scala
val dummy = DummyException("Boom!")

val f = {
 ...
 Task.raiseError(dummy)
}

s.tick()

assertEquals(f.value, Some(Failure(dummy))
```

Otherwise looks good 👍 
```

paualarco: oh I see! cool will change that

PR #1189, opened 2020-05-23
WIP: Continuous integration via GitHub Actions (#1170)

paualarco:
This PR splits monix tests into `jvm` and `js` and adds automated release process. (In which it was pretty much implemented already).

Unfortunately it [was not possible](https://github.community/t5/GitHub-Actions/reusing-sharing-inheriting-steps-between-jobs-declarations/td-p/37849) to share steps between jobs, therefore `Cache SBT Coursier directory` and `Cache SBT directory` are replicated.


.github/scripts/build
```diff
@@ -23,4 +23,8 @@ if [[ -z "$SBT_COMMAND" ]]; then
     exit 1
 fi
 
-#exec sbt ++$SCALA_VERSION $SBT_COMMAND
+if [ $SCALA_VERSION == '2.11.12' ] && [ $SCALAJS_VERSION == '1.0.0' ]; then
+echo "No tests executed since sjs 1.0.0 and scala 2.11 are not compatible."
```
paualarco, line 27: Notice the workaround here that checks versions before running tests in case they are incompatible, tried to add that check just before to trigger the job but found out that it was not possible to access matrix values [before the initialisation](https://github.community/t5/GitHub-Actions/Matrix-cannot-be-used-in-jobs-level-if/td-p/41048).

Avasil:
Nice! How many tests can run concurrently? If not all, we could remove some configurations, e.g. don't test each Scala version with both jdk versions

.github/scripts/build
paualarco: > Nice! How many tests can run concurrently? If not all, we could remove some configurations, e.g. don't test each Scala version with both jdk versions

So I guess all of them can run concurrently, right now there is a dependency in which scalajs tests does not start until jvm have not finished, but will remove this logic since it does not makes much sense.

paualarco: Removed that dependency, definitely they all run in parallel :D

.github/workflows/build.yml
```diff
@@ -45,10 +73,50 @@ jobs:
...
+          java-version: "adopt@1.${{ matrix.java }}"
+     
+      - name: Run Tests for Java ${{ matrix.java }}, Scala ${{ matrix.scala }}, ScalaJS ${{ matrix.scalajs }}
         run: |
           ./.github/scripts/build
         env:
           SCALA_VERSION: ${{ matrix.scala }}
           SCALAJS_VERSION: ${{ matrix.scalaJS }}
-          PLATFORM: ${{ matrix.platform }}
+          SBT_COMMAND: ci-jvm-mima
```
Avasil, line 106: We could change `ci-jvm-mima` to not run tests, it runs them twice for this version, right?

.github/workflows/build.yml
```diff
@@ -45,10 +73,50 @@ jobs:
...
+          SBT_COMMAND: ci-js
+  
+  mima:
+    name: Mima binary compatibility test
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        java: [ 8 ]
+        scala: [ 2.12.11 ]
```
Avasil, line 91: I think we should check compatibility for the earliest Scala version (2.11)

.github/workflows/build.yml
```diff
@@ -45,10 +73,50 @@ jobs:
...
+      - name: Run Tests for Java ${{ matrix.java }}, Scala ${{ matrix.scala }}, ScalaJS ${{ matrix.scalajs }}
         run: |
           ./.github/scripts/build
         env:
           SCALA_VERSION: ${{ matrix.scala }}
           SCALAJS_VERSION: ${{ matrix.scalaJS }}
-          PLATFORM: ${{ matrix.platform }}
+          SBT_COMMAND: ci-jvm-mima
+      
+  publish:
```
Avasil, line 108: Does it run release after each commit to master?

paualarco: yup - each commit to master for any fork but without the right credentials it should do nothing

.github/workflows/build.yml
paualarco: did not know that mima task was running test too, therefore yes, tests for 2.12.11 are being executed twice.

Avasil: It's defined in build.sbt: https://github.com/monix/monix/blob/master/build.sbt#L22

.github/workflows/build.yml
Avasil: Cool, automated snapshots could be nice for many users waiting for their fixes

.github/workflows/build.yml
paualarco: In order to make this easier, could you tick the tests that needs to be executed and left empty the not required ones?

JVM
- [ ] 8, 2.11
- [ ] 8, 2.11 (by mima check)
- [ ] 8, 2.12
- [ ] 8, 2.13
- [ ] 11, 2.11
- [ ] 11, 2.12
- [ ] 11, 2.13

JS
- [ ] 8, 2.11, 0.6.32
- [ ] 8, 2.12, 0.6.32
- [ ] 8, 2.13, 0.6.32
- [ ] 8, 2.11, 1.0.0 
- [ ] 8, 2.12, 1.0.0 
- [ ] 8, 2.13, 1.0.0
- [ ] 11, 2.11, 0.6.32
- [ ] 11, 2.12, 0.6.32
- [ ] 11, 2.13, 0.6.32
- [ ] 11, 2.11, 1.0.0
- [ ] 11, 2.12, 1.0.0
- [ ] 11, 2.13, 1.0.0

Avasil: Sure, I ticked few boxes - I feel like it's enough to focus on jdk8 + sanity check on one jdk11 combination. I don't know if it matters in ScalaJS

I checked ` 8, 2.11 (by mima check)` but we can do `8, 2.11` and mima check separately if it's easier to configure

.github/scripts/build
paualarco: Note that this PR is not pointing to `master` but to `github-actions`. So before it goes to master, the secrets for automated releases needs to be [configured](https://github.com/olafurpg/sbt-ci-release).

alexandru:
Anything left to do?

The build seems to be failing. Seems to be a spurious fail.

paualarco:
> Anything left to do?
> 
> The build seems to be failing. Seems to be a spurious fail.

There was some tests that were not correctly skipping, although not sure if they were failing because of that.
I have rebased and changed PR's base branch to master.

paualarco:
Not sure why tests still failing after rebasing master, it succeeds some times but most of them it fails on:  `monix.catnap.ConcurrentChannelJVMParallelism8Suite`.

alexandru:
I finally managed to get GH Actions up and running, based on this PR. Closing now.

PR #1195, opened 2020-05-31
Fix nondeterministic behavior in Observable.zip operators

### LINKED ISSUE #1164: Prepending  an element to an Observable before zipping it with another Observable causes a non-deterministic result.
## Versions

JDK: 11
Monix: 3.1.0
Scala: 2.13.1
SBT: 1.3.10

## How to reproduce

I have a project with running code that you can run and reproduce the problem I am facing: https://github.com/wjlow/observable-prepend

This piece of code results in non-deterministic output. 

```scala
import cats.effect.ExitCode
import monix.eval.{Task, TaskApp}

object Main extends TaskApp {
  override def run(args: List[String]): Task[ExitCode] = {
    intermittentFailure.flatMap(x => Task(println(x.mkString("\n")))).map(_ => ExitCode.Success)
  }

  // This is where the problem is
  def intermittentFailure: Task[List[(Int, Int)]] = {
    val streamA = 0 +: Observable(1, 2, 3)
    val streamB = Observable(0, 1, 2, 3)
    streamA.zip(streamB).toListL
  }
}
```

Sometimes, the output is incorrect: 

```
(0,0)
(2,1)
(3,2)
```

Other times, the output is as expected:

```
(0,0)
(1,1)
(2,2)
(3,3)
```

## Additional comments

It looks like replacing `0 +: Observable(1, 2, 3)` with `Observable(Observable(0), Observable(1, 2, 3)).concat` fixes the non-determinism and gives the expected output.

### LINKED ISSUE #1167: Zipping two Observables of different sizes yields non-deterministic result.
## Versions
JDK: 11
Monix: 3.1.0
Scala: 2.13.1
SBT: 1.3.10

## How to reproduce

```
    val task = Observable(1, 2, 3).zip(Observable(Observable(1, 2, 3), Observable.raiseError(new RuntimeException("This should not be zipped"))).concat).completedL
    // Run it 50 times
    Task.traverse(Range(0, 50).toList)(_ => task).map(_ => ExitCode.Success)
```

The first `Observable` contains 3 values, the second `Observable` contains 4 values (the last value is an error).

When zipped, I expect the error to be left out, because there is no value for it to be paired with. However, when I run this many times, it eventually fails with an exception `RuntimeException("This should not be zipped")`, which indicates that `zip` isn't dropping that value.

Changing the second `Observable` to `Observable(1,2,3,4)` results in a program that occasionally does not terminate. I suspect these are moments when it would've thrown an exception in the original case.

I recently raised an issue related to zipping a prepended `Observable`, I don't know if they're related: https://github.com/monix/monix/issues/1164

Avasil:
Fixes #1167 
Fixes #1164

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Zip2Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Zip2Observable.scala
@@ -88,8 +88,9 @@ private[reactive] final class Zip2Observable[A1, A2, +R](obsA1: Observable[A1], 
           }
       }
 
-      continueP.completeWith(lastAck)
+      val oldP = continueP
       continueP = Promise[Ack]()
+      oldP.completeWith(lastAck)
       lastAck
     }
 
@@ -109,7 +110,9 @@ private[reactive] final class Zip2Observable[A1, A2, +R](obsA1: Observable[A1], 
         }
 
       lock.synchronized {
-        if (!hasElem) {
+        // Other source could already set completeWithNext
+        // so it won't emit any elements
+        if (!hasElem || completeWithNext) {
           lastAck match {
             case Continue => rawOnComplete()
             case Stop => () // do nothing
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Zip3Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Zip3Observable.scala
@@ -137,7 +137,7 @@ private[reactive] final class Zip3Observable[A1, A2, A3, +R](
         continueP.trySuccess(Stop)
         lastAck = Stop
       } else {
-        sourcesCompleted += 1.toByte
+        sourcesCompleted += 1
       }
     }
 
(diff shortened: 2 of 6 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Zip2Observable.scala
```diff
@@ -88,8 +88,9 @@ private[reactive] final class Zip2Observable[A1, A2, +R](obsA1: Observable[A1],
           }
       }
 
-      continueP.completeWith(lastAck)
+      val oldP = continueP
       continueP = Promise[Ack]()
+      oldP.completeWith(lastAck)
```
Avasil, line 93: This one is a bit puzzling... The method is called from "synchronized" block but I noticed that `onNext` was allowing the next element in between completing `lastAck` and replacing `continueP` with a new promise

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Zip3Observable.scala
```diff
@@ -56,7 +56,9 @@ private[reactive] final class Zip3Observable[A1, A2, A3, +R](
     // MUST BE synchronized by `lock`
     var continueP = Promise[Ack]()
     // MUST BE synchronized by `lock`
-    var completeWithNext = false
+    var sourcesCompleted: Int = 0
```
Avasil, line 59: Might be faster with N booleans or `Byte` but I'll just leave an issue about it.
It might not matter so I'd rather not spend time on it right now

PR #1197, opened 2020-06-02
Fix memory leak

ppressives:
After upgrading Monix to the latest version I found a memory leak related to the new bracket implementation.

Git bisect shows the culprit commit: 
https://github.com/monix/monix/pull/1120

Below is a simple piece of code that reproduces the issue:
`F.guarantee(F.unit)(F.unit).foreverM` - it fails with OutOfMemoryError on monix.eval.Task, but works fine with cats.effect.IO

I'm not sure if this is a correct fix of the issue and I need help to write a test on this case

### DIFF
--- a/monix-eval/shared/src/main/scala/monix/eval/internal/TaskBracket.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/internal/TaskBracket.scala
@@ -190,7 +190,7 @@ private[monix] object TaskBracket {
                 val fb =
                   try use(value)
                   catch { case NonFatal(e) => Task.raiseError(e) }
-                fb.flatMap(releaseFrame).flatMap(r => Task { conn.pop(); r })
+                fb.flatMap(releaseFrame)
               }
 
               Task.unsafeStartNow(onNext, ctx, cb)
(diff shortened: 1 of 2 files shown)


monix-eval/shared/src/main/scala/monix/eval/internal/TaskBracket.scala
```diff
@@ -190,7 +190,7 @@ private[monix] object TaskBracket {
                 val fb =
                   try use(value)
                   catch { case NonFatal(e) => Task.raiseError(e) }
-                fb.flatMap(releaseFrame)
+                fb.flatMap(releaseFrame).flatMap(r => Task { conn.pop(); r })
```
Avasil, line 193: I feel like we could call `ctx.connection.pop()` in `unsafeRecover` and `unsafeApply` which I removed by mistake in the PR you mentioned

Avasil:
Big thank you!

> I'm not sure if this is a correct fix of the issue and I need help to write a test on this case

I think we're fine skipping the test if it takes a long time to go out of memory

monix-eval/shared/src/main/scala/monix/eval/internal/TaskBracket.scala
ppressives: If I put them there tests fail. I think it breaks the feature you did in that PR.

Avasil: Ah, I see.

I think we could do it in a similar way to Cats IO: https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala#L189

so

```scala
  private[this] val disableUncancelableAndPop: (Any, Throwable, Context, Context) => Context =
    (_, _, old, _) => {
      old.connection.pop()
      old
    }
```

instead of a `null` in `ContextSwitch`.

I'm not able to check right now but the testa should pass since they are passing for Cats IO and hopefully the memory leak will go away :)

ppressives: Sure, it works fine. Thank you!

ppressives:
I rolled back Monix to 3.1.0 in my project, so it can wait for some time.

Issue #891, opened 2019-05-29
Bring back Observable.Transformer?


Avasil:
We used to have ` type Transformer[-A,+B] = Observable[A] => Observable[B]` type alias with corresponding function

```scala
override def transform[B](transformer: Transformer[A, B]): Observable[B] =
    transformer(this)
```

which was removed in `3.x`. It's trivial to implement for the user but I feel like it's quite convenient and many people like this style of transforming streams (looking at the usage of `fs2.through`, `AkkaStream.via` and even [Monix presentation](https://monix.io/presentations/2018-practical-reactive-streams-jacek-kunicki-sbtb.html) etc).

What do you think? And would the syntax stay the same, or shorter?

oleg-py:
I used to like that method, as it makes chaining observable operations easier, but

- `typelevel/mouse` and 2.13s `scala.util.chaining._` already give you more universal syntax
- fs2 has a handful of methods that take it's version (Pipe), not just `through` (e.g. observe, broadcastTo and balanceTo)

So, I'm not convinced it's very valuable going forward, but I'm not a heavy user of `Observable`, so would be open to know if anybody else wants it.

Avasil:
@oleg-py 
> typelevel/mouse and 2.13s scala.util.chaining._ already give you more universal syntax

Good point, though both options require extra import and it will take a while until everyone migrates to 2.13

There are several thumbs up on my first comments and people ask about it from time to time on gitter so I feel like having it available could ease migration. I'm thinking we could either bring it back or have it deprecated but available for now. 

@alexandru Do you remember why it was removed?

babyman:
I pulled this from @Avasil comment in GITTER for anyone who, like me, has been confounded by the code in older presentations

```scala
  implicit class ObservableSyntax[A](val o: Observable[A]) extends AnyVal {
    def transform[B](f: Observable[A] => Observable[B]): Observable[B] = f(o)
  }
```

paualarco:
In my opinion the `transform(Transformer)` signature was very useful and would be very nice to bring it back in future releases :)

PR #1203, opened 2020-06-14
Feature/transformer

paualarco:
**Draft, to gather feedback**

This PR was aimed to brought back the observable's `.transform` signature https://github.com/monix/monix/issues/891 , but while doing it, some ideas came up and just wanted to see how fare they could go. 

From the actual implementation you could have a broadly idea what this would be, but it has not been fully implemented since not really sure if people would like / approve it .

So, with the original idea to bring back the `transform(O[A] => O[B])`, being `Transformation[A, B]` an alias of `O[A] => O[B]`.  

A new builder that would create the `transformer`  was introduced (being similar to the creation of akka `Flow`), allowing then to construct transformations in a more idiomatic way while still being compatible with the same transform signature.

The following code better shows both examples:

```
def transformer(observable: Observable[Int]): Observable[Seq[String]] = {
  observable
     .map(i => i.toString)
     .bufferTumbling(n)
}
```  

The next example shows how transformer would be constructed using `Transformer` builder.
```
val transformer = {
  Transformer  
     .map(i => i.toString)
     .bufferTumbling(n)
     .chain
}
```  

Finally, both example could be used in the same way from on `transform` signature. 

```
val ob: Observable[Int] = ???
ob.transform(transformer)
```  

Again, this builder would just be a different way to do so. I will be comprensible if people don't like it, just to know if its worth to do more work on it :)

### DIFF
--- a/monix-reactive/shared/src/test/scala/monix/reactive/internal/transformer/MapTransformerSuite.scala
+++ b/monix-reactive/shared/src/test/scala/monix/reactive/internal/transformer/MapTransformerSuite.scala
@@ -15,36 +15,46 @@
  * limitations under the License.
  */
 
-package monix.reactive.observables
+package monix.reactive.internal.transformer
 
 import minitest.TestSuite
 import monix.execution.schedulers.TestScheduler
-import monix.reactive.Observable.Transformation
 import monix.reactive.{Observable, Transformer}
+import monix.reactive.Observable.Transformation
 
-import scala.concurrent.duration._
 import scala.concurrent.Await
+import scala.concurrent.duration._
 
-object TransformerSuite extends TestSuite[TestScheduler] {
+object MapTransformerSuite extends TestSuite[TestScheduler] {
   def setup(): TestScheduler = TestScheduler()
   def tearDown(s: TestScheduler): Unit = {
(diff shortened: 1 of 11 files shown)


monix-reactive/shared/src/test/scala/monix/reactive/internal/transformer/MapTransformerSuite.scala
```diff
@@ -0,0 +1,60 @@
...
+    s.tick()
+
+    val f = Observable.now(1).transform(transformer).headL.runToFuture
+    val r = Await.result(f, 1.seconds)
+
+    assertEquals(r, "1")
+  }
+
+  test("should allow chaining multiple map transformations") { implicit s =>
+    val transformer: Transformation[Int, String] = Transformer
```
Avasil, line 47: What are advantages of this syntax versus:

```scala
val transformer: Transformation[Int, String] = 
  _.map(i => i.toString).map(s => s + s)
```

paualarco: @Avasil I guess there is no... maybe it adds bit of consistency on how `flows` are created in akka, but is not relevant at all.
Thanks for opening my eyes, will just not spend more time on this.

Issue #1202, opened 2020-06-12
Support merging observables


ctoomey:
We're using the monix reactive library and love it, thanks!

One of the needs we currently have is to [merge](http://reactivex.io/documentation/operators/merge.html) multiple observables into 1, and more particularly using priorities assigned to the observables, such that the merged observable emits elements from the highest-priority source when multiple sources have new elements available. An example of this feature in akka-streams is [Source.combine](https://doc.akka.io/api/akka/current/akka/stream/scaladsl/Source$.html#combine[T,U](first:akka.stream.scaladsl.Source[T,_],second:akka.stream.scaladsl.Source[T,_],rest:akka.stream.scaladsl.Source[T,_]*)(strategy:Int=%3Eakka.stream.Graph[akka.stream.UniformFanInShape[T,U],akka.NotUsed]):akka.stream.scaladsl.Source[U,akka.NotUsed]) using [MergePrioritized](https://doc.akka.io/api/akka/current/akka/stream/scaladsl/MergePrioritized.html).

I'm envisioning at 2 flavors, a plain, unprioritized merge, and a prioritized one. Is this functionality something you're receptive to incorporating if I contribute an implementation?

Avasil:
> I'm envisioning at 2 flavors, a plain, unprioritized merge, and a prioritized one. Is this functionality something you're receptive to incorporating if I contribute an implementation?

Of course, it will be much appreciated!

We already have a plain merge but a prioritized one would be useful. I would also like to see a merge with limited parallelism (https://github.com/monix/monix/issues/1139)

ctoomey:
@Avasil I'd overlooked the `merge` as I embarassingly didn't grok it's signature and so didn't realize it was doing what I was looking for.

After looking at the code though I see that it doesn't support the normal consumer-driven back-pressure and instead returns immediate `Ack`s upstream and buffers elements per the given `OverflowStrategy`. For our use case we want the normal consumer-driven back pressure, and the closest we can get would be using the `BackPressure` overflow strategy. But that requires a minimum buffer of size 2 and so still wouldn't do what we want. Curious, why isn't a buffer size of 1 allowed?  And even if it were, the implementation would still send an immediate `Ack` upstream, right?

[...]

PR #1205, opened 2020-06-19
1202: prioritized merge

ctoomey:
Incomplete draft for #1202 for review/discussion. Tests to follow as appropriate.


monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
```diff
@@ -0,0 +1,157 @@
...
+
+    // MUST BE synchronized by `lock`
+    var lastAck = Continue: Future[Ack]
+
+    case class PQElem(data: A, promise: Promise[Ack], priority: Int) extends Comparable[PQElem] {
+      override def compareTo(o: PQElem): Int =
+        priority.compareTo(o.priority)
+    }
+
+    val pq = new PriorityBlockingQueue[PQElem](sources.size)
```
Avasil, line 54: Could we use normal `PriorityQueue` instead of a blocking one? It seems like the access is synchronized anyway

ctoomey: Sure, I will look at replacing it and adding necessary synchronization.

Otherwise it looks good?

Avasil: Sorry, I haven't got a moment to take a close look, but it seems to look good. 

I think it's not too far from `mergeMapPrioritized` version If you'd like to go further:
- `onNext` of the source `Observable` (upstream) would add a new `Observable` to the composite instead of doing it from the starting list
- `RefCountObservable` is used to control the number of active children and their lifecycle
- It would be awesome to support different overflow strategies, but it would require to update `BufferedSubscriber` to allow priority queues. The implementation is in `AbstractBackPressuredBufferedSubscriber` for Back-Pressure strategy. Unfortunately I don't know if there is a good implementation of concurrent priority queue out there. I imagine it's too much work, and I don't have capacity to work on it right now so I suppose we can stay with the current behavior which is a back-pressure with bufferSize of 0.

If you don't feel like doing it - that's okay, `mergePrioritizedList` is already better than the status quo

It might also be nice to add a benchmark vs normal merge, but that's just my curiosity and completely optional.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
```diff
@@ -0,0 +1,157 @@
...
+    sources.zip(priorities).foreach { pair =>
+      composite += pair._1.unsafeSubscribeFn(new Subscriber[A] {
+        implicit val scheduler: Scheduler = out.scheduler
+
+        def onNext(elem: A): Future[Ack] = {
+          if (isDone) {
+            Stop
+          } else {
+            pq.add(PQElem(elem, Promise(), pair._2))
+            signalOnNext()
```
gr-alecz, line 144: is this returning the right thing? It looks like this is going to recieve the ack for the next value sent, regardless of whether it's `elem` or not. Should this return the `.future` side of the Promise created on line 143?

ctoomey: Yes you're right, good catch, will fix.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
```diff
@@ -0,0 +1,157 @@
...
+          }
+        }
+
+        def onError(ex: Throwable): Unit =
+          signalOnError(ex)
+
+        def onComplete(): Unit =
+          signalOnComplete()
```
gr-alecz, line 152: I think there's a race here where you can get an `onComplete` before the value passed in the final `onNext` has actually processed.

Avasil: Can you elaborate?

gr-alecz: 1) `onNext` gets called with the final value for this observable, but this observable isn't the highest priority and there's a value waiting, so the `signalOnNext` call triggers the sending of a different value
1) upstream is now exhausted, and isn't obliged to wait for the ack, so it triggers `onComplete`, which calls `signalOnComplete`
1) `onComplete` then increments `completedCount` when there is still an outstanding value for that observable not yet sent
1) If this happens to be the last incomplete observable, then we call `out.onComplete` while there is still at least one value outstanding

ctoomey: Will look into this when I work on this again. I hope to be able to this week.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
ctoomey: Since updating `BufferedSuscriber` would be lots of work I kept my implementation as is. I swapped out the Java `PriorityBlockingQueue` for the Scala `mutable.PriorityQueue`, and added tests.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
ctoomey: Fixed

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
ctoomey: Fixed by adding back `lock.synchronized` to the `onNext` method (I'd dropped that in copying over the code from `CombinedLatestListObservable`). Since `onNext` will be called before `onComplete`, the latter will now not trigger downstream `onComplete` until the last downstream `onNext` completes.

gr-alecz: I'm not sure that actually fixes the scenario I'm describing?

AlecZorab: I'm not sure that actually fixes issue in the situation I've described?

I think the easiest way to make sure that completion doesn't get ahead of the last value is to store the promise created in `onNext`, and then `def onComplete() = lastFuture.foreach(_ => signalOnComplete())`

ctoomey: The key is that so long as there are items the downstream hasn't yet processed, `lastAck` will remain a `Future[Ack]`, and `signalOnComplete` won't call `out.onComplete` until that future completes, i.e., downstream has finished processing all pending items including the last one from the last source.

Take the simplest case of your example: 2 sources w/ 1 item each:
* Source A calls `onNext`, item gets sent downstream: `lastAck = Future[Ack for A's item]`
* Source A calls `onComplete`: `completedCount = 1`
* Source B calls `onNext`, item gets queued: `lastAck = Future[Ack for A's item].flatMap(Future[Ack for B's item])`
* Source B calls `onComplete`: `completedCount = 2; lastAck.onComplete { ... out.onComplete }`

ctoomey:
@Avasil will you go ahead and merge this, or is there anything else needed? Thanks.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -6056,6 +6056,19 @@ object Observable extends ObservableDeprecatedBuilders {
     }
   }
 
+  /** Given an observable sequence and associated priorities, it combines them
+    * into a new observable, preferring higher-priority sources when multiple
+    * sources have items available. If items are available from sources with the
+    * same priority, the order is undefined.
+    */
+  def mergePrioritizedList[A](sources: Observable[A]*)(priorities: Seq[Int]): Observable[A] = {
```
alexandru, line 6064: Very interesting implementation @ctoomey.

I do have one suggestion about this signature ... it's not clear that the `priorities` should match the sources and there's no reason for it.

How about ...

```scala
def mergePrioritizedList[A](sources: (Int, Observable[A])*): Observable[A]
```

Then you could do ...

```scala
Observable.mergePrioritizedList(
  1 -> observable1,
  2 -> observable2,
  //...
)
```

That way the signature is clearer, and you don't need that runtime `assert` in the implementation.

alexandru:
@ctoomey btw, thanks for the contribution, looking good, just dropped a comment . Will try and take it for a test drive.

@Avasil don't hold the merge on my account 🙂

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
ctoomey: Ya I'd prefer that signature too and in fact had it that way originally, but decided to match the way Akka streams did it. I'll change it.

ctoomey:
CI failed due to unrelated test `monix.catnap.ConcurrentChannelJVMParallelism8Suite` failure. Will somebody please re-trigger it?

ctoomey:
Another unrelated test failure. Please re-run the CI, thanks.

Avasil:
I'm not sure about `ConcurrentChannelJVMParallelism8Suite` but I know about `InputObservableSuite` - I'll fix it during the weekend and review your PR. :)

ctoomey:
Thanks @Avasil -- looking forward to having this merged and released -- any idea when you'll do the next release?

Avasil:
I wanted to do the next release with #1213 but it might take a while and we could do something earlier.
We can also release a snapshot any time

ctoomey:
It'd be great to have a release soon after this PR is merged, if that's not too much trouble.

AlecZorab:
I'm not sure if the issue I spotted is fixed or not.

It looks like `signalOnComplete` can still trigger before the final message gets processed, but maybe there's some subtlety to the execution model that I'm missing?

The example that springs to mind:

Observable calls `onNext` with its final value, there are higher priority values waiting, so it gets put into the pqueue. The observable is now complete, so it calls `onComplete`, triggering `signalOnComplete`. This can't acquire the lock, so it blocks. In the async case of `signalOnNext`, there's now a race where the waiting `signalOnComplete` can enter the lock before the value processing happens.

But maybe the system invariants forbid that from happening?

ctoomey:
> I'm wondering about how strict should we be about prioritizing sources.

Good question. I thought more about this and how to document it and came up w/ the following doc. update that'll be in the next commit. One improvement this triggered was changing the order of subscribing to the sources from the given order to prioritized order, such that if multiple sources have items immediately available, the highest priority ones will come through first.

```
  /** Given a sequence of priority/observable pairs, combines them into a new
    * observable that eagerly emits source items downstream as soon as demand is
    * signaled, choosing the item from the highest priority (greater numbers
    * mean higher priority) source when items from multiple sources are
    * available. If items are available from multiple sources with the same
    * highest priority, one of them is chosen arbitrarily.
    *
    * Source items are buffered only to the extent necessary to accommodate
    * backpressure from downstream, and thus if only a single item is available
    * when demand is signaled, it will be emitted regardless of priority.
    *
    * Backpressure is propagated from downstream to the source observables, so
    * that items from a given source will always be emitted downstream in the
    * same order as received from the source, and at most a single item from a
    * given source will be in flight at a time.
    */
```

In your example, the proposed ordering is not guaranteed though, as the sequencing of the sources calling `onNext` is non-deterministic and so per the eager downstream emitting, so is the item ordering.

ctoomey:
> I'm not sure if the issue I spotted is fixed or not.
> 
> It looks like `signalOnComplete` can still trigger before the final message gets processed, but maybe there's some subtlety to the execution model that I'm missing?
> 
> The example that springs to mind:
> 
> Observable calls `onNext` with its final value, there are higher priority values waiting, so it gets put into the pqueue. The observable is now complete, so it calls `onComplete`, triggering `signalOnComplete`. This can't acquire the lock, so it blocks. In the async case of `signalOnNext`, there's now a race where the waiting `signalOnComplete` can enter the lock before the value processing happens.
> 
> But maybe the system invariants forbid that from happening?

I had replied to this with a detailed response explaining why this wasn't an issue, and illustrating with sample output from running the [test](https://github.com/monix/monix/pull/1205/files#diff-95d676c25204b052f111e2b44fbe72a4R89) I added to demonstrate this with additional printlns enabled. That was one of the comments -- along w/ @AlecZorab's description of the possible issue -- that was deleted. Will one of you please undelete my comment? Or if for some reason that's not doable, I'm hoping that one of you has an email notification w/ the contents of my comment that you could post back as a new comment? Thanks. P.S. Why were the comments deleted?

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
```diff
@@ -0,0 +1,164 @@
...
+    def processNext(): Future[Ack] = {
+      val e = pq.dequeue()
+      val fut = rawOnNext(e.data)
+      e.promise.completeWith(fut)
+      fut
+    }
+
+    // MUST BE synchronized by `lock`
+    def signalOnNext(): Future[Ack] =
+      lock.synchronized {
```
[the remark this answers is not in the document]
ctoomey: Fixed

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
```diff
@@ -0,0 +1,164 @@
...
+
+    // MUST BE synchronized by `lock`
+    var completedCount = 0
+
+    // MUST BE synchronized by `lock`
+    def rawOnNext(a: A): Future[Ack] = {
+      if (isDone) Stop else out.onNext(a)
+    }
+
+    def processNext(): Future[Ack] = {
```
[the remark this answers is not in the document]
ctoomey: Fixed

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -6056,6 +6056,19 @@ object Observable extends ObservableDeprecatedBuilders {
     }
   }
 
+  /** Given a sequence of priority/observable pairs, combines them
```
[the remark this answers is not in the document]
ctoomey: Fixed

alexandru:
@ctoomey I haven't had time to look at the implementation in detail.

Question — what happens when the consumer is too slow? Are the source observers back-pressured?

Note that the normal `Observable.merge` uses an underlying back-pressured buffer. An implementation using bounded buffers for a prioritized merge could be more complicated, but I'd like for us to give it a try.

So in case your implementation isn't back-pressured, that is fine for now, but then I'd like to know if you'd be interested in exploring an alternative using bounded buffers afterwards, to make this safer. I can provide you with some ideas if interested.

ctoomey:
> Question — what happens when the consumer is too slow? Are the source observers back-pressured?

Yep they are. I just updated the documentation substantially, see [above](https://github.com/monix/monix/pull/1205#issuecomment-671573631).

alexandru:
@ctoomey 
> P.S. Why were the comments deleted?

TBH I don't know, neither does @Avasil, I think @AlecZorab changed his mind, or maybe he deleted those comments by mistake (maybe he tried to mark the conversation as resolved or smth).

I still have the emails if interested — not sure if this is the answer you had in mind ...

---

@gr-alecz commented on this pull request.
> I'm not sure that actually fixes issue in the situation I've described?
> 
> I think the easiest way to make sure that completion doesn't get ahead of the last value is to store the promise created in > onNext, and then def onComplete() = lastFuture.foreach(_ => signalOnComplete())

@ctoomey commented on this pull request.
> The key is that so long as there are items the downstream hasn't yet processed, lastAck will remain a Future[Ack], and signalOnComplete won't call out.onComplete until that future completes, i.e., downstream has finished processing all pending items including the last one from the last source.
> 
> Take the simplest case of your example: 2 sources w/ 1 item each:
> 
>  * Source A calls onNext, item gets sent downstream: lastAck = Future[Ack for A's item]
>  * Source A calls onComplete: completedCount = 1
>  * Source B calls onNext, item gets queued: lastAck = Future[Ack for A's item].flatMap(Future[Ack for B's item])
>  * Source B calls onComplete: completedCount = 2; lastAck.onComplete { ... out.onComplete }

alexandru:
>> Question — what happens when the consumer is too slow? Are the source observers back-pressured?
>
> Yep they are. I just updated the documentation substantially, see above.

Good 👍 and thanks for the detailed Scaladoc.

ctoomey:
> I still have the emails if interested — not sure if this is the answer you had in mind ...

Thanks @alexandru, yes that was the one, though there was more to it, including the example output from the test w/ printlns. Maybe those came in a subsequent edit.

Avasil:
> Note that the normal Observable.merge uses an underlying back-pressured buffer. An implementation using bounded buffers for a prioritized merge could be more complicated, but I'd like for us to give it a try.
>
> So in case your implementation isn't back-pressured, that is fine for now, but then I'd like to know if you'd be interested in exploring an alternative using bounded buffers afterwards, to make this safer. I can provide you with some ideas if interested.

Could we provide the current implementation with API that would allow us to add customizable back-pressure in later release? 
If I remember correctly, @ctoomey implementation supports `OverflowStrategy.BackPressure(1)` which ironically is not yet possible for other operators :D Though even if we add a new parameter with that default it won't be consistent with other operators where it is `os: OverflowStrategy[B] = OverflowStrategy.Default` and I need to double check if it's binary compatible

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
```diff
@@ -0,0 +1,174 @@
...
+    def signalOnComplete(): Unit =
+      lock.synchronized {
+        completedCount += 1
+
+        if (completedCount == numberOfObservables && !isDone) {
+          lastAck match {
+            case Continue =>
+              isDone = true
+              out.onComplete()
+              completePromises()
```
Avasil, line 119: Regarding the issue:
> The key is that so long as there are items the downstream hasn't yet processed, lastAck will remain a Future[Ack], and signalOnComplete won't call out.onComplete until that future completes, i.e., downstream has finished processing all pending items including the last one from the last source.
>
>Take the simplest case of your example: 2 sources w/ 1 item each:
>
>Source A calls onNext, item gets sent downstream: lastAck = Future[Ack for A's item]
>Source A calls onComplete: completedCount = 1
>Source B calls onNext, item gets queued: lastAck = Future[Ack for A's item].flatMap(Future[Ack for B's item])
>Source B calls onComplete: completedCount = 2; lastAck.onComplete { ... out.onComplete }

Looking at the contract: https://monix.io/docs/3x/reactive/observers.html#contract
> 3. Back-pressure for onComplete and onError is optional: when calling onComplete or onError you are not required to wait on the Future[Ack] of the previous onNext.

It implies that Source B could call:

```scala
out.onNext(lastElem)
out.onComplete()
```
 
In that case, we could send remaining elements from the queue to the downstream instead of completing each promise with `Stop`.

alexandru:
> Could we provide the current implementation with API that would allow us to add customizable back-pressure in later release?
>
>  If I remember correctly, @ctoomey implementation supports OverflowStrategy.BackPressure(1) which ironically is not yet possible for other operators :D Though even if we add a new parameter with that default it won't be consistent with other operators where it is os: OverflowStrategy[B] = OverflowStrategy.Default and I need to double check if it's binary compatible

We can't add an extra parameter, if that's what you mean. But should we add a buffered alternative, can always add an overload, with the same name or with a different name.

alexandru:
Note I'm fine with a `BackPressure(1)` behavior, that's valuable in itself.

If some buffering is desired however, it would probably be incorrect to add it after the merge, because you can buffer lower-priority items, and keep the high priority items from being processed.

So I'd like us to open an issue for a buffered variant too. I'm thinking that each priority can have its own buffer and then a loop would go and select between them, cycling from high priority to low priority.

alexandru:
... speaking of this, this is how RxJava built a kick-ass `merge` operation, last time I checked. They created multiple buffers (concurrent queues), assigning them per observable being merged, to reduce the contention on the producers side. But this isn't very relevant to this PR. Mentioning it to keep it in mind for the future.

Note: I enabled GitHub Actions, so a "stable" "snapshot" version will be published as soon as this PR is merged.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
ctoomey: > In that case, we could send remaining elements from the queue to the downstream instead of completing each promise with `Stop`.

So long as the downstream returns `Continue` and none of the upstreams call `onError`, the downstream will always get all the items emitted by the upstreams. That's what the "should push all items downstream before calling onComplete" test is verifying. Again:

> The key is that so long as there are items the downstream hasn't yet processed, lastAck will remain a Future[Ack], and signalOnComplete won't call out.onComplete (or completePromises) until that future completes, i.e., downstream has finished processing all pending items including the last one from the last source.

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
```diff
@@ -0,0 +1,194 @@
...
+              Stop
+            case Continue => Continue
+          }
+          .recover {
+            case NonFatal(e) =>
+              lock.synchronized(completePromises())
+              throw e
+          }
+      }
+    }
```
ctoomey, line 92: Added the above logic to ensure that we complete the upstream onNext promises when the downstream stops or errors early while we still have pending upstream items. Added tests to verify as well.

ctoomey:
> So I'd like us to open an issue for a buffered variant too. I'm thinking that each priority can have its own buffer and then a loop would go and select between them, cycling from high priority to low priority.

Created #1227

> Note: I enabled GitHub Actions, so a "stable" "snapshot" version will be published as soon as this PR is merged.

Great, thanks @alexandru

AlecZorab:
So having done some println debugging, I'm convinced that the issue I highlighted isn't an issue any more. The subtlety in the implementation that I missed is in the way the chaining of futures works in `signalOnNext`. For the early completion I was thinking about to occur, you'd have to have an item in the pqueue waiting to be sent, whilst also having `lastAck == Continue`, which isn't a reachable state.

@ctoomey, my apologies for taking so long to understand what you were talking about

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/MergePrioritizedListSuite.scala
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  // Enables verification that MergePrioritizedListObservable completes
+  // upstream onNext promises when downstream stops early
+  private class OnNextExposingObservable[+A](elem: A)(implicit s: Scheduler) extends Observable[A] {
+    var onNextRes: Option[Future[Ack]] = None
+
+    def unsafeSubscribeFn(subscriber: Subscriber[A]): Cancelable = {
+      val onNext = subscriber.onNext(elem)
+      onNext.onComplete(_ => subscriber.onComplete())
```
AlecZorab, line 166: since it's allowed in the contract, it might be a better test to change this to call `onComplete` immediately, rather than waiting for `onNext` to complete. Having tried it, the test still passes, so it's really just a case of making it a little bit more stringent

monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/MergePrioritizedListObservable.scala
Avasil: We could use our `syncOnStopOrFailure` syntax here, should be more efficient

ctoomey: Much nicer, thanks.

ctoomey:
> So having done some println debugging, I'm convinced that the issue I highlighted isn't an issue any more. The subtlety in the implementation that I missed is in the way the chaining of futures works in `signalOnNext`. For the early completion I was thinking about to occur, you'd have to have an item in the pqueue waiting to be sent, whilst also having `lastAck == Continue`, which isn't a reachable state.
> 
> @ctoomey, my apologies for taking so long to understand what you were talking about

No problem, good to be sure for this kind of thing.

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/MergePrioritizedListSuite.scala
ctoomey: Done

ctoomey:
@Avasil will you please merge if there's nothing else? Thanks

alexandru:
@Avasil at this point I think we should just merge, and let @ctoomey test with the published version. If there's anything wrong, I'm sure he'll be around to help.

ctoomey:
> @Avasil at this point I think we should just merge, and let @ctoomey test with the published version. If there's anything wrong, I'm sure he'll be around to help.

For sure

Avasil:
Awesome, I've merged into `series/3.x` but I'm not sure if it has an automatic release now?

Avasil:
And how do we go about keeping consistency between series, should we open a PR from this branch and target 4.x.?

alexandru:
> Awesome, I've merged into series/3.x but I'm not sure if it has an automatic release now?

It did, I introduced a parameter in our secrets, called `AUTO_PUBLISH_SERIES_3X`, and it was set to `false`. I've set it to `true` and re-running the build, will let you know when it's ready.

> And how do we go about keeping consistency between series, should we open a PR from this branch and target 4.x.?

Unfortunately we have to start making incompatible changes, and maintaining two branches is inevitable, because the 4.x release will not be stable enough to release for a couple of months.

So yes, new developments that are compatible with Monix 3.x should target the `series/3.x` branch, which is now the main branch, and then we can back-port the changes. Fortunately we have been using "squash" when merging, keeping our history linear. So I just did this in the `series/4.x` branch:

```scala
git cherry-pick 1a5b9b1b41eaa2a1e955302e741a6daab1ff73f6
```

In this case it was clear, no PRs needed, I just committed it. The author could do that too, in which case a PR would be needed. And a PR would also be needed in case of conflicts.

alexandru:
We have version `3.2.2+37-1a5b9b1b` published.

ctoomey:
> We have version `3.2.2+37-1a5b9b1b` published.

@alexandru the `monix-internal-jctools` artifact is missing:

```
[info] downloading https://repo1.maven.org/maven2/io/monix/monix_2.12/3.2.2+37-1a5b9b1b/monix_2.12-3.2.2+37-1a5b9b1b.jar ...
[info] downloading https://repo1.maven.org/maven2/io/monix/monix-catnap_2.12/3.2.2+37-1a5b9b1b/monix-catnap_2.12-3.2.2+37-1a5b9b1b.jar ...
[info] downloading https://repo1.maven.org/maven2/io/monix/monix-eval_2.12/3.2.2+37-1a5b9b1b/monix-eval_2.12-3.2.2+37-1a5b9b1b.jar ...
[info] downloading https://repo1.maven.org/maven2/io/monix/monix-reactive_2.12/3.2.2+37-1a5b9b1b/monix-reactive_2.12-3.2.2+37-1a5b9b1b.jar ...
[info] downloading https://repo1.maven.org/maven2/org/typelevel/cats-effect_2.12/2.1.4/cats-effect_2.12-2.1.4.jar ...
[info] downloading https://repo1.maven.org/maven2/io/monix/monix-tail_2.12/3.2.2+37-1a5b9b1b/monix-tail_2.12-3.2.2+37-1a5b9b1b.jar ...
[info] downloading https://repo1.maven.org/maven2/io/monix/monix-java_2.12/3.2.2+37-1a5b9b1b/monix-java_2.12-3.2.2+37-1a5b9b1b.jar ...
[info] downloading https://repo1.maven.org/maven2/io/monix/monix-execution_2.12/3.2.2+37-1a5b9b1b/monix-execution_2.12-3.2.2+37-1a5b9b1b.jar ...
[info] 	[SUCCESSFUL ] io.monix#monix_2.12;3.2.2+37-1a5b9b1b!monix_2.12.jar (393ms)
[info] 	[SUCCESSFUL ] io.monix#monix-java_2.12;3.2.2+37-1a5b9b1b!monix-java_2.12.jar (776ms)
[info] 	[SUCCESSFUL ] io.monix#monix-catnap_2.12;3.2.2+37-1a5b9b1b!monix-catnap_2.12.jar (1109ms)
[info] 	[SUCCESSFUL ] io.monix#monix-tail_2.12;3.2.2+37-1a5b9b1b!monix-tail_2.12.jar (1255ms)
[info] 	[SUCCESSFUL ] org.typelevel#cats-effect_2.12;2.1.4!cats-effect_2.12.jar (1354ms)
[info] 	[SUCCESSFUL ] io.monix#monix-eval_2.12;3.2.2+37-1a5b9b1b!monix-eval_2.12.jar (1358ms)
[info] 	[SUCCESSFUL ] io.monix#monix-execution_2.12;3.2.2+37-1a5b9b1b!monix-execution_2.12.jar (1365ms)
[info] 	[SUCCESSFUL ] io.monix#monix-reactive_2.12;3.2.2+37-1a5b9b1b!monix-reactive_2.12.jar (1728ms)
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: io.monix#monix-internal-jctools_2.12;3.2.2+37-1a5b9b1b: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 
[warn] 	Note: Unresolved dependencies path:
[warn] 		io.monix:monix-internal-jctools_2.12:3.2.2+37-1a5b9b1b (/Users/christoomey/unix/liv/service-template/api/build.sbt#L40-75)
[warn] 		  +- io.monix:monix-execution_2.12:3.2.2+37-1a5b9b1b (/Users/christoomey/unix/liv/service-template/api/build.sbt#L40-75)
[warn] 		  +- io.monix:monix-catnap_2.12:3.2.2+37-1a5b9b1b (/Users/christoomey/unix/liv/service-template/api/build.sbt#L40-75)
[warn] 		  +- io.monix:monix-eval_2.12:3.2.2+37-1a5b9b1b (/Users/christoomey/unix/liv/service-template/api/build.sbt#L40-75)
[warn] 		  +- io.monix:monix-reactive_2.12:3.2.2+37-1a5b9b1b (/Users/christoomey/unix/liv/service-template/api/build.sbt#L40-75)
[warn] 		  +- io.monix:monix_2.12:3.2.2+37-1a5b9b1b (/Users/christoomey/unix/liv/service-template/api/build.sbt#L40-75)
```

alexandru:
@ctoomey sorry, I updated JCTools, and in the process I ended up "shading" it, but forgot to "aggregate" it, so it didn't get published.

We're fixing it.

alexandru:
@ctoomey try `3.2.2+41-615286ba`

ctoomey:
Thanks @alexandru, looks good now.

PR #1209, opened 2020-07-02
Brings back transformer (#891)

paualarco:
Brings back `transform` signature and `type Transformer[+A, -B] = Observable[A] => Observable[B]`.

Some extra motivations of this PR are that the current `monix-dynamodb` already exposes this operator in its api, and it is also shown in the [documentation](https://monix.github.io/monix-connect/docs/dynamodb#transformer), which could create confusions.

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -4712,7 +4712,10 @@ object Observable extends ObservableDeprecatedBuilders {
     */
   type Operator[-I, +O] = Subscriber[O] => Subscriber[I]
 
-  /** A `Transformer` is a function used for transforming observables */
+  /** A `Transformer` is a function used for transforming observables.
+    *
+    * See [[Observable.transform]]
+    */
   type Transformer[-A, +B] = Observable[A] => Observable[B]
 
   /** Given a sequence of elements, builds an observable from it. */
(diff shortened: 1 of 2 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -4707,6 +4712,9 @@ object Observable extends ObservableDeprecatedBuilders {
     */
   type Operator[-I, +O] = Subscriber[O] => Subscriber[I]
 
+  /** A `Transformer` is a function used for transforming observables */
```
Avasil, line 4715: I would add a reference to Observable.transform, like in `liftByOperator`

Avasil:
I'm wondering where to put the alias - `Observable` companion object or `reactive` package object

paualarco:
> I'm wondering where to put the alias - Observable companion object or reactive package object
It might be more intuitive to put it in the `reactive` package.
And that should also apply to `Operator[-I, +O]`, don't you think?

Avasil:
I think it's fine for `Operator` because it is mostly used internally anyway and moving it would break compatibility

paualarco:
> I'm wondering where to put the alias - Observable companion object or reactive package object

Will address the change whenever there is a final decision :-)

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -4664,6 +4664,11 @@ abstract class Observable[+A] extends Serializable { self =>
     Task.create { (s, onFinish) =>
       unsafeSubscribeFn(new ForeachSubscriber[A](cb, onFinish, s))
     }
+
+  /** Transforms the source using the given transformer function. */
+  final def transform[B](transformer: Transformer[A, B]): Observable[B] =
```
alexandru, line 4669: This should probably not be final. This is because it's sometimes cool to override that return type in things that inherit from `Observable`. For example `transform` on a `Subject` could return a `Subject`.

PR #1217, opened 2020-07-22
Fix performance regression in bufferSliding

Avasil:
I also updated to Scala 2.13.3 - sorry for the noise

2.12:

```
Benchmark                                         (chunkCount)  (chunkSize)   Mode  Cnt   Score   Error  Units
ChunkedEvalFilterMapSumBenchmark.monixObservable          1000         1000  thrpt   20  66.211 ± 2.483  ops/s
```

2.13, before the change:
```
Benchmark                                         (chunkCount)  (chunkSize)   Mode  Cnt   Score   Error  Units
ChunkedEvalFilterMapSumBenchmark.monixObservable          1000         1000  thrpt   20  35.212 ± 3.410  ops/s (2.13 before)
```

2.13, after the change:
```
Benchmark                                         (chunkCount)  (chunkSize)   Mode  Cnt   Score   Error  Units
ChunkedEvalFilterMapSumBenchmark.monixObservable          1000         1000  thrpt   20  72.209 ± 3.425  ops/s (2.13 after)
```

### DIFF
--- a/monix-execution/shared/src/main/scala_2.13+/monix/execution/compat.scala
+++ b/monix-execution/shared/src/main/scala_2.13+/monix/execution/compat.scala
@@ -31,5 +31,8 @@ object compat {
     def hasDefiniteSize[X](i: IterableOnce[X]): Boolean = i.knownSize >= 0
 
     def newBuilder[From, A, C](bf: BuildFrom[From, A, C], from: From): mutable.Builder[A, C] = bf.newBuilder(from)
+
+    @inline def toSeq[A](array: Array[AnyRef]): Seq[A] =
+      new scala.collection.immutable.ArraySeq.ofRef(array).asInstanceOf[Seq[A]]
   }
 }
--- a/benchmarks/shared/src/main/scala/monix/benchmarks/ChunkedEvalFilterMapSumBenchmark.scala
+++ b/benchmarks/shared/src/main/scala/monix/benchmarks/ChunkedEvalFilterMapSumBenchmark.scala
@@ -1,0 +1,152 @@
+/*
+ * Copyright (c) 2014-2020 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.benchmarks
+
+import java.util.concurrent.TimeUnit
+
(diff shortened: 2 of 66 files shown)


monix-execution/shared/src/main/scala_2.13+/monix/execution/compat.scala
```diff
@@ -31,5 +31,8 @@ object compat {
     def hasDefiniteSize[X](i: IterableOnce[X]): Boolean = i.knownSize >= 0
 
     def newBuilder[From, A, C](bf: BuildFrom[From, A, C], from: From): mutable.Builder[A, C] = bf.newBuilder(from)
+
+    @inline def toSeq[A](array: Array[AnyRef]): Seq[A] =
+      new scala.collection.immutable.ArraySeq.ofRef(array).asInstanceOf[Seq[A]]
```
Avasil, line 36: This is the change

Avasil:
I'm running benchmarks again and the results are all over the place, I'll rerun it over the night with Slack/Chrome/etc. closed

Avasil:
I'm running benchmarks again and the results are all over the place, I'll rerun it over the night with Slack/Chrome/etc. closed

monix-execution/shared/src/main/scala_2.13+/monix/execution/compat.scala
alexandru: It will be hard to keep track of these internal helpers.

Avasil: Any ideas? Alternatively, it could be placed in `reactive` (it has only one usage) but it would be even more scattered then.
I have put it in `private [monix]` in case we can delete it one day.

Initially I wanted to point to the same `ArraySeq` but mutable one doesn't perform well in my benchmarks. It seems like `toSeq` is quite expensive there.

Avasil:
I've run it more times and it seems to be OK:

```
2.13 this PR:
[info] Benchmark                                         (chunkCount)  (chunkSize)   Mode  Cnt   Score   Error  Units
[info] ChunkedEvalFilterMapSumBenchmark.monixObservable          1000         1000  thrpt   30  77.968 ± 3.677  ops/s


2.13 master:
[info] Benchmark                                         (chunkCount)  (chunkSize)   Mode  Cnt   Score   Error  Units
[info] ChunkedEvalFilterMapSumBenchmark.monixObservable          1000         1000  thrpt   30  53.559 ± 0.364  ops/s

2.12:
[info] Benchmark                                         (chunkCount)  (chunkSize)   Mode  Cnt   Score   Error  Units
[info] ChunkedEvalFilterMapSumBenchmark.monixObservable          1000         1000  thrpt   30  65.132 ± 1.613  ops/s
```

Still a lot of variance in some results between runs but there's consistent trend that it might be even slightly faster on Scala 2.13 now

monix-execution/shared/src/main/scala_2.13+/monix/execution/compat.scala
alexandru: That's fine.

alexandru: We'll have to remember we have it I guess.

alexandru:
I think that's fine. You would help me if you merged this, given your 2.13 upgrade 🙂

PR #1218, opened 2020-07-22
Dependency updates 2020-07

alexandru:
The problematic change I'm introducing is that JCTools is now a "shaded" library, meaning that it's getting published under a new `monix-internal-jctools` artifact and its classes moved to `monix.execution.internal.jctools`.

Making this change because other libraries depend on JCTools and we can end up in conflict with them.

Will have to test the generated artifacts.


Avasil:
BTW I have just updated to Scala 2.13.3 in https://github.com/monix/monix/pull/1217
There were lots of errors about lack of `()`

project/plugins.sbt
```diff
@@ -1,13 +1,15 @@
...
-addSbtPlugin("com.eed3si9n"         % "sbt-unidoc"      % "0.4.3")
-addSbtPlugin("pl.project13.scala"   % "sbt-jmh"         % "0.3.7")
-addSbtPlugin("com.typesafe"         % "sbt-mima-plugin" % "0.7.0")
-addSbtPlugin("com.typesafe.sbt"     % "sbt-git"         % "1.0.0")
-addSbtPlugin("org.xerial.sbt"       % "sbt-sonatype"    % "3.9.4")
-addSbtPlugin("de.heikoseeberger"    % "sbt-header"      % "5.6.0")
-addSbtPlugin("org.scalameta"        % "sbt-scalafmt"    % "2.4.0")
-addSbtPlugin("com.github.tkawachi"  % "sbt-doctest"     % "0.9.6")
-addSbtPlugin("org.scoverage"        % "sbt-scoverage"   % "1.6.1")
+val scalaJSVersion = Option(System.getenv("SCALAJS_VERSION")).filter(_.nonEmpty).getOrElse("1.1.1")
```
Avasil, line 1: I'm not sure we want to do this: https://www.scala-js.org/news/2020/05/18/announcing-scalajs-1.1.0/


> It is backward binary compatible with all earlier versions in the 1.x series: libraries compiled with 1.0.x can be used with 1.1.0 without change.
It is not forward binary compatible with 1.0.x: libraries compiled with 1.1.0 cannot be used with 1.0.x.

alexandru: Yikes, that sucks.

alexandru:
@Avasil I'll wait for that PR then, as I also wanted to update Scala and preferably to add the `sbt-tpolecat` plugin.

project/plugins.sbt
alexandru: So I think we should go ahead with `1.1.1`. It's not the same restrictions as with the JVM, because the Scala.js compiler is needed when swapping a JAR anyway and from what I understand it will complain if the library isn't compatible.

Avasil: 👍

alexandru:
I'm also preparing a PR activating `sbt-tpolecat` and fixing a ton of compiler warnings 😐

I will merge this one for now. If you disagree with the shading, let me know.

The artifacts need to be tested anyway, so we'll have to publish a sha release.

PR #1221, opened 2020-07-26
Scala 2.13.3 + sbt-tpolecat, fix all code for activated linting options

### LINKED ISSUE #1030: CharsReaderObservableSuite has a flaky test
It sometimes fails on CI with 
```
[info] - fromCharsReader closes the file handle on cancel *** FAILED ***
[info]   received Some(Success(())) != expected None (CharsReaderObservableSuite.scala:181)
```

e.g. in https://travis-ci.org/monix/monix/jobs/585428074

### LINKED ISSUE #1078: flaky test AsyncSchedulerSuite
`monix.execution.internal.AsyncSchedulerSuite` is failing frequently these days without any specific reason.

alexandru:
I actually discovered some issues while fixing the code, mostly in tests, but also in the implementation.

Implicit conversion from `Int` to `Long` can actually create issues, even if no information is lost. For example `request(n: Long)` from `org.reactivestreams.Subscriber` could be used with a `request: Int` by mistake, thus artificially limiting the size of requests.

Ditto for discarding `non-Unit` values. E.g. when subscribing to an `Observable`, we actually have to think if the returned `Cancelable` can be discarded or not. I discovered an instance in which it wasn't really OK to ignore it (see the new `TakeLastObservable`, replacing the old `TakeLastOperator`).

Also:

- fixes #1030
- fixes #1078


alexandru:
This is a massive PR, containing mostly minor fixes in order to appease the linting options, with a couple of exceptions.

Will merge this, and taking responsibility for it 😅, b/c I'd like a sha release, and to unblock other PRs.

PR #1230, opened 2020-08-11
Fix #1229: Drop support for Scala 2.11

### LINKED ISSUE #1229: Drop support for Scala 2.11
Scala 2.11 is costing us developer resources that we don't have, and as much as I'd like to maintain it, the ecosystem has moved on.

- Cats no longer builds for 2.11
- Cats-Effect no longer builds for 2.11
- sbt has moved to Scala 2.12 some time ago
- Spark has moved to Scala 2.12, finally

What sparked this desire is that publishing to Sonatype for Scala 2.11 is currently broken. This is because Scaladoc packaging fails on Scala 2.11.12, with the compiler crashing. And disabling Scaladoc packaging for Scala 2.11 is not possible, as it would fail when uploading to Sonatype b/c the staging servers on Sonatype require doc packages.

Current issue: #1231 

**UPDATE:** we could still wait for as long as we can. Meaning the 3.x series. But after Cats-Effect 3.0, which isn't source compatible, we'll have no choice.

alexandru:
See #1229 for details.

Changes:

- removed 2.11 from CI (`build.yml`)
- removed 2.11 specific source directories from `build.sbt`
- refactored the code:
  - changed anonymous `Runnable` instances to SAMs (wherever I spotted instances that could be changed)
  - removed source-specific compatibility hacks meant for 2.11
  - got rid of `scala_2.12+` directories, since that code essentially becomes valid for all supported Scala versions
- updated the `README`


alexandru:
Note that I've switched the base to `series/4.x`.

PR #1244, opened 2020-08-24
Add support for compression: Gzip and deflate

ghostbuster91:
This PR adds operators to compress/decompress stream of bytes using gzip and deflate algorithms.

Most of the heavy work was done by my colleague @LGLO while implementing this feature in ZIO (https://github.com/zio/zio/pull/3825)

I've just adapted his work to monix. I also did some improvements based on the fs2 code (added option to provide `fileName`, `comment` and `modificationTimestamp` when compressing a stream).

Not sure if the packages are ok, looking forward for your review.


monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
```diff
@@ -0,0 +1,172 @@
...
+      * @param bufferSize             Size of internal buffer used for pulling data from deflater, affects performance.
+      * @param chunkSize              Size of chunks, for most of the use-cases should be equal to bufferSize
+      * @param compressionParameters  See [[CompressionParameters]]
+      * @param noWrap                 Whether output stream is wrapped in ZLIB header and trailer. For HTTP 'deflate' content-encoding should be false, see https://tools.ietf.org/html/rfc2616.
+      */
+    def deflate(
+      bufferSize: Int = 32 * 1024,
+      chunkSize: Int = 32 * 1024,
+      compressionParameters: CompressionParameters = CompressionParameters.Default,
+      noWrap: Boolean = false
```
ghostbuster91, line 140: Should we keep it as `Boolean` or model as in fs2: https://github.com/typelevel/fs2/blob/main/core/jvm/src/main/scala/fs2/compression.scala#L15 ?

monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/InflateOperator.scala
```diff
@@ -0,0 +1,142 @@
...
+
+private[compress] final class InflateOperator(bufferSize: Int, noWrap: Boolean)
+  extends Operator[Array[Byte], Array[Byte]] {
+
+  def apply(out: Subscriber[Array[Byte]]): Subscriber[Array[Byte]] =
+    new Subscriber[Array[Byte]] {
+      implicit val scheduler = out.scheduler
+
+      private[this] var isDone = false
+      private[this] var ack: Future[Ack] = _
```
Avasil, line 41: Seems like it could be `Continue`

monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/InflateOperator.scala
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+      def onComplete(): Unit =
+        if (!isDone) {
+          isDone = true
+          if (ack == null) ack = Continue
+          ack.syncOnContinue {
+            try {
+              out.onNext(inflater.finish())
+            } catch {
```
Avasil, line 75: Similar issue as `GunzipOperator` - it could call both `onError` and `onComplete`

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
```diff
@@ -0,0 +1,172 @@
...
+
+import java.time.Instant
+import java.util.zip.Deflater
+
+import monix.reactive.compress.internal.operators.{DeflateOperator, GunzipOperator, GzipOperator, InflateOperator}
+
+import scala.collection.compat.immutable.ArraySeq
+
+// From https://github.com/typelevel/fs2/blob/main/core/jvm/src/main/scala/fs2/compression.scala
+package object compress {
```
Avasil, line 28: I'm leaning towards `compression` to be consistent with both fs2 and ZIO

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
```diff
@@ -0,0 +1,172 @@
...
+  private[compress] object gzipExtraFlag {
+    val DEFLATE_MAX_COMPRESSION_SLOWEST_ALGO: Byte = 2
+    val DEFLATE_FASTEST_ALGO: Byte = 4
+  }
+
+  private[compress] object gzipCompressionMethod {
+    val DEFLATE: Byte = Deflater.DEFLATED.toByte
+  }
+
+  implicit class RichByteObservable(source: Observable[Byte]) {
```
Avasil, line 73: Alternatively, we could use `Transformer` that was brought recently but it would be a precedence. :D

I feel like it could be a bit nicer and more discoverable (`source.transform(gzip)`) than syntax import.
I would also move the methods to the top of the file so users don't have to scroll through private fields if they explore the file

WDYT?

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
```diff
@@ -0,0 +1,172 @@
...
+      * @return
+      */
+    def gzip(
+      bufferSize: Int = 32 * 1024,
+      chunkSize: Int = 32 * 1024,
+      compressionParameters: CompressionParameters = CompressionParameters.Default,
+      fileName: Option[String] = None,
+      comment: Option[String] = None,
+      modificationTime: Option[Instant] = None
+    ): Observable[Byte] = {
```
Avasil, line 93: Is it desirable to return `Observable[Byte]` ?

I feel like it would be better to take and return `Observable[Array[Byte]]` and if the users need `Observable[Byte]` then they could do buffering and concatenation themselves.

fs2 and ZIO return Stream[Byte] but in reality, it's a stream of Chunk[Byte] which has to be explicit in `Observable`

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
Avasil: I'm personally fine with API without `DeflateParams` overload

monix-reactive/jvm/src/test/scala/monix/reactive/compress/DeflateIntegrationTest.scala
```diff
@@ -0,0 +1,93 @@
...
+
+import monix.eval.Task
+import monix.reactive.Observable
+import org.scalacheck.Prop
+
+object DeflateIntegrationTest extends BaseTestSuite with DeflateTestUtils {
+  private implicit def a[A]: Task[Boolean] => Prop =
+    _.runSyncUnsafe()
+
+  test("inflate(deflate(_)) == identity") {
```
Avasil, line 28: Nitpick but we usually describe it as `<->` - feel free to ignore the comment :D

monix-reactive/jvm/src/test/scala/monix/reactive/compress/InflateTest.scala
```diff
@@ -0,0 +1,101 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.reactive.compress
+
+import minitest.api.AssertionException
+import monix.reactive.Observable
+
+object InflateTest extends BaseTestSuite with DeflateTestUtils {
```
Avasil, line 23: I feel like we could reuse a lot of test cases between `InflateTest` and `GunzipTest` - do you think it's worth it?

Avasil (commented):
Thank you for the hard work, it looks very good and the test coverage is fantastic. :)

I've left some comments, mainly nitpicks.
I didn't fully review compression logic and I only glanced at the tests, I'll have another pass later, PR is quite big 😅

monix-reactive/jvm/src/main/scala/monix/reactive/compress/CompressionParameters.scala
```diff
@@ -0,0 +1,64 @@
...
+
+object CompressionParameters {
+  val Default = CompressionParameters(
+    CompressionLevel.DefaultCompression,
+    CompressionStrategy.DefaultStrategy,
+    FlushMode.NoFlush
+  )
+}
+
+sealed abstract class CompressionLevel(val jValue: Int)
```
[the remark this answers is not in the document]
ghostbuster91: Changed to `value`

monix-reactive/jvm/src/main/scala/monix/reactive/compress/CompressionParameters.scala
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.reactive.compress
+
+case class CompressionParameters(
```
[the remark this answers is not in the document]
ghostbuster91: Done

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
ghostbuster91: If the `Transformer` is a new way of doing such things then we should go with it.

I don't know if it is nicer or more discoverable but certainly it would be easier to use it from java. 

I will take a look at `Transformer` and try to use it.

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
ghostbuster91: When you read data in a streaming fashion in binary format (e.g. from file) you mostly get `Observable[Byte]`, so that would require users to always do buffering before using any compression transformer. Also they would need to know what chunk size to choose. Choosing too small or too big value can cause performance degradation. This explains my intention behind using `Observable[Byte]` as the input.

To not introduce any confusion and be transparent for the end users I decided to do not change the type and return `Observable[Byte]`

monix-reactive/jvm/src/test/scala/monix/reactive/compress/InflateTest.scala
ghostbuster91: Good point, done.

ghostbuster91:
Last but not least fs2 returns `GunzipResult` when using `gunzip` transformer which contains some additional metadata along with te decompressed stream. I decided not to go this way since it seems unnatural for observables but I would like to know your opinion about that.

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
ghostbuster91: I was wrong, you might get `Stream[Byte]` but that's rather popular among libraries that chunk data under the hood (fs2/zio) which effectively means that you always deal with `Stream[Array[Byte]]` either explicitly or not. That's good as transforming bytes one by one might not be the optimal solution in terms of performance.

ghostbuster91: Also monix-nio returns `Observable[Array[Byte]]` and if we should be compatible with anything, that should be monix-nio.

Avasil: As another reason for `Array`, Akka Streams (which doesn't do chunking) uses `ByteString` which extends `IndexedSeq[Byte]`

> When you read data in a streaming fashion in binary format (e.g. from file) you mostly get Observable[Byte], so that would require users to always do buffering before using any compression transformer.

Related operators, like `Observable.fromInputStream` return `Observable[Array[Byte]]` and if we add more, I feel like we should emit `Array[Byte]` for performance reasons, as you mentioned.

>  Also they would need to know what chunk size to choose. 

That's a fair point, maybe we could add a comment with a suggestion/example how to use it with `Observable[Byte]`?

monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/GunzipOperator.scala
```diff
@@ -0,0 +1,334 @@
...
+          out.onError(ex)
+        }
+
+      def onComplete(): Unit =
+        if (!isDone) {
+          isDone = true
+          if (ack == null) ack = Continue
+          ack.syncOnContinue {
+            try {
+              out.onNext(gunzipper.finish())
```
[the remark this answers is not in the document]
Avasil: I think it should be:

```scala
      def onComplete(): Unit =
        if (!isDone) {
          isDone = true
          if (ack == null) ack = Continue

          ack.syncOnComplete {
            case Success(Continue) =>
              // Protect against contract violations - we are only allowed to
              // call onError if no other terminal event has been called.
              var streamErrors = true
              try {
                val lastArray = gunzipper.finish()
                streamErrors = false
                out.onNext(lastArray)
                out.onComplete()
              } catch {
                case NonFatal(e) if streamErrors =>
                  out.onError(e)
              } finally {
                gunzipper.close()
              }

            case _ => gunzipper.close()
          }
        }
```

For the following reasons:
1. We need to call `gunzipper.close()` even if downstream returned Stop / failed Future
2. If there is an error on `gunzipper.finish()` and we handle it by calling `out.onError`, we don't want to call `onComplete` as well
3. If there is an exception in `onNext` (protocol breakage) then we don't call onError in case it already happened downstream

monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/GzipOperator.scala
```diff
@@ -0,0 +1,246 @@
...
+          isDone = true
+          gzipper.close()
+          out.onError(ex)
+        }
+
+      def onComplete(): Unit =
+        if (!isDone) {
+          isDone = true
+          if (ack == null) ack = Continue
+          ack.syncOnContinue {
```
Avasil, line 91: I expect that we need to call `gzipper.close()` on `Stop` / `Failure` too

monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/GzipOperator.scala
```diff
@@ -0,0 +1,246 @@
...
+            out.onNext(gzipper.finish())
```
Avasil, line 92: gzipper.finish() never throws, right?

monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/GunzipOperator.scala
```diff
@@ -0,0 +1,334 @@
...
+        if (isDone) {
+          Stop
+        } else {
+          try {
+            val result = gunzipper.onChunk(elem)
+            // signaling downstream
+            ack = out.onNext(result)
+            ack
+          } catch {
+            case e if NonFatal(e) =>
```
Avasil, line 60: We need to do `streamErrors` protection here as well

monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/InflateOperator.scala
```diff
@@ -0,0 +1,142 @@
...
+        if (isDone) {
+          Stop
+        } else {
+          try {
+            val result = inflater.onChunk(elem)
+            // signaling downstream
+            ack = out.onNext(result)
+            ack
+          } catch {
+            case e if NonFatal(e) =>
```
Avasil, line 54: We need to do `streamErrors` protection here as well

monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/InflateOperator.scala
Avasil: I explained in this comment: https://github.com/monix/monix/pull/1244/files#r495495768

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
Avasil: Let's rename it to `compression` unless you have good arguments against it

Avasil (commented):
Sorry for the delay, I've made another pass - I've resolved cosmetic comments and elaborated on those that should be fixed. If you don't have time then let me know and I can finish it - it already looks fantastic!

Avasil:
> Last but not least fs2 returns `GunzipResult` when using `gunzip` transformer which contains some additional metadata along with the decompressed stream. I decided not to go this way since it seems unnatural for observables but I would like to know your opinion on that.

Yeah, I think that's fine. If there's ever a demand, we could create an extra method but working with stream of streams could be awkward for the general use case and I assume you're implementing it because you need it in sttp/tapir

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
Avasil: I don't use it too frequently myself, but many people like it, it's the equivalent of `via` / `transduce` / `through` etc

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
paualarco: Ideally we would write a small benchmark on both to compare their performance.

ghostbuster91: About that:
> That's a fair point, maybe we could add a comment with a suggestion/example how to use it with Observable[Byte]?

I thought that maybe such helpful extension method could be declared for those who will wonder?

  ```
final def chunked[B >: A](chunkSize :Int = 8 * 1024)(implicit ev: B =:= Byte) : Observable[Array[Byte]] = {
    bufferTumbling(chunkSize)
      .map(seq => ev.liftCo[Seq](seq).toArray)
  }
```

To make it more discover-able I would define it on `Observable` itself although that's a precedence.

File: monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/GzipOperator.scala (hunk 13 above)
Comment (ghostbuster91) [reply monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/GzipOperator.scala:92]: It seems that it can, should be fixed now.

File: monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/GunzipOperator.scala (hunk 11 above)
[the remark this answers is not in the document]
Comment (ghostbuster91) [reply monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/GunzipOperator.scala:80]: Yes, absolutely. I've just finished your recent talk about monix-reactive internals, so now it is all clear.

File: monix-reactive/jvm/src/main/scala/monix/reactive/compress/internal/operators/DeflateOperator.scala [hunk 16]
```diff
@@ -0,0 +1,109 @@
...
+          out.onError(ex)
+        }
+
+      def onComplete(): Unit =
+        if (!isDone) {
+          isDone = true
+          if (ack == null) ack = Continue
+          ack.syncOnContinue {
+            out.onNext(deflate.finish())
+            deflate.close()
```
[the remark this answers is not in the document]
ghostbuster91: Yes, should be fixed now.

ghostbuster91:
> Sorry for the delay, I've made another pass - I've resolved cosmetic comments and elaborated on those that should be fixed. If you don't have time then let me know and I can finish it - it already looks fantastic!

No problem, I know how it is ;) Thanks for the careful code-review. I think that now the code should comply with monix-reactive protocol.

One think that bothers me is that this inconsistency with existing protocol wasn't caught by tests. Correct me if I am wrong, but I think that you mentioned (during your last presentation) something about having a generic test for checking protocol compliance. Maybe I've just forgot to implement some trait in tests?

monix-reactive/jvm/src/main/scala/monix/reactive/compress/package.scala
Avasil: I think this operator might make sense for performance scenarios, not only for `Byte`.

`bufferTumbling` uses `Array` as a buffer but it converts it to `Seq` before sending downstream.
We could copy-paste implementation and send the copy of the array without the conversion.

Although it would be better to do with benchmarks (there are some that use bufferTumbling and it's a bottleneck) so I feel like it would be preferable to tackle in another PR

Avasil:
> One think that bothers me is that this inconsistency with existing protocol wasn't caught by tests. Correct me if I am wrong, but I think that you mentioned (during your last presentation) something about having a generic test for checking protocol compliance. Maybe I've just forgot to implement some trait in tests?

There is `BaseOperatorSuite` - I've forgotten to mention it on this PR. Now I'm actually curious how many violations would it catch :D

ghostbuster91:
We can check it, but I am not sure how to combine this `BaseOperatorSuite` with existing decompression tests. Ideally I would mix it into `GunzipTest`, `GzipTest`, `InflateTest` and `DeflateTest` but this is an abstract class (any reason for that?) so I can't. Also I think that I need to rename these classes to suits to follow convention. On the other hand the `BaseOperatorSuite` is about something different than these tests are so maybe it should go into separate files?

Avasil:
>  Even if I could I would have to use this testScheduler which requires manual ticking which seems irrelevant from the perspective of these tests. 

Many other operators are synchronous too so that shouldn't be a problem. 
For custom test cases, you can just supply normal Scheduler if you prefer to use it

> Also I think that I need to rename these classes to suits to follow convention. On the other hand the `BaseOperatorSuite` is about something different than these tests are so maybe it should go into separate files?

Usually, we just add test cases below implementation of `BaseOperatorSuite`

I don't know how well will the test suite fit for these operators so feel free to skip it if there are too many problems

Avasil (approved):
It looks great to me - thanks once again :) 

Although few tests seem to fail

PR #1251, opened 2020-08-24
Mention Monix BIO in the readme

Avasil:
Not sure where to put it but I want to include it


alexandru:
Sure.

We should also list `monix-bio` and `monix-connect` in a new sub-section.

PR #1265, opened 2020-09-12
Implement Observable#bufferWhile and bufferWhileInclusive

### LINKED ISSUE #1260: Add Observable.bufferWhile and bufferWhileInclusive
@ghostbuster91 has suggested a nice buffering operator

```scala
def bufferWhile(p: A => Boolean): Observable[Seq[A]]
def bufferWhileInclusive(p: A => Boolean): Observable[Seq[A]]
```

The former is similar to `bufferWithSelector` if we use hot `Observable`:

```scala
def bufferWhile(p: A => Boolean): Observable[Seq[A]] =
  publishSelector { hot =>
    hot.bufferWithSelector(hot.filterNot(p))
  }
```

but `inclusive` implementation could be more tricky.

I think both methods could be quite easy to write as native operators - a combination of `BufferSlidingOperator` and `TakeByPredicateOperator`. 

I can add more details if anyone is interested in contributing

Avasil:
Fixes #1260

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/BufferWhileOperator.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/BufferWhileOperator.scala
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2014-2020 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/BufferWhileOperator.scala
```diff
@@ -0,0 +1,101 @@
...
+          try {
+            val keepBuffering = p(elem)
+            streamError = false
+
+            if (!keepBuffering) {
+              if (inclusive) {
+                val toEmit = buffer.append(elem).toList
+                buffer = ListBuffer.empty
+                ack = out.onNext(toEmit)
+              } else if (buffer.nonEmpty) {
```
Avasil, line 52: Not 100% sure if we should emit empty buffer if it's `false` for the first element, or not

PR #1267, opened 2020-09-13
Implement Asynchronous Stack Traces for Task

### LINKED ISSUE #1015: Better stack traces for Task/Coeval/IO? Possible solutions?
For post `3.0.0` lets discuss possible solutions for better stack traces when using `Task` and `Coeval`.

We'll probably avoid thinking about this now, as we are busy with the `3.0.0` release ... but opening this now in order to not forget about it.

Avasil:
Fixes #1015

Latest snapshot version: `3.2.2+55-9adf112c-SNAPSHOT`

This is a direct port of `cats.effect.IO` solution, [released in 2.2.0](https://github.com/typelevel/cats-effect/releases/tag/v2.2.0).
[Tracing docs](https://typelevel.org/cats-effect/tracing/)

See the following PR's for more context:
- https://github.com/typelevel/cats-effect/pull/854
- https://github.com/typelevel/cats-effect/pull/981
- https://github.com/typelevel/cats-effect/pull/1077

So pretty much all credit goes to @RaasAhsan and @djspiewak 👏 

TODO:
- [x] Proper attributions
- [x] Documentation, created issue #1300
- [x] Double check if all internals are private
- [x] Consider reusing classes with Cats-Effect after upgrade to 2.2.0
- [x] Benchmarks
- [x] Consider tracing more operators (e.g. race, parTraverse) created issue #1288

The feature works exactly the same as in Cats-Effect and is enabled by default. 
It can be configured the same way as Cats version, which is [documented here](https://typelevel.org/cats-effect/tracing/)

The prefix is `monix.eval` instead of `cats.effect` e.g.:

`-Dmonix.eval.stackTracingMode=full -Dmonix.eval.traceBufferLogSize=5`

Example of the current state:

```scala
package test.app

import monix.eval.Task
import cats.implicits._

object TestTracingApp extends App {
  implicit val s = Scheduler.global

  def customMethod: Task[Unit] =
    Task.now(()).guarantee(Task.sleep(10.millis))

  val tracingTestApp: Task[Unit] = for {
    _ <- Task.shift
    _ <- Task.unit.attempt
    _ <- (Task(println("Started the program")), Task.unit).parTupled
    _ <- customMethod
    _ <- if (true) Task.raiseError(new Exception("boom")) else Task.unit
  } yield ()

  tracingTestApp.onErrorHandleWith(ex => Task(ex.printStackTrace())).runSyncUnsafe
}
```

Old stack trace:

```
java.lang.Exception: boom
        at test.app.TestTracingApp$.$anonfun$tracingTestApp$5(TestTracingApp.scala:36)
        at monix.eval.internal.TaskRunLoop$.startFull(TaskRunLoop.scala:188)
        at monix.eval.internal.TaskRestartCallback.syncOnSuccess(TaskRestartCallback.scala:101)
        at monix.eval.internal.TaskRestartCallback$$anon$1.run(TaskRestartCallback.scala:118)
        at monix.execution.internal.Trampoline.monix$execution$internal$Trampoline$$immediateLoop(Trampoline.scala:66)
        at monix.execution.internal.Trampoline.startLoop(Trampoline.scala:32)
        at monix.execution.schedulers.TrampolineExecutionContext$JVMNormalTrampoline.super$startLoop(TrampolineExecutionContext.scala:142)
        at monix.execution.schedulers.TrampolineExecutionContext$JVMNormalTrampoline.$anonfun$startLoop$1(TrampolineExecutionContext.scala:142)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
        at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:94)
        at monix.execution.schedulers.TrampolineExecutionContext$JVMNormalTrampoline.startLoop(TrampolineExecutionContext.scala:142)
        at monix.execution.internal.Trampoline.execute(Trampoline.scala:40)
        at monix.execution.schedulers.TrampolineExecutionContext.execute(TrampolineExecutionContext.scala:57)
        at monix.execution.schedulers.BatchingScheduler.execute(BatchingScheduler.scala:50)
        at monix.execution.schedulers.BatchingScheduler.execute$(BatchingScheduler.scala:47)
        at monix.execution.schedulers.AsyncScheduler.execute(AsyncScheduler.scala:31)
        at monix.eval.internal.TaskRestartCallback.onSuccess(TaskRestartCallback.scala:72)
        at monix.eval.internal.TaskRunLoop$.startFull(TaskRunLoop.scala:183)
        at monix.eval.internal.TaskRestartCallback.syncOnSuccess(TaskRestartCallback.scala:101)
        at monix.eval.internal.TaskRestartCallback.onSuccess(TaskRestartCallback.scala:74)
        at monix.eval.internal.TaskSleep$SleepRunnable.run(TaskSleep.scala:71)
        at java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1402)
        at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
        at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
        at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
        at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
```

Cached Mode:

```
java.lang.Exception: boom
        at test.app.TestTracingApp$.$anonfun$tracingTestApp$5(TestTracingApp.scala:36)
        at guarantee @ test.app.TestTracingApp$.customMethod(TestTracingApp.scala:29)
        at flatMap @ test.app.TestTracingApp$.$anonfun$tracingTestApp$4(TestTracingApp.scala:35)
        at parTupled @ test.app.TestTracingApp$.$anonfun$tracingTestApp$2(TestTracingApp.scala:34)
        at parTupled @ test.app.TestTracingApp$.$anonfun$tracingTestApp$2(TestTracingApp.scala:34)
        at flatMap @ test.app.TestTracingApp$.$anonfun$tracingTestApp$2(TestTracingApp.scala:34)
        at flatMap @ test.app.TestTracingApp$.$anonfun$tracingTestApp$1(TestTracingApp.scala:33)
        at flatMap @ test.app.TestTracingApp$.delayedEndpoint$test$app$TestTracingApp$1(TestTracingApp.scala:32)
```

Full Tracing Mode:

```
java.lang.Exception: boom
        at test.app.TestTracingApp$.$anonfun$tracingTestApp$5(TestTracingApp.scala:36)
        at raiseError @ test.app.TestTracingApp$.$anonfun$tracingTestApp$5(TestTracingApp.scala:36)
        at now @ test.app.TestTracingApp$.customMethod(TestTracingApp.scala:29)
        at guarantee @ test.app.TestTracingApp$.customMethod(TestTracingApp.scala:29)
        at flatMap @ test.app.TestTracingApp$.$anonfun$tracingTestApp$4(TestTracingApp.scala:35)
        at <clinit> @ test.app.TestTracingApp$.delayedEndpoint$test$app$TestTracingApp$1(TestTracingApp.scala:32)
        at apply @ test.app.TestTracingApp$.$anonfun$tracingTestApp$2(TestTracingApp.scala:34)
        at parTupled @ test.app.TestTracingApp$.$anonfun$tracingTestApp$2(TestTracingApp.scala:34)
        at parTupled @ test.app.TestTracingApp$.$anonfun$tracingTestApp$2(TestTracingApp.scala:34)
        at flatMap @ test.app.TestTracingApp$.$anonfun$tracingTestApp$2(TestTracingApp.scala:34)
        at <clinit> @ test.app.TestTracingApp$.delayedEndpoint$test$app$TestTracingApp$1(TestTracingApp.scala:32)
        at flatMap @ test.app.TestTracingApp$.$anonfun$tracingTestApp$1(TestTracingApp.scala:33)
        at flatMap @ test.app.TestTracingApp$.delayedEndpoint$test$app$TestTracingApp$1(TestTracingApp.scala:32)
```

### DIFF
--- a/monix-eval/jvm/src/main/java/monix/eval/internal/TracingPlatform.java
+++ b/monix-eval/jvm/src/main/java/monix/eval/internal/TracingPlatform.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2014-2020 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.eval.internal;
+
+import java.util.Optional;
+
+/**
+  * All Credits to https://github.com/typelevel/cats-effect and https://github.com/RaasAhsan
+  *
+  * Holds platform-specific flags that control tracing behavior.
+  *
+  * The Scala compiler inserts a volatile bitmap access for module field accesses.
+  * Because the `tracingMode` flag is read in various Task combinators, we are opting
+  * to define it in a Java source file to avoid the volatile access.
+  *
+  * INTERNAL API.
+  */
+public final class TracingPlatform {
+
+  /**
+    * Sets stack tracing mode for a JVM process, which controls
+    * how much stack trace information is captured.
+    * Acceptable values are: NONE, CACHED, FULL.
+    */
+  private static final String stackTracingMode = Optional.ofNullable(System.getProperty("monix.eval.stackTracingMode"))
+    .filter(x -> !x.isEmpty())
+    .orElse("cached");
+
+  public static final boolean isCachedStackTracing = stackTracingMode.equalsIgnoreCase("cached");
+
+  public static final boolean isFullStackTracing = stackTracingMode.equalsIgnoreCase("full");
+
+  public static final boolean isStackTracing = isFullStackTracing || isCachedStackTracing;
+
+  /**
+    * The number of trace lines to retain during tracing. If more trace
+    * lines are produced, then the oldest trace lines will be discarded.
+    * Automatically rounded up to the nearest power of 2.
+    */
+  public static final int traceBufferLogSize = Optional.ofNullable(System.getProperty("monix.eval.traceBufferLogSize"))
+    .filter(x -> !x.isEmpty())
+    .flatMap(x -> {
+      try {
+        return Optional.of(Integer.valueOf(x));
+      } catch (Exception e) {
+        return Optional.empty();
+      }
+    })
+    .orElse(4);
+
+  /**
+    * Sets the enhanced exceptions flag, which controls whether or not the
+    * stack traces of IO exceptions are augmented to include async stack trace information.
+    * Stack tracing must be enabled in order to use this feature.
+    * This flag is enabled by default.
+    */
+  public static final boolean enhancedExceptions = Optional.ofNullable(System.getProperty("monix.eval.enhancedExceptions"))
+    .map(Boolean::valueOf)
+    .orElse(true);
+
+}
--- a/monix-eval/shared/src/main/scala/monix/eval/internal/TaskTracing.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/internal/TaskTracing.scala
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2014-2020 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.eval.internal
+
+import java.util.concurrent.ConcurrentHashMap
+
+import monix.eval.Task
+import monix.eval.Task.Trace
+import monix.eval.tracing.TaskEvent
+
+/**
+  * All Credits to https://github.com/typelevel/cats-effect and https://github.com/RaasAhsan
+  */
+private[eval] object TaskTracing {
+
+  def decorated[A](source: Task[A]): Task[A] =
+    Trace(source, buildFrame())
+
+  def uncached(): TaskEvent =
+    buildFrame()
+
+  def cached(clazz: Class[_]): TaskEvent =
+    buildCachedFrame(clazz)
+
+  private def buildCachedFrame(clazz: Class[_]): TaskEvent = {
+    val currentFrame = frameCache.get(clazz)
+    if (currentFrame eq null) {
+      val newFrame = buildFrame()
+      frameCache.put(clazz, newFrame)
+      newFrame
+    } else {
+      currentFrame
+    }
+  }
+
+  private def buildFrame(): TaskEvent =
+    TaskEvent.StackTrace(new Throwable().getStackTrace.toList)
+
+  /**
+    * Global cache for trace frames. Keys are references to lambda classes.
+    * Should converge to the working set of traces very quickly for hot code paths.
+    */
+  private[this] val frameCache: ConcurrentHashMap[Class[_], TaskEvent] = new ConcurrentHashMap()
+
+}
--- a/monix-eval/shared/src/main/scala/monix/eval/tracing/TaskTrace.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/tracing/TaskTrace.scala
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2014-2020 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.eval.tracing
+
+import monix.eval.Task
+
+import scala.reflect.NameTransformer
+
+/**
+  * All Credits to https://github.com/typelevel/cats-effect and https://github.com/RaasAhsan
+  */
+final case class TaskTrace(events: List[TaskEvent], captured: Int, omitted: Int) {
+
+  import TaskTrace._
+
+  def printFiberTrace(options: PrintingOptions = PrintingOptions.Default): Task[Unit] =
+    Task(System.err.println(showFiberTrace(options)))
+
+  def showFiberTrace(options: PrintingOptions = PrintingOptions.Default): String = {
+    val TurnRight = "╰"
+    val InverseTurnRight = "╭"
+    val Junction = "├"
+    val Line = "│"
+
+    val acc0 = s"TaskTrace: $captured frames captured\n"
+    if (options.showFullStackTraces) {
+      val stackTraces = events.collect { case e: TaskEvent.StackTrace => e }
+
+      val acc1 = stackTraces.zipWithIndex
+        .map {
+          case (st, index) =>
+            val tag = getOpAndCallSite(st.stackTrace)
+              .map {
+                case (methodSite, _) =>
+                  NameTransformer.decode(methodSite.getMethodName)
+              }
+              .getOrElse("(...)")
+            val op = if (index == 0) s"$InverseTurnRight $tag\n" else s"$Junction $tag\n"
+            val relevantLines = st.stackTrace
+              .slice(options.ignoreStackTraceLines, options.ignoreStackTraceLines + options.maxStackTraceLines)
+            val lines = relevantLines.zipWithIndex
+              .map {
+                case (ste, i) =>
+                  val junc = if (i == relevantLines.length - 1) TurnRight else Junction
+                  val codeLine = renderStackTraceElement(ste)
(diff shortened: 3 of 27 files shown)


monix-eval/jvm/src/main/java/monix/eval/internal/TracingPlatform.java
```diff
@@ -0,0 +1,76 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix.eval.internal;
+
+import java.util.Optional;
+
+/**
+  * All Credits to https://github.com/typelevel/cats-effect and https://github.com/RaasAhsan
```
Avasil, line 23: I have just added this comment to all files that I copy-pasted with minimal changes

I'm not sure how should I give credits properly (and in a legal way) - I'm more than happy to follow any suggestions @RaasAhsan , @djspiewak

monix-eval/shared/src/main/scala/monix/eval/tracing/TaskTrace.scala
```diff
@@ -0,0 +1,139 @@
...
+package monix.eval.tracing
+
+import monix.eval.Task
+
+import scala.reflect.NameTransformer
+
+/**
+  * All Credits to https://github.com/typelevel/cats-effect and https://github.com/RaasAhsan
+  */
+final case class TaskTrace(events: List[TaskEvent], captured: Int, omitted: Int) {
```
Avasil, line 27: I'm wondering whether classes like that should be reused with Cats-Effect, or not.
I suppose it's safer to have it separate and create a type class if there are any libraries that operate on it.

monix-eval/jvm/src/main/java/monix/eval/internal/TracingPlatform.java
djspiewak: IMO this is pretty much all that is really needed. From an etiquette standpoint, this seems ample. From a legal standpoint, both Monix and CE are under ASLv2, so derivative works like this are entirely expected and normal.

monix-eval/shared/src/main/scala/monix/eval/tracing/TaskTrace.scala
djspiewak: Another idea would be to slurp this out into a common library which both can share. You would probably also be interested in the `ConcurrentHashMap` replacement that we're working on for 2.2.1, which should shrink the performance hit pretty dramatically.

jvican: @djspiewak that looks interesting, where can I learn more about what replacement you're preparing for `ConcurrentHashMap`?

djspiewak: @jvican https://github.com/typelevel/cats-effect/pull/1101

It's not a *general* replacement. CHM is actually insanely good at what it does. But tracing is a rather weird case. The cache converges to a relatively small size (strictly bounded by the number of classes) and then usually never changes again. Cache misses aren't fatal, they're just poorly optimized, and writes are not *mandatory* (follows from the miss bit). But within the context of all of that, we want reads to be *blisteringly* fast once fully converged.

In our synthetic benchmarks, tracing imposes a roughly 16% performance penalty on programs' straight-line performance (i.e. long chains of `flatMap`s). From what we can tell, nearly all of that is tied up in just *reading* from `ConcurrentHashMap`, due to the memory barriers which are necessary to ensure proper consistency, and secondarily due to the fact that the table doesn't aggressively optimize for a small and convergent set of keys. We can relax both of those constraints with a custom implementation and theoretically shave quite a bit off of that performance penalty.

Still a work in progress. :-)

Avasil: > Another idea would be to slurp this out into a common library which both can share.

That would be nice. Most of the code seems to be nicely separated and quite generic

Currently, I'm fine with copy-pasting and keeping an eye on updates but I will probably get tired of it pretty quickly, considering I plan to do the same for `Coeval` and `monix-bio` in the future

> You would probably also be interested in the ConcurrentHashMap replacement that we're working on for 2.2.1, which should shrink the performance hit pretty dramatically.

Sounds awesome!
Could you use any help?

jvican: @djspiewak thanks for elaborating, will follow that thread. I asked because I, too, believe CHM is incredibly good at what it does and it's an amazing fit for many broad use cases. It makes sense to try to go for a barrier-free approach for tracing.

djspiewak: > That would be nice. Most of the code seems to be nicely separated and quite generic
>
> Currently, I'm fine with copy-pasting and keeping an eye on updates but I will probably get tired of it pretty quickly, considering I plan to do the same for Coeval and monix-bio in the future

We should definitely pull it upstream into a new lib then. This is also relevant for CE3, where we want to reimplement tracing, but it would be nice to not have to duplicate things. Ideas on a name?

RaasAhsan: 👍  on factoring out code for reusability. I would start with something like the `UnsafeConvergentCache` because it's actually very general-purpose and initialization-safe, thread-unsafe data structures may have applications outside tracing. Pulling out tracing-specific code like `IOTrace`/`TaskTrace` is also an awesome idea, but my only caution is that functionality will most certainly evolve in the CE2/CE3/Monix world (e.g. thread tracking will be introduced to CE2 soon), and saying that they won't diverge is probably short-sighted. It would be unfortunate to end up in a situation where the library prescribes too much on implementations or implementations are restricted by the library, but I think we can definitely find a middle ground. Ultimately if things don't work out, we should be able to sever the dependency easily (keeping use of that library internal is probably a reasonable precaution).

> Currently, I'm fine with copy-pasting and keeping an eye on updates but I will probably get tired of it pretty quickly, considering I plan to do the same for Coeval and monix-bio in the future

If there are any improvements, it would be awesome to see those back in CE as well! :D But that's a really good reason to share all this code.

>  I rather like Sacagawea (tracking, tracing, etc), partially because Native American names and words are sorely underrepresented in the software space, and partially because she's cool and it seems like a distinctive name.

Sacagawea is an awesome name :)

monix-eval/shared/src/main/scala/monix/eval/internal/TaskTracing.scala
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  private def buildFrame(): TaskEvent =
+    TaskEvent.StackTrace(new Throwable().getStackTrace.toList)
+
+  /**
+    * Global cache for trace frames. Keys are references to lambda classes.
+    * Should converge to the working set of traces very quickly for hot code paths.
+    */
+  private[this] val frameCache: ConcurrentHashMap[Class[_], TaskEvent] = new ConcurrentHashMap()
```
retronym, line 58: Echoing my comment from https://github.com/typelevel/cats-effect/pull/1101#issuecomment-691845566

Storing strong references to the lambda classes as the key of the frame cache smells like a potential classloader leak when the lambda class is loaded from a classloader with a shorter lifecycle than cats-eval's classloader.

`java.lang.ClassValue` is designed to avoid such leaks. It's also designed to be extremely fast for lookups.

monix-eval/shared/src/main/scala/monix/eval/tracing/TaskTrace.scala
djspiewak: > Sacagawea is an awesome name :)

I can create a repo and set up some infra for it a little later this morning and you guys can shovel things in there as you see fit. :-) If it turns out to not be a good idea, it's always easy to un-shovel.

djspiewak: https://github.com/typelevel/sacagawea

Avasil:
BTW @jvican - do you have any performance tests in bloop? 
If so, could you run them with `3.2.2+55-9adf112c-SNAPSHOT` ? 
I imagine that bloop will probably turn off stack traces anyway but it would be great to see what's the impact on real applications. 

In microbenchmarks, I experience 10-30% slow-down for cached tracing

jvican:
Hey @Avasil, I have performance tests in bloop but bloop hasn't yet migrated to 3.x, it's something I'm planning to get around to soon, so unfortunately we won't be able to use our performance checks for a real-world performance evaluation. 
  
> In microbenchmarks, I experience 10-30% slow-down for cached tracing
  
What do you mean by cached tracing? Will this be the baseline when tracing is enabled by default?

Avasil:
> What do you mean by cached tracing? Will this be the baseline when tracing is enabled by default?

There are multiple configurations of tracing: https://typelevel.org/cats-effect/tracing/#configuration

It is going to be default but it can be disabled if someone notices any performance degradation

monix-eval/shared/src/main/scala/monix/eval/tracing/TaskTrace.scala
Avasil: Okay, I think I'll try to settle this PR as is, and then we could move towards keeping internals in `sacagawea`. Hopefully, I can find some improvements to pay you back a little. :)

I've counted only three public pieces (apart from configuration):

- `final case class IOTrace(events: List[IOEvent], captured: Int, omitted: Int)`
- `IOEvent`
- `PrintingOptions`

And it should be perfectly fine to keep it per library, mostly cosmetic stuff apart from `IOTrace`.
In case of `IOTrace`, we could just use a type class in any future libraries.

> Pulling out tracing-specific code like IOTrace/TaskTrace is also an awesome idea, but my only caution is that functionality will most certainly evolve in the CE2/CE3/Monix world (e.g. thread tracking will be introduced to CE2 soon)

Could you elaborate on this feature? 
I'm wondering about `IOTrace#showFiberTrace` for other types of events. 
Would it add extra lines about the current `Thread`?

9rb:
This would be a very helpful feature. If a small release can be done that includes this feature, that would be great

Avasil:
> This would be a very helpful feature. If a small release can be done that includes this feature, that would be great

@9rb I plan to finish this PR and do an official release within two weeks but there is a snapshot that you could try: `3.2.2+55-9adf112c-SNAPSHOT` - feedback would be greatly appreciated :)

9rb:
> 
> 
> > This would be a very helpful feature. If a small release can be done that includes this feature, that would be great
> 
> @9rb I plan to finish this PR and do an official release within two weeks but there is a snapshot that you could try: `3.2.2+55-9adf112c-SNAPSHOT` - feedback would be greatly appreciated :)
Thanks. Will do.. @nzpr @tgrospic

PR #1284, opened 2020-10-01
Add left/right builders for Task and Coeval

### LINKED ISSUE #1280: Add left/right builders for Task and Coeval
On companion objects:

```scala
def left[A](a: A): Task[Either[A, Nothing]]

def right[B](b: B): Task[Either[Nothing, B]]
```

I'm just wondering if it's better to use `Nothing` or `def left[A, B](a: A): Task[Either[A, B]]`

adrian-salajan:
Closes #1280


Avasil (approved):
Thank you :)

PR #1293, opened 2020-10-12
removing no-op onComplete() call from onNext()

### LINKED ISSUE #1277: TakeLeftByTimespanObservable should not call onComplete in onNext
[TakeLeftByTimespanObservable calls onComplete](https://github.com/monix/monix/blob/series/3.x/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/TakeLeftByTimespanObservable.scala#L55) in `onNext` but it is a no-op because the `isActive` flag will be always `false` at this moment.

jyoo980:
Should resolve #1277


Avasil (approved):
Thank you :)

PR #1299, opened 2020-10-24
Fix a bug in Local.value

### LINKED ISSUE #1298: Local.value causes infinite loop
When calling `value` on the local within the bound context, the infinite loop occurs. Snippet to reproduce the issue:

```scala
val a, b = Local(1)
 
a.bind(2) {
  b.update(3)
  b.value
}
```

tafit3:
Fixes #1298


Avasil (approved):
Thank you, good catch!

PR #1307, opened 2020-11-06
fromIteratorBuffered

### LINKED ISSUE #1275: Add Observable.fromIteratorChunkedUnsafe
Posting @deusaquilus from gitter:

> I did some profiling of the Quill performance issue. What we really need is a Observeable.fromIteratorChunkedUnsafe that basically assumes the Iterator.next call will (usually) not block for a specified N elements, and that the most efficient course of action is to fetch the next N into a chunk (i.e. buffer), over and over until the end of the result set. This is basically what the JDBC ResultSet does via the FetchSize property (here's an example: https://docs.oracle.com/middleware/1212/toplink/TLJPA/q_jdbc_fetch_size.htm#TLJPA647) which I then write into an iterator.

Not sure if you still need it, but it could be useful to have

Avasil:
Fixes #1275 

I've run a simple benchmark against `fromIterator` + `bufferSliding` and there seems to be a decent gain (around 35%, perhaps more than 3.2.2 because we fixed perf regression on Scala 2.13) but I'm not sure if it is enough to make a difference for your use case @deusaquilus

If it's not, I'm happy to try and help to solve the issue

```
[info] Benchmark                                               (chunkCount)  (chunkSize)   Mode  Cnt    Score   Error  Units
[info] ObservableIteratorBenchmark.bufferTumbling                      1000         1000  thrpt   40   95.434 ± 0.998  ops/s
[info] ObservableIteratorBenchmark.fromIteratorBufferedUnsafe          1000         1000  thrpt   40  130.350 ± 2.840  ops/s
```

### DIFF
--- a/benchmarks/vnext/src/main/scala/monix/benchmarks/ObservableIteratorBenchmark.scala
+++ b/benchmarks/vnext/src/main/scala/monix/benchmarks/ObservableIteratorBenchmark.scala
@@ -63,23 +63,21 @@ class ObservableIteratorBenchmark {
     allElements = chunks.flatten
   }
 
-//  @Benchmark
-//  def monixObservable(): Int = {
-//    val stream = Observable
-//      .fromIteratorUnsafe(allElements.iterator)
-//      .bufferTumbling(chunkSize)
-//      .map(sumIntScala)
-//      .filter(_ % 2 == 0)
-//
-//    sum(stream)
-//  }
+  @Benchmark
+  def bufferTumbling(): Int = {
+    val stream = Observable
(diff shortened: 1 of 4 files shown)


deusaquilus:
@Avasil Actually it might be even more then this number in my use-case because the JDBC ResultSet fetches results in a chunked way itself. That means that by synchronizing Monix Chunks and JDBC chunks, we might be freeing a whole lot of IO cycles downstream. Will let you know how it turns out.

Btw what release can I hope to see this in?

Avasil:
@deusaquilus It's already released in `3.3.0` :) Will be waiting for your report

PR #1313, opened 2020-11-14
remove better-monadic-for

larsrh:
I thought we'd need that for Dotty, but turns out tests pass unchanged anyway, so maybe we don't.


larsrh:
I don't think we'll need this, so I'll close it for now.

Issue #1319, opened 2020-11-18
Add Observable.whileBusyAggregateEvents


lukestephenson:
I've recently migrated a project from akka streams to monix Observable.  That project was making use of [conflateWithSeed](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/conflateWithSeed.html) from akka streams.

This operator will aggregate elements emitted by the Observable if the upstream is faster than the downstream.

While called `conflateWithSeed` in akka streams, I'm suggesting `whileBusyAggregateEvents` in Monix for alignment with existing `whileBusyDropEvents` which is already supported by Monix.

I've taken a stab at this operator and will submit a Draft PR for feedback.

PR #1320, opened 2020-11-18
Add Observable.whileBusyAggregateEvents

lukestephenson:
This is useful for when a downstream consumer of events is slower than the upstream consumer
and events can be aggregated.

Note the implementation of the operator is largely based on https://github.com/monix/monix/blob/5c24f19ea805d5223c9e38961703aa955d691fdd/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/BufferWithSelectorObservable.scala

Related issue: #1319

I'd appreciate any feedback!

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
@@ -23,6 +23,7 @@ import monix.reactive.Observable.Operator
 import monix.reactive.observers.Subscriber
 
 import scala.concurrent.Future
+import scala.util.control.NonFatal
 
 private[reactive] final class WhileBusyAggregateEventsOperator[A, S](seed: A => S, aggregate: (S, A) => S) extends Operator[A, S] {
 
@@ -36,12 +37,18 @@ private[reactive] final class WhileBusyAggregateEventsOperator[A, S](seed: A => 
       private[this] var pendingAck: Boolean = false
       private[this] var downstreamIsDone = false
 
-      def onNext(elem: A): Ack = {
+      override def onNext(elem: A): Ack = {
         upstreamSubscriber.synchronized {
           if (downstreamIsDone) Stop
           else {
             if (!pendingAck) {
-              val downstreamAck = downstream.onNext(seed(elem))
+              val downstreamAck = try {
+                downstream.onNext(seed(elem))
+              } catch {
+                case ex if NonFatal(ex) =>
+                  downstream.onError(ex)
+                  Stop
+              }
               lastAck = downstreamAck
 
               if (downstreamAck == Continue) Continue
@@ -62,11 +69,17 @@ private[reactive] final class WhileBusyAggregateEventsOperator[A, S](seed: A => 
                 Continue
               }
             } else {
-              aggregated = Some(aggregated match {
-                case Some(agg) => aggregate(agg, elem)
-                case None => seed(elem)
-              })
-              Continue
+              try {
+                aggregated = Some(aggregated match {
+                  case Some(agg) => aggregate(agg, elem)
+                  case None => seed(elem)
+                })
+                Continue
+              } catch {
+                case ex if NonFatal(ex) =>
+                  downstream.onError(ex)
+                  Stop
+              }
             }
           }
         }
@@ -94,7 +107,7 @@ private[reactive] final class WhileBusyAggregateEventsOperator[A, S](seed: A => 
         }
       }
 
-      def onError(ex: Throwable): Unit =
+      override def onError(ex: Throwable): Unit =
         upstreamSubscriber.synchronized {
           if (!downstreamIsDone) {
             downstreamIsDone = true
@@ -102,7 +115,7 @@ private[reactive] final class WhileBusyAggregateEventsOperator[A, S](seed: A => 
           }
         }
 
-      def onComplete(): Unit =
+      override def onComplete(): Unit =
         // Can't emit the aggregated element immediately as the previous `onNext` may not yet have been acked.
         upstreamSubscriber.synchronized {
           lastAck.map {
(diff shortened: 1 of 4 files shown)


larsrh:
btw, build fails because of some changes in GitHub Actions. #1321 should fix it.

lukestephenson:
Thanks @larsrh .  Just rebased

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
```diff
@@ -0,0 +1,121 @@
...
+      private[this] var lastAck: Future[Ack] = Continue
+      private[this] var pendingAck: Boolean = false
+      private[this] var downstreamIsDone = false
+
+      def onNext(elem: A): Ack = {
+        upstreamSubscriber.synchronized {
+          if (downstreamIsDone) Stop
+          else {
+            if (!pendingAck) {
+              val downstreamAck = downstream.onNext(seed(elem))
```
Avasil, line 44: We usually assume that user-provided methods could throw an exception and we should protect against that. There is an example in `StateActionObservable`

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
```diff
@@ -0,0 +1,121 @@
...
+              lastAck = downstreamAck
+
+              if (downstreamAck == Continue) Continue
+              else if (downstreamAck == Stop) {
+                downstreamIsDone = true
+                Stop
+              }
+              else {
+                pendingAck = true
+                downstreamAck.map { ack =>
```
Avasil, line 54: I think we should use `onComplete` instead of a `map` to handle `Failure` case

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
```diff
@@ -0,0 +1,121 @@
...
+
+      private def emitAggregated(): Unit = {
+        upstreamSubscriber.synchronized {
+          aggregated match {
+            case Some(agg) =>
+              aggregated = None
+              if (!downstreamIsDone) {
+                lastAck = downstream.onNext(agg)
+
+                lastAck.map { _ =>
```
Avasil, line 83: It seems like we ignore `Stop`. We could also look into using some of `Ack` extensions, e.g. `syncOnContinue` to prevent redundant async boundaries

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
```diff
@@ -0,0 +1,121 @@
...
+                case None => seed(elem)
+              })
+              Continue
+            }
+          }
+        }
+      }
+
+      private def emitAggregated(): Unit = {
+        upstreamSubscriber.synchronized {
```
Avasil, line 76: We have some redundancy in synchronized calls to this method - I think we could remove `synchronized` here and make sure each caller synchronizes which is already done in some places like `onComplete` and `lastAck.map` in this method

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
```diff
@@ -0,0 +1,121 @@
...
+          if (!downstreamIsDone) {
+            downstreamIsDone = true
+            downstream.onError(ex)
+          }
+        }
+
+      def onComplete(): Unit =
+        // Can't emit the aggregated element immediately as the previous `onNext` may not yet have been acked.
+        upstreamSubscriber.synchronized {
+          lastAck.map {
```
Avasil, line 108: We could use `lastAck.syncOnContinue`

Avasil (changes requested):
Thank you, that's a nice operator!

Very well done, scala docs are great!

I've left some general comments. 
I'm also wondering if we could relax synchronization a bit but I'll have to look a bit more carefully into it.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
lukestephenson: This seemed like a good suggestion, but when I tried changing to make use of `syncOnContinue` some of my tests started to break (the last element which should have been emitted here wasn't received).

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
lukestephenson: Have removed.  It was redundant

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
lukestephenson: Addressed and enabled the tests to verify this.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
lukestephenson: Switched this to use `onComplete` (as well as the instance in `emitAggregated`

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/WhileBusyAggregateEventsOperator.scala
lukestephenson: Fixed.

lukestephenson:
@Avasil Thanks for reviewing.  I've addressed the feedback, but note I didn't have success using `syncOnContinue`

Avasil (approved):
Thank you:)

PR #1322, opened 2020-11-19
remove scala.js cross-build

larsrh:
I thought I'd get rid of that Scala.js cross-building business to reduce complexity.


Avasil (approved):
👍

PR #1323, opened 2020-11-20
Scala 3 (Dotty) support

larsrh:
Collaborative PR: @alexandru @Avasil @larsrh 

<hr>

Prequel PRs:

* [x] #1326 
* [x] #1327 
* [x] #1328 
* [x] #1329
* [x] #1331
* [x] #1344


monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/GroupByOperator.scala
```diff
@@ -51,7 +51,7 @@ private[reactive] final class GroupByOperator[A, K](
       private def retryOnNext(elem: A): Future[Ack] =
         onNext(elem)
 
-      @tailrec def onNext(elem: A): Future[Ack] =
```
larsrh, line 54: Are you sure you want to remove all those `@tailrec` annotations? We could rewrite the functions to make them tail recursive.

Avasil: They are still tail recursive, it just can't be verified with the annotation, right?

It would be preferable to keep the annotation but I'm not sure how to do that in cases like:

```scala
  def unsafeSubscribeFn(subscriber: Subscriber[A]): Cancelable = {
    import subscriber.scheduler
    val state = stateRef.get()

    if (state.errorThrown != null) {
      subscriber.onError(state.errorThrown)
      Cancelable.empty
    } else if (state.isDone) {
      Observable
        .now(state.cached)
        .unsafeSubscribeFn(subscriber)
    } else {
      val c = ConnectableSubscriber(subscriber)
      val newState = state.addNewSubscriber(c)

      if (stateRef.compareAndSet(state, newState)) {
        c.pushFirst(state.cached)
        val connecting = c.connect()

        val cancelable = Cancelable { () =>
          removeSubscriber(c)
        }
        connecting.syncOnStopOrFailure(_ => cancelable.cancel())
        cancelable
      } else {
        // retry
        unsafeSubscribeFn(subscriber)
      }
    }
  }
```

larsrh: It's true, but in case something changes in the future, the compiler won't tell us if it's still tail recursive. I refactored this in #1326 by extracting the core loop into an inner method that I've annotated accordingly. However, I don't have any strong opinions on this.

Avasil: Ahh, I was too hasty - I've corrected these method in the latest commit

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/PublishSelectorSuite.scala
```diff
@@ -32,7 +32,7 @@ object PublishSelectorSuite extends BaseTestSuite {
       .publishSelector { source =>
         Observable(source, source, source).merge
       }
-      .sumL
+      .sumL[Long]
```
Avasil, line 35: There is this error (doesn't happen for `Int` ) without type ascription:

```
|ambiguous implicit arguments: both object IntIsIntegral in object Numeric and object DoubleIsFractional in object Numeric match type Numeric[B] of parameter B of method sumL in class Observable
```

didn't have time to investigate why

larsrh:
There a still four compile errors in the doctest `reactiveJVM`:

```
[error] -- Error: /home/lars/proj/monix/monix/monix-reactive/jvm/target/scala-3.0.0-M1/src_managed/test/monix/reactive/ObservableDoctest.scala:446:11 
[error] 446 |    ).merge
[error]     |           ^
[error]     |no implicit argument of type monix.reactive.OverflowStrategy[B] was found for parameter os of method merge in class Observable
```

I find that strange; I thought the type ascription to the implicit argument of `merge` fixed these.

Avasil:
> There a still four compile errors in the doctest `reactiveJVM`:
> 
> ```
> [error] -- Error: /home/lars/proj/monix/monix/monix-reactive/jvm/target/scala-3.0.0-M1/src_managed/test/monix/reactive/ObservableDoctest.scala:446:11 
> [error] 446 |    ).merge
> [error]     |           ^
> [error]     |no implicit argument of type monix.reactive.OverflowStrategy[B] was found for parameter os of method merge in class Observable
> ```
> 
> I find that strange; I thought the type ascription to the implicit argument of `merge` fixed these.

I've opened an issue about it https://github.com/lampepfl/dotty/issues/10497

larsrh:
> I've opened an issue about it

Ah right, I forgot that existed :grin:

Avasil:
@larsrh I have finally spent some time on it and the good news is that almost all tests pass! I think the only issue left is `Observable#merge`...

monix-execution/shared/src/main/scala_2.13+/monix/execution/compat.scala
```diff
@@ -25,14 +25,13 @@ object compat {
...
   private[monix] object internal {
-
     type IterableOnce[+X] = scala.collection.IterableOnce[X]
     def toIterator[X](i: IterableOnce[X]): Iterator[X] = i.iterator
     def hasDefiniteSize[X](i: IterableOnce[X]): Boolean = i.knownSize >= 0
 
     def newBuilder[From, A, C](bf: BuildFrom[From, A, C], from: From): mutable.Builder[A, C] = bf.newBuilder(from)
 
-    @inline def toSeq[A](array: Array[AnyRef]): Seq[A] =
+    def toSeq[A](array: Array[AnyRef]): Seq[A] =
```
Avasil, line 34: I think `inline` made some difference in some `bufferSliding` benchmarks ( https://github.com/monix/monix/blob/series/4.x/benchmarks/shared/src/main/scala/monix/benchmarks/ChunkedEvalFilterMapSumBenchmark.scala#L123 ) but I'm not 100% sure

alexandru: Maybe. But it's risky, and I'd rather get rid of these. We can redo the benchmarks, and then add some explicit comments for that inlining, to ensure that we did our homework. 

What I want to avoid is overuse of the `inline` annotation, which in the beginning I believed that it was a good idea, and added such annotations based on gut feeling.

Avasil: Makes sense, I'll just rerun the benchmark before releasing

alexandru:
I solved `merge` by fixing the tests 🤷‍♂️ as there's nothing we can do for now, the only other options being:

1. to wait for that bug to get fixed in Scala 3, or ...
2. break compatibility

We can break compatibility in 4.0.0, but for now this will have to do.

---

I also upgraded sbt, and cleaned up all warnings, except for those weird ones that aren't reported as actual warnings and that may be the result of some plugin.

Not sure if I want to do anything else before merging.

Avasil (approved):

larsrh:
Please hold back with merging until I get the chance to do one final review pass, okay?

alexandru:
@larsrh sounds good, thanks!

monix-eval/jvm/src/test/scala/monix/eval/TaskLocalJVMSuite.scala
```diff
@@ -26,6 +26,7 @@ import monix.execution.exceptions.DummyException
 import monix.execution.{ExecutionModel, Scheduler}
 import monix.execution.misc.Local
 import monix.execution.schedulers.TracingScheduler
+import monix.execution.cancelableFutureCatsInstances
```
oleg-py, line 29: So, Scala 3 dropped adding object prefixes to implicit scope, so the trick I proposed quite some time ago with making these options work out of the box is broken. I wonder if we can/should adapt something like technique from [here](https://blog.7mind.io/no-more-orphans.html) or simply bite the bullet and drop the "monix-execution doesn't depend on cats directly" policy.

monix-execution/js/src/main/scala_3.0-/monix/execution/atomic/AtomicNumber.scala
```diff
@@ -53,4 +53,4 @@ abstract class AtomicNumber[A] extends Atomic[A] {
   def getAndDecrement(v: Int = 1): A
   /** Subtracts from the atomic number and returns the value before the update. */
   def getAndSubtract(v: A): A
-}
+}
```
oleg-py, line 56: :D

oleg-py (approved):
Looks good to me

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DoOnTerminateOperator.scala
```diff
@@ -71,12 +71,13 @@ private[reactive] final class DoOnTerminateOperator[A](
         }
       }
 
-      private def onFinish(ex: Option[Throwable]): Task[Unit] = {
-        @inline def triggerSignal(): Unit = ex match {
+      private def triggerSignal(ex: Option[Throwable]): Unit =
```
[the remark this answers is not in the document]
alexandru: Because the inner-method captures the current context, it can do an extra-allocation, since it's probably going to get translated into a class initialized with whatever it captures in its constructor. Moving it outside ensures it's not capturing anything, and invoking it will not create an extraneous object.

monix-execution/shared/src/main/scala_3.0-/monix/execution/misc/test/TestInlineMacros.scala
```diff
@@ -22,19 +22,19 @@ import scala.reflect.macros.whitebox
...
-    macro Macros.testInlineSingleArgUnderscore
+  macro Macros.testInlineSingleArgUnderscore
 
   def testInlineMultipleArgsUnderscore(): Either[String, Unit] =
-    macro Macros.testInlineMultipleArgsUnderscore
+  macro Macros.testInlineMultipleArgsUnderscore
 
   def testInlinePatternMatch(): Either[String, Unit] =
-    macro Macros.testInlinePatternMatch
+  macro Macros.testInlinePatternMatch
```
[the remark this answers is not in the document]
alexandru: This was probably a mistake. One of these days we should just enable Scalafmt in the build.

monix-eval/shared/src/main/scala/monix/eval/internal/TaskRunToFutureWithLocal.scala
```diff
@@ -68,7 +68,7 @@ private[eval] object TaskRunToFutureWithLocal {
               bRest.push(bFirst)
             }
             /*_*/
-            bFirst = bindNext /*_*/
+            bFirst = bindNext.asInstanceOf[Bind] /*_*/
```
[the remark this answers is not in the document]
alexandru: Because there we need an `Any => Task[Any]` and it doesn't match, compiler inferring `Nothing`, can't remember where.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DoOnTerminateOperator.scala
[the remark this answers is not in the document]
oleg-py: Compiler has never done that. The desugaring of inner methods is actually pretty much the same as hand-written there. You can see it by doing `-Xprint:lambdalift` (scala2) or `-Xprint:lambdaLift` (scala 3). You can check it easily on Scastie, [e.g. this](https://scastie.scala-lang.org/qwOq9riDRQWSHi6zVJDsRQ)

monix-eval/shared/src/main/scala/monix/eval/internal/TaskRunToFutureWithLocal.scala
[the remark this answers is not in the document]
oleg-py: Oh yeah dotty inferring Nothing/Any changed in way I don't quite understand but was annoyed by so I think I get what you're talking about

monix-eval/jvm/src/test/scala/monix/eval/TaskLocalJVMSuite.scala
alexandru: I don't know what to do next. I don't like the "no more orphans" solution, as it looks like something that can break in the future. I also noticed that Scala 3 started to give advice on imports, not sure how reliable it is, but it looks good. The Scala 3 compiler actually suggested that import in the error report.

Another possibility would be to have another sub-project called `monix-execution-instances` (also a dependency of `monix-catnap`) and have people do ...

```scala
import monix.execution.instances._
```

And hopefully the Scala 3 compiler will suggest it.

alexandru: What I mean is, if we want an optional dependency on Cats, and if the Scala 3 compiler (aka the future) can suggest the import, then maybe such orphans are fine.

oleg-py: Yeah I see what you mean. I'd put those into catnap to not breed submodules.

monix-eval/shared/src/main/scala/monix/eval/internal/TaskRunToFutureWithLocal.scala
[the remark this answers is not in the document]
larsrh: The problem here is that the type of `bindNext` contains a skolem, whose handling in Scala 2 is unsound. Scala 3 tightened this.

monix-execution/jvm/src/main/scala_3.0/monix/execution/atomic/Atomic.scala
```diff
@@ -0,0 +1,220 @@
...
+    *
+    * @param update will be the new value returned by `get()`
+    */
+  def set(update: A): Unit
+
+  /** Alias for [[set]]. Updates the current value.
+    *
+    * @param value will be the new value returned by `get()`
+    */
+  final def update(value: A): Unit = set(value)
```
larsrh, line 40: same

monix-execution/jvm/src/main/scala_3.0/monix/execution/atomic/Atomic.scala
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+  * Base trait of all atomic references, no matter the type.
+  */
+abstract class Atomic[A] extends Serializable {
+  /** Get the current value persisted by this Atomic. */
+  def get(): A
+
+  /** Get the current value persisted by this Atomic, an alias for `get()`. */
+  final def apply(): A = get()
```
larsrh, line 28: Macro in Scala 2, make `inline`?

monix-execution/jvm/src/main/scala_3.0/monix/execution/atomic/Atomic.scala
```diff
@@ -0,0 +1,220 @@
...
+    *
+    * @param value will be the new value returned by `get()`
+    */
+  final def update(value: A): Unit = set(value)
+
+  /** Alias for [[set]]. Updates the current value.
+    *
+    * @param value will be the new value returned by `get()`
+    */
+  final def `:=`(value: A): Unit = set(value)
```
larsrh, line 46: same

monix-execution/jvm/src/main/scala_3.0/monix/execution/atomic/AtomicNumber.scala
```diff
@@ -0,0 +1,56 @@
...
+  *
+  * @tparam A should be something that's Numeric
+  */
+abstract class AtomicNumber[A] extends Atomic[A] {
+  /** Increment with the given integer */
+  def increment(v: Int = 1): Unit
+  /** Adds to the atomic number the given value. */
+  def add(v: A): Unit
+  /** Adds to the atomic number the given value. Alias for `add`. */
+  final def `-=`(value: A): Unit = subtract(value)
```
larsrh, line 31: same

monix-execution/jvm/src/main/scala_3.0/monix/execution/atomic/AtomicNumber.scala
```diff
@@ -0,0 +1,56 @@
...
+  /** Decrements the atomic number with the given integer. */
+  def decrement(v: Int = 1): Unit
+  /** Subtracts from the atomic number the given value. */
+  def subtract(v: A): Unit
+  /** Subtracts from the atomic number the given value. Alias for `subtract`. */
+  final def `+=`(value: A): Unit = add(value)
```
larsrh, line 37: same

monix-execution/shared/src/main/scala/monix/execution/Scheduler.scala
```diff
@@ -409,7 +409,6 @@ object Scheduler extends SchedulerCompanionImpl {
   val TRACING = Features.flag(2)
 
   /** Utilities complementing the `Scheduler` interface. */
-  @deprecated("Extension methods are now implemented on `Scheduler` directly", "3.4.0")
```
larsrh, line 412: Is this accidental or on purpose?

monix-execution/shared/src/main/scala/monix/execution/schedulers/SchedulerService.scala
```diff
@@ -118,7 +118,6 @@ trait SchedulerService extends Scheduler {
 
 object SchedulerService {
   /** Extensions for the [[SchedulerService]] interface. */
-  @deprecated("Extension methods are now implemented on `SchedulerService` directly", "3.4.0")
```
larsrh, line 121: same here

monix-execution/shared/src/main/scala_3.0/monix/execution/misc/Local.scala
```diff
@@ -0,0 +1,312 @@
...
+      case Some(v) => saveKey(key, v)
+    }
+
+  /** If `b` evaluates to `true`, execute a block of code using a current
+    * state of `Local.Context` and restore the current state when complete.
+    */
+  private[monix] def bindCurrentIf[R](b: Boolean)(f: => R)(implicit cb: CanBindLocals[R] = CanBindLocals.synchronous[R]): R =
+    if (!b) f  else {
+      Local.isolate(f)
+    }
```
larsrh, line 138: used to be a macro in Scala 2

monix-java/src/main/scala/monix/java8/eval/package.scala
```diff
@@ -30,7 +30,6 @@ package object eval {
   /**
     * DEPRECATED — switch to Scala 2.12+ and [[monix.eval.Task.from Task.from]].
     */
-  @deprecated("Switch to Scala 2.12+ and Task.from", "3.0.0")
```
larsrh, line 33: Why is this undeprecated?

monix-java/src/main/scala/monix/java8/execution/package.scala
```diff
@@ -31,7 +31,6 @@ package object execution {
     * DEPRECATED — switch to Scala 2.12+ and
     * [[monix.execution.CancelableFuture.fromJavaCompletable CancelableFuture.fromJavaCompletable]].
     */
-  @deprecated("Switch to Scala 2.12+ and CancelableFuture.fromJavaCompletable", "3.0.0")
```
larsrh, line 34: Why is this undeprecated?

monix-java/src/main/scala/monix/java8/execution/package.scala
```diff
@@ -65,7 +64,6 @@ package object execution {
     * DEPRECATED — switch to Scala 2.12+ and
     * [[monix.execution.FutureUtils.toJavaCompletable FutureUtils.toJavaCompletable]].
     */
-  @deprecated("Switch to Scala 2.12+ and FutureUtils.toJavaCompletable", "3.0.0")
```
larsrh, line 68: same

larsrh (commented):
I noted a few issues where there are inconsistencies between Scala 2 and 3 and (possibly?) unrelated changes that I don't understand. But those aren't blockers. Good work, everyone :muscle:

monix-execution/js/src/main/scala_3.0/monix/execution/atomic/AtomicNumber.scala
```diff
@@ -0,0 +1,56 @@
...
+  *
+  * @tparam A should be something that's Numeric
+  */
+abstract class AtomicNumber[A] extends Atomic[A] {
+  /** Increment with the given integer */
+  def increment(v: Int = 1): Unit
+  /** Adds to the atomic number the given value. */
+  def add(v: A): Unit
+  /** Adds to the atomic number the given value. Alias for `add`. */
+  final def `-=`(value: A): Unit = subtract(value)
```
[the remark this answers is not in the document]
alexandru: We have abused macros in `Atomic` and it's time to take a step back. I don't think those aliases being defined as macros have any value. Macros help in those `transform` methods for inlining the given function parameter in what is a hot loop, otherwise the JVM is perfectly capable of inlining such method calls by itself.

That said maybe we should have feature parity and then get rid of those macros in 4.0.0.

So I'm going to revert this change, to have consistency.

monix-execution/shared/src/main/scala/monix/execution/Scheduler.scala
alexandru: It's on purpose. The Scala 3 compiler ends up triggering that warning even when those extensions methods aren't used, don't know why.

Deprecating just the methods is enough IMO. Who knows, we might want to add other extensions in the future.

monix-java/src/main/scala/monix/java8/eval/package.scala
alexandru: For the same reason. Scala 3 compiler triggers that warning even when not used. Don't know if this is a bug or what changed in Scala 3.

I have a zero tolerance for warnings :)

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/DoOnTerminateOperator.scala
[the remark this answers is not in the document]
alexandru: Hmm, my knowledge there may be outdated, goes to show that we should always check and re-check our assumptions.

Will revert the code change.

monix-execution/shared/src/main/scala_3.0/monix/execution/misc/Local.scala
alexandru: Nice catch.

monix-java/src/main/scala/monix/java8/execution/package.scala
alexandru: Same as the others — Scala 3 generates warnings, even if the extensions don't get used, which is why I prefer to have the deprecation warnings only on the methods themselves.

monix-java/src/main/scala/monix/java8/execution/package.scala
alexandru: Same as the others — Scala 3 generates warnings, even if the extensions don't get used, which is why I prefer to have the deprecation warnings only on the methods themselves.

monix-execution/shared/src/main/scala_3.0/monix/execution/misc/Local.scala
alexandru: Solved it.

PR #1326, opened 2020-11-21
Scala 2/3 compatiblity

larsrh:
This PR contains syntactic changes in the `monix-execution` module that could be adopted regardless of Dotty.

### DIFF
--- a/monix-execution/shared/src/main/scala/monix/execution/cancelables/SingleAssignCancelable.scala
+++ b/monix-execution/shared/src/main/scala/monix/execution/cancelables/SingleAssignCancelable.scala
@@ -85,23 +85,26 @@ final class SingleAssignCancelable private (extra: Cancelable) extends Assignabl
     }
   }
 
-  @tailrec
   override def cancel(): Unit = {
-    state.get() match {
-      case IsCanceled | IsEmptyCanceled => ()
-      case IsActive(s) =>
-        state.set(IsCanceled)
-        if (extra != null) extra.cancel()
-        s.cancel()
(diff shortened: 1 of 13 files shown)


monix-execution/shared/src/main/scala/monix/execution/cancelables/SingleAssignCancelable.scala
```diff
@@ -85,23 +85,26 @@ final class SingleAssignCancelable private (extra: Cancelable) extends Assignabl
     }
   }
 
-  @tailrec
   override def cancel(): Unit = {
-    state.get() match {
-      case IsCanceled | IsEmptyCanceled => ()
-      case IsActive(s) =>
-        state.set(IsCanceled)
-        if (extra != null) extra.cancel()
```
larsrh, line 94: Dotty complains about this call, which is working as intended: lampepfl/dotty#9054

monix-execution/shared/src/main/scala/monix/execution/schedulers/TestScheduler.scala
```diff
@@ -155,7 +155,7 @@ final class TestScheduler private (
   }
 
   @tailrec
-  protected override def executeAsync(r: Runnable): Unit = {
+  override def executeAsync(r: Runnable): Unit = {
```
larsrh, line 158: I'm unsure why Scala 2 admitted access to this method from non-subclasses, but given that this is a `TestScheduler` it seems low-risk to just make this public.

Avasil (approved):

PR #1327, opened 2020-11-21
deprecate `Scheduler` extension methods, move them to `Scheduler`

larsrh:
Dotty does not trigger `implicit class`es for extension methods that overload existing methods.

### DIFF
--- a/monix-execution/shared/src/main/scala/monix/execution/schedulers/ExecuteExtensions.scala
+++ b/monix-execution/shared/src/main/scala/monix/execution/schedulers/ExecuteExtensions.scala
@@ -18,66 +18,21 @@
 package monix.execution
 package schedulers
 
-/** Defines extension methods for [[Scheduler]] meant for	
+/** Defines extension methods for [[Scheduler]] meant for
   * executing runnables.
   */
 private[execution] trait ExecuteExtensions extends Any {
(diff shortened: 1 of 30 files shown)


monix-execution/shared/src/main/scala/monix/execution/schedulers/ExecuteExtensions.scala
```diff
@@ -18,66 +18,21 @@
...
-    *
-    * On Scala 2.12 because of the Java 8 SAM types integration,	
-    * this extension macro is replaced with a method that takes	
-    * a plain `Runnable` as parameter.	
-    *
-    * @param cb the callback to execute asynchronously	
-    */
+  @deprecated("Use `execute` directly, since Scala 2.11 support has been dropped", "3.4.0")
   def executeAsync(cb: Runnable): Unit =
     source.execute(cb)
```
larsrh, line 29: I did not create an `executeAsync` method directly on `Scheduler`, because some subclass of `Scheduler` defines a `protected def executeAsync` with the exact same signature. I left this one in place, but deprecated it, and replaced all calls that do not directly call the subclass method.

Avasil (approved):
Good to see `executeAsync` ambiguity improved 👍

PR #1344, opened 2020-12-06
Dotty compatibility

larsrh:
Extracted from #1323 

Should unblock #1338

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -759,10 +759,10 @@ abstract class Observable[+A] extends Serializable { self =>
     * @param sizeOf is the function to compute the weight of each
     *        element in the buffer
     */
-  final def bufferTimedWithPressure(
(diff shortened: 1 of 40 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -759,10 +759,10 @@ abstract class Observable[+A] extends Serializable { self =>
     * @param sizeOf is the function to compute the weight of each
     *        element in the buffer
     */
-  final def bufferTimedWithPressure(
+  final def bufferTimedWithPressure[AA >: A](
```
Avasil, line 762: Why dotty requires `AA >: A` ?

Avasil (approved):

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
larsrh: Because `A` is in contravariant position here, and Scala 2 didn't recognize that.

PR #1345, opened 2020-12-06
bump ScalaTest version

larsrh:
Not sure why this didn't get picked up by scala-steward.


larsrh:
I know now: it's not actually referenced anywhere. Will send a follow-up PR that removes the dependency.

Issue #1292, opened 2020-10-09
Corrupted concatenation after Observable.publishSelector


squadgazzz:
Here's a sample code.
```
  val obs = Observable.fromIterable(1 to 2)
  val result = obs.publishSelector { o =>
    o ++ Observable.fromIterable(11 to 12)
  }
  result.toListL.executeAsync.runToFuture.foreach(l => println(l.mkString("[", ",", "]")))
  Thread.sleep(5.seconds.toMillis)
```
result: `[1,2,11,12]`
```
  val obs = Observable.fromIterable(1 to 2)
  val result = obs.publishSelector { o =>
    o ++ Observable.fromIterable(11 to 12)
  } ++ Observable.fromIterable(101 to 102)
  result.toListL.executeAsync.runToFuture.foreach(l => println(l.mkString("[", ",", "]")))
  Thread.sleep(5.seconds.toMillis)
 ```
 result: `[11,12,101,102]`

Where did `obs`'s elements go?

alexandru:
As mentioned on the Gitter chat, I smell a subscription timing issue. 

From the API's usage, this looks normal, I think the expectation is correct, that stream should indeed start with `1, 2`.

This needs to be investigated.

PR #1355, opened 2020-12-29
Publish selector subscription issue (#1292)

paualarco:
This is an attempt to show/help to further investigate the issue with publish selector raised on the ticket #1292.

As @alexandru mentioned, this is a timing on subscription, which only happens when a publish selector along with two `ChainedObservable` operators, one that applies to the hot one (within the brackets) and used after.

As you can appreciate in the ticket the issue was reported with `++` but for simplicity, I have added a test that uses `takeLast`, which is also a chained one.

The issue can be fixed by avoiding the trampolined execution on the subscription of the `ChainedObservable`, thus being representing a synchronous subscription flow. Probably there are important reasons why done in that way, so I am unsure if that change would impact negatively on other functionalities. So would be good if someone know of a better alternative to solve it :)

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/observables/ChainedObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/observables/ChainedObservable.scala
@@ -66,14 +66,14 @@ object ChainedObservable {
     // Subscription can be synchronous, which can trigger really
     // weird ordering issues, therefore we need a light async boundary
     // to delay it and force ordering; plus it protects from stack overflows
-    out.scheduler.executeTrampolined { () =>
       source match {
         case _: ChainedObservable[_] =>
-          source.asInstanceOf[ChainedObservable[A]].unsafeSubscribeFn(conn, out)
+          out.scheduler.executeTrampolined { () =>
+            source.asInstanceOf[ChainedObservable[A]].unsafeSubscribeFn(conn, out)
+          }
         case _ =>
           conn := source.unsafeSubscribeFn(out)
           ()
-      }
     }
   }
 }
--- a/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/PublishSelectorSuite.scala
+++ b/monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/PublishSelectorSuite.scala
@@ -22,6 +22,7 @@ import monix.execution.atomic.Atomic
 import monix.reactive.{BaseTestSuite, Observable}
 
 import scala.util.Success
+import scala.concurrent.duration._
 
 object PublishSelectorSuite extends BaseTestSuite {
   test("publishSelector sanity test") { implicit s =>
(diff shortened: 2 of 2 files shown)


Avasil:
Thanks for tackling this!

It seems like [upstream is subscribed](https://github.com/monix/monix/blob/series/3.x/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/PipeThroughSelectorObservable.scala#L55) before we subscribe to the inner `Observable`.

I'm not quite sure why though. 

I'd expect

```scala
s.executeTrampolined(() => subscribeA())
subscribeB()
```

to always call `subscribeA` first but maybe I'm rusty or missing something else.

> Probably there are important reasons why done in that way, so I am unsure if that change would impact negatively on other functionalities. So would be good if someone know of a better alternative to solve it :)

I feel like this light async boundary is not necessary for `PipeThroughSelectorObservable` but we should confirm if your solution doesn't break other use cases. I'd be mainly concerned about stack safety.

Although, we should probably start by figuring out what exactly is going on. :D

paualarco:
It seems like on one hand, we need the async boundary for ensuring ordering subscription, removing it only breaks `monix.reactive.internal.operators.ConcatCancellationSuite` which is a test to ensure that cancelation for chained observables is performed correctly... 

 A question to clarify myself: Is there any difference (in terms of external functionality for the user) when using 
`pipeThrougSelector` with `Pipe.publish` and `Pipe.replay`? 
To me, difference is clear when the subjects are used directly, but I am unsure of the difference in the behaviour when they are used with the wrapper method.  

I am asking this because this issue does not happen when using `Replay` with `pipeThrough`, and that's because there is no weird subscription behaviour happening in there, and maybe we could try replacing it by `publishSelector` by `replySelector` or something like that.

Avasil:
>A question to clarify myself: Is there any difference (in terms of external functionality for the user) when using
pipeThrougSelector with Pipe.publish and Pipe.replay?
To me, difference is clear when the subjects are used directly, but I am unsure of the difference in the behaviour when they are used with the wrapper method.

The difference is the same, this example might clear it:

```scala
  val source = Observable.intervalAtFixedRate(300.millis)

  val publishSelector = source.publishSelector { hot =>
    Observable.combineLatest2(hot, hot.delayExecution(1.second))
  }

  // List((4,4), (5,4), (5,5), (6,5), (6,6), (7,6), (7,7), (8,7), (8,8), (9,8))
  publishSelector.take(10).toListL.map(println).runSyncUnsafe()
```

By the time the second `Observable` starts, other elements are consumed and discarded by the first one and the latest is `4`.

```scala
  val replaySelector = source.pipeThroughSelector(Pipe.replay[Long], { hot: Observable[Long] =>
    Observable.combineLatest2(hot, hot.delayExecution(1.second))
  })

  // List((3,0), (3,1), (3,2), (3,3), (4,3), (4,4), (5,4), (5,5), (6,5), (6,6))
  replaySelector.take(10).toListL.map(println).runSyncUnsafe()
```

The first `Observable` also consumes and discards elements from the source. However, they are cached so when a delayed stream starts, it starts from the first element.

> I am asking this because this issue does not happen when using Replay with pipeThrough, and that's because there is no weird subscription behaviour happening in there, and maybe we could try replacing it by publishSelector by replySelector or something like that.

It doesn't happen with `replay` because elements that are missed due to late subscription are cached and emitted to all new subscribers.

On the surface, it matches the expected behavior but it is less efficient. 
All elements that are emitted are cached, even if the source is consuming them and we know that it is the only source.

paualarco:
@Avasil thanks for that example, it was useful! Will look at it again to see if I find out.

Issue #1360, opened 2021-01-15
Add Observable.paginate & Observable.paginateEval


Avasil:
```scala
def paginate[S, A](seed: => S)(f: S => (A, Option[S])): Observable[A]

def paginateEval[S, A](seed: => S)(f: S => Task[(A, Option[S])]): Observable[A]
```

Implementation is similar to [unfold](https://github.com/monix/monix/blob/series/3.x/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/UnfoldEvalObservable.scala) but we will emit one more `A` before completing on `None`

See [fs2 unfoldLoopEval](https://github.com/typelevel/fs2/blob/main/core/shared/src/main/scala/fs2/Stream.scala#L3610) and [ZIO's paginateM](https://github.com/zio/zio/blob/master/streams/shared/src/main/scala/zio/stream/ZStream.scala#L3800) for usage examples

Wosin:
Happy to tackle that.

Avasil:
@Wosin That's awesome - let me know if you have any questions or need some direction

Wosin:
Hey, if I understand correctly that's just a `unfold` which in case of `None` returns the `A` again, right ?

Avasil:
That's correct.

This is the result I'd expect:

```scala
Observable.unfold(0)(i => if (i < 10) Some((i, i + 1)) else None).toListL
// result: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

Observable.paginate(0)(i => if (i < 10) Some((i, i + 1)) else None).toListL
// result: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```

Wosin:
Can be closed I guess :)

PR #1381, opened 2021-02-25
Implement pagination for Observable.

Wosin:
This PR implements pagination as requested in #1360.

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -5632,9 +5632,23 @@ object Observable extends ObservableDeprecatedBuilders {
   def unfoldEval[S, A](seed: => S)(f: S => Task[Option[(A, S)]]): Observable[A] =
     new UnfoldEvalObservable(seed, f)
 
+  /** Similar to [[unfold]], but allows to take emission one step further.
+    * @example {{{
+    *  Observable.paginate(0)(i => if (i < 10) (i, Some(i + 1)) else (i, None)).toListL
+    *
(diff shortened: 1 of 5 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -5632,6 +5632,12 @@ object Observable extends ObservableDeprecatedBuilders {
   def unfoldEval[S, A](seed: => S)(f: S => Task[Option[(A, S)]]): Observable[A] =
     new UnfoldEvalObservable(seed, f)
 
+  def paginate[S, A](seed: => S)(f: S => (A, Option[S])): Observable[A] =
```
Avasil, line 5635: Could you please add a scaladoc similar to `unfold` and `unfoldEval`?

Avasil (approved):
The implementation looks great, thank you!

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
Wosin: Sure! Wanted to get the opinion on the implementation first. I will add this today!

Wosin: I've added doc, that is written as comparison to `unfold` and `unfoldEval`. I thought that would be easiest to understand, but I've also noticed that most of the functions have separate docs that do not depend on other functions so let me know if You would want me to change  :)

Avasil: I think it's fine in this case since those are very similar methods. 
It might be cool to add `@see` in `unfold` / `unfoldEval` that would mention `paginate` / `paginateEval`

PR #1390, opened 2021-03-08
Bump cats-effect to 3.0.0-RC2; port catnap+tests to CE3

oleg-py:
Here's catnap upgrade I've been slowly doing over the past few weeks. Putting it out if you guys want to pick it up faster than I can do it (we should probably start a branch and make it a part of 4.x going forward since bincompat is going to be broken anyway).

A lot of our tests were relying on cats IO having a model where it executes stuff on current thread as much as possible, and that came to bite us of course. We should really consider writing tests using high-level pure-fp stuff, with cats-effect-testing or something.

The big change is that Async/Concurrent split is now very different and most of our stuff can built only using Async. No need for `OrElse` and such. And ContextShift and Timer are kinda gone. I'll write some more detailed comments in the upcoming days, feel free to ask if anything looks suspicious for now.


Avasil:
> I think we should switch to the 4.x branch, although there might be a need to cherry-pick PRs from 3.x first

Yeah, there wasn't any action there so I just ignored it over the last half a year

monix-catnap/jvm/src/test/scala/monix/catnap/SemaphoreJVMSuite.scala
```diff
@@ -45,41 +46,11 @@ abstract class BaseSemaphoreJVMTests(parallelism: Int) extends TestSuite[Schedul
...
-  implicit def contextShift(implicit ec: Scheduler): ContextShift[IO] =
-    SchedulerEffect.contextShift[IO](ec)(IO.ioEffect)
-  implicit def timer(implicit ec: Scheduler): Timer[IO] =
-    SchedulerEffect.timerLiftIO[IO](ec)(IO.ioEffect)
-
   // ----------------------------------------------------------------------------
   val iterations = if (isCI) 1000 else 10000
   val timeout = if (isCI) 30.seconds else 10.seconds
 
-  test("Semaphore (concurrent) — issue #380: — producer keeps its thread, consumer stays forked") { implicit ec =>
```
oleg-py, line 57: This is an example of what we can't test for with IO anymore. IO is now like if you always did `executeAsync` on Task, will always shift to the pool that backs provided runtime. Not sure if it's required by law tho as I haven't looked into new laws.

monix-catnap/shared/src/main/scala/monix/catnap/CircuitBreaker.scala
```diff
@@ -622,7 +632,7 @@ object CircuitBreaker extends CircuitBreakerDocs {
     exponentialBackoffFactor: Double = 1.0,
     maxResetTimeout: Duration = Duration.Inf,
     padding: PaddingStrategy = NoPadding
-  )(implicit F: Sync[F], clock: Clock[F]): F[CircuitBreaker[F]] = {
+  )(implicit F: Sync[F], F2: MonadCancelThrow[F], clock: Clock[F]): F[CircuitBreaker[F]] = {
```
oleg-py, line 635: I need to revisit this - since I started porting on RC1, Sync now extends MonadCancelThrow.

monix-catnap/shared/src/main/scala/monix/catnap/MVar.scala
```diff
@@ -61,10 +60,10 @@ import monix.execution.internal.GenericVar.Id
   * from Haskell.
   */
 
-final class MVar[F[_], A] private (underlying: MVar.Impl[F, A]) extends CatsMVar[F, A] {
```
oleg-py, line 64: Cats-effect no longer has MVar to extend

monix-catnap/shared/src/main/scala/monix/catnap/Semaphore.scala
```diff
@@ -146,53 +129,7 @@ final class Semaphore[F[_]] private (provisioned: Long, ps: PaddingStrategy)(
...
-    *
-    * The returned task also takes care of resource handling,
-    * releasing its permit after being complete.
-    *
-    * @param fa is an effect to execute once the permit has been
-    *        acquired; regardless of its result, the permit is
-    *        released to the pool afterwards
-    */
-  def withPermit[A](fa: F[A]): F[A] =
-    withPermitN(1)(fa)
```
oleg-py, line 174: `withPermit` is gone from CE3 semaphore, _and_ RC2 added a `mapK` requirement. If we want to extend CE3 semaphore, we need to provide transformation for any `MonadCancel[F, _]`, but that's not enough to do `permit.surround` (Resource.use needs `MonadCancel[F, Throwable]` specifically). We can still have the syntax  with a slightly different signature than before (requiring implicit `MonadCancel[F, Throwable]`), I'm just not sure if we want to keep it.

monix-catnap/shared/src/main/scala/monix/catnap/Semaphore.scala
```diff
@@ -134,10 +117,10 @@ final class Semaphore[F[_]] private (provisioned: Long, ps: PaddingStrategy)(
     *
     * @see [[tryAcquireN]] for the version that can acquire multiple permits
     * @see [[acquire]] for the version that can wait for acquisition
-    * @see [[withPermit]] the preferred way to acquire and release
+    * @see [[permit]] the preferred way to acquire and release
     */
   @UnsafeProtocol
-  override def tryAcquire: F[Boolean] = tryAcquireN(1)
+  def tryAcquire: F[Boolean]
```
oleg-py, line 123: I kept signatures like that to override docs. I think monix version is better, for the most part

monix-catnap/shared/src/main/scala/monix/catnap/internal/AsyncUtils.scala
```diff
@@ -17,28 +17,20 @@
...
+    * The `cancelable` builder from cats-effect 2
     */
   def cancelable[F[_], A](k: (Either[Throwable, A] => Unit) => F[Unit])(implicit F: Async[F]): F[A] =
-    F.asyncF { cb =>
-      val p = Promise[A]()
-      val awaitPut = Callback.fromPromise(p)
-      val future = p.future
-      val futureF = FutureLift.scalaToAsync(F.pure(future)).attempt.map(cb)
-      val cancel = k(awaitPut)
+    F.async[A] { cb => F.pure(Some(k(cb))) }
```
oleg-py, line 27: `F.cancelable` and `F.asyncF` have been rolled into one `F.async`. Old `F.async` became `F.async_`. I made the helpers to avoid larger source changes.

monix-catnap/shared/src/main/scala/monix/catnap/package.scala
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package monix
+
+package object catnap {
+  type CancelToken[F[_]] = F[Unit]
```
oleg-py, line 21: Cats has dropped this type alias. Not sure if we want to keep it, but I did for now.

monix-catnap/shared/src/test/scala/monix/catnap/CancelableFSuite.scala
```diff
@@ -17,14 +17,14 @@
...
 
 object CancelableFSuite extends SimpleTestSuite {
   test("apply") {
     var effect = 0
-    val task = IO { effect += 1 }
-    val ref = CancelableF[IO](task)
+    val task = SyncIO {effect += 1 }
+    val ref = CancelableF[SyncIO](task)
 
     val cf = ref.unsafeRunSync()
```
oleg-py, line 29: I've changed all cancelable tests to use SyncIO b/c unsafeRunSync simply doesn't exist on IO when built for JS.

monix-catnap/shared/src/main/scala/monix/catnap/MVar.scala
```diff
@@ -200,175 +199,113 @@ object MVar {
...
       _ => shift
     }
 
     private[this] val bindForkA: (Any => F[Any]) = {
-      val shift = cs.shift
       x => F.map(shift)(_ => x)
     }
   }
-
-  private final class AsyncImpl[F[_], A](initial: Option[A], ps: PaddingStrategy)(
```
oleg-py, line 319: Since CE3 is dropping whole "uncancelable Async" thing, and our Task is already always old Concurrent, new Async, I think we should follow suit. The original rationale was to allow users to not require `Concurrent` if they just need a tiny bit of semantic blocking, but since Concurrent was a control flow thing, it was required quite pervasively anyway.

monix-catnap/shared/src/test/scala/monix/catnap/TestUtils.scala
```diff
@@ -0,0 +1,53 @@
...
+
+      override def nowMillis(): Long = sc.clockRealTime(TimeUnit.MILLISECONDS)
+      override def monotonicNanos(): Long = sc.clockMonotonic(TimeUnit.NANOSECONDS)
+    }, () => {})
+
+  implicit class IOExtensions[A](io: IO[A]) {
+    def unsafeRunCancelableTick(tick: FiniteDuration = Duration.Zero)(cb: Either[Throwable, A] => Unit)(implicit tsc: TestScheduler): CancelToken[IO] =
+      io.attempt.map(cb).start.unsafeRunSyncTick(tick).cancel
+
+    def unsafeRunSyncTick(tick: FiniteDuration = Duration.Zero)(implicit tsc: TestScheduler) = {
```
oleg-py, line 43: Since cats IO always shifts first, even doing `unsafeRunSync` would require the EC to do some scheduling. These helpers allowed me to migrate test code quite easily, esp. cases where we did `assert(io.unsafeRunSync() == 42)` assuming it'll just run on current thread if there are no async boundaries.

monix-catnap/jvm/src/test/scala/monix/catnap/MVarJVMSuite.scala
```diff
@@ -175,7 +134,7 @@ abstract class BaseMVarJVMSuite(parallelism: Int) extends TestSuite[SchedulerSer
     }
   }
 
-  test("MVar (concurrent) — issue #380: with foreverM; without latch") { implicit ec =>
+  test("MVar  — issue #380: with foreverM; without latch") { implicit ec =>
```
oleg-py, line 137: uh, I derped and left double spaces as I was removing the concurrent/async distinction 😅

monix-catnap/jvm/src/test/scala/monix/catnap/SemaphoreJVMSuite.scala
```diff
@@ -19,12 +19,13 @@ package monix.catnap
 
 import java.util.concurrent.atomic.AtomicBoolean
 
-import cats.effect.concurrent.Deferred
-import cats.effect.{ContextShift, IO, Timer}
+import cats.effect.kernel.Deferred
+import cats.effect.{IO}
```
paualarco, line 23: No need for brackets here :)

monix-catnap/shared/src/main/scala/monix/catnap/FutureLift.scala
```diff
@@ -98,32 +100,14 @@ object FutureLift extends internal.FutureLiftForPlatform {
   /**
     * Utility for converting [[scala.concurrent.Future Future]] values into
     * data types that implement
-    * [[https://typelevel.org/cats-effect/typeclasses/async.html cats.effect.Async]].
+    * [[https://typelevel.org/cats-effect/typeclasses/concurrent.html cats.effect.Async]].
```
paualarco, line 103: Wasn't that correct before?

alexandru:
I changed the base to `series/4.x`.

monix-catnap/shared/src/main/scala/monix/catnap/SchedulerEffect.scala
```diff
@@ -1,136 +0,0 @@
...
-
-import cats.implicits._
-import cats.effect._
-import monix.execution.Scheduler
-import monix.execution.internal.AttemptCallback.RunnableTick
-
-import scala.concurrent.ExecutionContext
-import scala.concurrent.duration.{FiniteDuration, TimeUnit}
-
-object SchedulerEffect {
```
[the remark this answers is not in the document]
alexandru: I don't really like the `Dispatcher` idea but who knows 🤷‍♂️ 

Either way, it's OK to remove it. We can probably provide advice on how to build an `IORuntime` from a `Scheduler` in the documentation. Or create another project :)

PR #1396, opened 2021-03-23
Add throttleLatest to Observable

### LINKED ISSUE #1385: `Observable.throttleLatest` support
JavaRx `Observable` has an additional throttling method [throttleLatest](http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#throttleLatest-long-java.util.concurrent.TimeUnit-).
It is pretty useful and could be a nice addition to Monix's `Observable`.

Wosin:
This closes #1385. 

This PR introduces `throttleLatest` as defined in [here](http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#throttleLatest-long-java.util.concurrent.TimeUnit-boolean-). PR introduces some tests to verify the correctness of the behaviour implemented.

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -3351,6 +3351,39 @@ abstract class Observable[+A] extends Serializable { self =>
   final def throttleLast(period: FiniteDuration): Observable[A] =
     sample(period)
 
+  /** Emit first element emitted by the source and then
+    * emit the most recent items emitted by the source within
+    * periodic time intervals.
+    *
+    ** 0 200
+    * 1 400
+    * 2 600
+    * 3 800
+    * 4 1000
+    * 5 1200
+    * 6 1400
+    * 7 1600
+    * 8 1800
+    * 9 2000
+    * 10 2200
+    * Usage:
+    *
+    * {{{
+    *   import scala.concurrent.duration._
+    *
+    *   // emits 0 after 200 ms and then 4,9 in 1 sec intervals and 10 after the observable completes
+    *   Observable.fromIterable(0 to 10)
+    *     // without delay, it would return only 0, 10
+    *     .delayOnNext(200.millis)
+    *     .throttleLatest(1.second, true)
+    * }}}
+    *
+    * @param period duration of windows within which the last item
+    *        emitted by the source Observable will be emitted
+    * @param period duration of windows within which the last item
+    *        emitted by the source Observable will be emitted
+    *
+    */
   final def throttleLatest(period: FiniteDuration, emitLast: Boolean): Observable[A] =
     new ThrottleLatestObservable[A](self, period, emitLast)
   /** Emit the most recent items emitted by the source within
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -3354,18 +3354,6 @@ abstract class Observable[+A] extends Serializable { self =>
   /** Emit first element emitted by the source and then
     * emit the most recent items emitted by the source within
     * periodic time intervals.
-    *
-    ** 0 200
-    * 1 400
-    * 2 600
-    * 3 800
-    * 4 1000
-    * 5 1200
-    * 6 1400
-    * 7 1600
-    * 8 1800
-    * 9 2000
-    * 10 2200
     * Usage:
     *
     * {{{
@@ -3382,7 +3370,6 @@ abstract class Observable[+A] extends Serializable { self =>
     *        emitted by the source Observable will be emitted
     * @param period duration of windows within which the last item
     *        emitted by the source Observable will be emitted
-    *
     */
   final def throttleLatest(period: FiniteDuration, emitLast: Boolean): Observable[A] =
     new ThrottleLatestObservable[A](self, period, emitLast)
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -3368,8 +3368,8 @@ abstract class Observable[+A] extends Serializable { self =>
     *
     * @param period duration of windows within which the last item
     *        emitted by the source Observable will be emitted
-    * @param period duration of windows within which the last item
-    *        emitted by the source Observable will be emitted
+    * @param emitLast if true last element will be emitted when source completes
+    *                 no matter if interval has passed or not
     */
   final def throttleLatest(period: FiniteDuration, emitLast: Boolean): Observable[A] =
     new ThrottleLatestObservable[A](self, period, emitLast)
--- a/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
@@ -3351,6 +3351,8 @@ abstract class Observable[+A] extends Serializable { self =>
   final def throttleLast(period: FiniteDuration): Observable[A] =
     sample(period)
 
+  final def throttleLatest(period: FiniteDuration, emitLast: Boolean): Observable[A] =
+    new ThrottleLatestObservable[A](self, period, emitLast)
   /** Emit the most recent items emitted by the source within
     * periodic time intervals.
     *
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ThrottleLatestObservable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ThrottleLatestObservable.scala
@@ -45,10 +45,9 @@ private[reactive] final class ThrottleLatestObservable[A](
       private[this] val durationMilis = duration.toMillis
       private[this] var isDone = false
       private[this] var lastEvent: A = _
-      @volatile private[this] var hasValue = false
-      @volatile private[this] var shouldEmitNext = true
-      private[this] var lastAck: Future[Ack] = _
-
+      private[this] var hasValue = false
+      private[this] var shouldEmitNext = true
+      
       def scheduleNext(delayMillis: Long): Unit = {
         // No need to synchronize this assignment, since we have a
         // happens-before relationship between scheduleOnce invocations.
@@ -60,11 +59,11 @@ private[reactive] final class ThrottleLatestObservable[A](
         if (!isDone) {
           if (hasValue) {
             hasValue = false
-            val now = scheduler.clockMonotonic(TimeUnit.SECONDS)
+            val now = scheduler.clockMonotonic(TimeUnit.MILLISECONDS)
 
             out.onNext(lastEvent).syncFlatMap {
               case Continue =>
-                val elapsed = scheduler.clockMonotonic(TimeUnit.SECONDS) - now
+                val elapsed = scheduler.clockMonotonic(TimeUnit.MILLISECONDS) - now
                 val delay =
                   if (durationMilis > elapsed)
                     durationMilis - elapsed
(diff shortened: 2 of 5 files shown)


monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ThrottleLatestObservable.scala
```diff
@@ -0,0 +1,134 @@
...
+          if (hasValue) {
+            hasValue = false
+            val now = scheduler.clockMonotonic(TimeUnit.SECONDS)
+
+            out.onNext(lastEvent).syncFlatMap {
+              case Continue =>
+                val elapsed = scheduler.clockMonotonic(TimeUnit.SECONDS) - now
+                val delay =
+                  if (durationMilis > elapsed)
+                    durationMilis - elapsed
```
oleg-py, line 70: Are you subtracting seconds from millis there?

Seconds is generally a bad resolution to use for scheduling tasks, I suggest using ms everywhere.

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ThrottleLatestObservable.scala
```diff
@@ -0,0 +1,134 @@
...
+      }
+
+      override def onComplete(): Unit = self.synchronized {
+        if (!isDone) {
+          if (emitLast && hasValue) {
+            out.onNext(lastEvent).syncTryFlatten.syncOnContinue {
+              isDone = true
+              out.onComplete()
+              task.cancel()
+            }
```
oleg-py, line 121: I wonder why `hasValue` and `shouldEmitNext` are `@volatile`?

This seems to be the only place where something can be written outside of `self.synchronized` block (if Ack fails to be synchronous) and it might theoretically be possible that the last event gets emitted twice if that block runs in parallel with the one in `def run` since the `isDone = true` here might not be propagated to the thread doing `def run`. Otherwise, JSR-133 should guarantee memory barriers when entering and exiting synchronized blocks.

The simple fix I can think of is to just do
```scala
if (!isDone) {
  isDone = true
  ... rest of the branching code ...
}
```
to both tell the `def run` to stop itself and keep the write in synchronized block and not in a Future continuation so everything can see it. 

I'm going to tag @alexandru so he can tell me that I'm wrong there.

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/ThrottleLatestSuite.scala
```diff
@@ -0,0 +1,85 @@
...
+  def createObservable(sourceCount: Int) = Some {
+    if (sourceCount == 1) {
+      val o = Observable.now(100L).delayExecution(500.millis).throttleLatest(1.second, true)
+      Sample(o, 1, 100, 500.millis, 1.second)
+    } else {
+
+      val o = Observable
+        .intervalAtFixedRate(2.second)
+        .take(sourceCount.toLong)
+        .throttleLatest(1.second, true)
```
oleg-py, line 39: If the rate of output is at 2 seconds and you're throttling by 1 second after first element (as per RxJava doc you linked), isn't that a noop?

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/ThrottleLatestSuite.scala
```diff
@@ -0,0 +1,85 @@
...
+import scala.concurrent.duration._
+
+object ThrottleLatestSuite extends BaseOperatorSuite {
+  def sum(sourceCount: Int) = {
+    sourceCount * (sourceCount - 1) / 2
+  }
+  def createObservable(sourceCount: Int) = Some {
+    if (sourceCount == 1) {
+      val o = Observable.now(100L).delayExecution(500.millis).throttleLatest(1.second, true)
+      Sample(o, 1, 100, 500.millis, 1.second)
```
oleg-py, line 33: Please test with something with more than one element in the output :D That might prove that emitLast handling is correct but does very little to actually prove that throttling is working as expected.

oleg-py (changes requested):
There are few suspicious spots. I'd also like to see test that shows exactly which elements have been _retained_ when using a producer that's significantly more frequent than the throttle interval.

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/ThrottleLatestSuite.scala
DomWos: I don't think we are tsting with one element on the output here :) I think `createObservable` gets called from outside and we pass the `sourceCount` from`BaseOperatorSuite`, so that is randomized. 
I've added this `sourceCount ==1` check because I've noticed that in several suites, but now that I've started chcecking I think it's not needed :)

oleg-py: Right. Disregard that line comment but please adjust your additional tests to have more than 1-2 elements :)

Wosin: Sure, sounds reasonable I will add few more tests.

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/ThrottleLatestSuite.scala
Wosin: Yeah, probably true. I will think about something, but as far as I know we can't control the exact elements that are going to be throttled in this test, same as in `SampleOnce`.

oleg-py: You should be able to do something like
```scala
Observable(1).delayOnNext(200.millis) ++ Observable(2).delayOnNext(100.millis) ++ ...
```
to create a predictable chain.

oleg-py: I'm also concerned about `Sample` two lines below, as that is literally saying it's not dropping anything from the input (2nd parameter of sample is expected output size). The purpose of throttling is to drop some so we should check this if possible. I think ThrottleFirstSuite has an example where it drops half.

Wosin: Yeah, that is correct. I was mostly basing on `SampleOnceSuite` here. No worries though I will fix it now :)

Wosin: > 

You should be able to do something like

Observable(1).delayOnNext(200.millis) ++ Observable(2).delayOnNext(100.millis) ++ ...

to create a predictable chain.

Not sure what's the difference between that and just doing the `intervalAtFixedRate` to be honest :)

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ThrottleLatestObservable.scala
Avasil: I think you're right @oleg-py and I think we could even do:

```scala
if (!isDone) { 
  isDone = true 
  task.cancel()
  ... rest of the branching code ...
}
```

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ThrottleLatestObservable.scala
```diff
@@ -0,0 +1,134 @@
...
+    mainTask := source.unsafeSubscribeFn(new Subscriber[A] with Runnable {
+      self =>
+      implicit val scheduler = out.scheduler
+
+      private[this] val durationMilis = duration.toMillis
+      private[this] var isDone = false
+      private[this] var lastEvent: A = _
+      @volatile private[this] var hasValue = false
+      @volatile private[this] var shouldEmitNext = true
+      private[this] var lastAck: Future[Ack] = _
```
Avasil, line 50: seems to be unused

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ThrottleLatestObservable.scala
```diff
@@ -0,0 +1,134 @@
...
+          isDone = true
+          out.onError(ex)
+          task.cancel()
+        }
+      }
+
+      override def onComplete(): Unit = self.synchronized {
+        if (!isDone) {
+          if (emitLast && hasValue) {
+            out.onNext(lastEvent).syncTryFlatten.syncOnContinue {
```
Avasil, line 117: Are we guaranteed that the last `onNext` has returned?

It seems like `shouldEmitNext` handles back-pressure for `onNext` but I'm not so sure for `onComplete`

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/ThrottleLatestSuite.scala
Avasil: I think we could still test cases from `SampleOnceSuite`

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/ThrottleLatestSuite.scala
Avasil: I think something like that can be useful if we would like to have a specific test case in a test outside `BaseOperatorSuite`, e.g. [ReactiveX example](http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#throttleLatest-long-java.util.concurrent.TimeUnit-)

Avasil (commented):
Thanks a lot for the PR @Wosin and sorry for the delay with review

monix-reactive/shared/src/main/scala/monix/reactive/internal/operators/ThrottleLatestObservable.scala
Wosin: Yeah, will remove that.

monix-reactive/shared/src/test/scala/monix/reactive/internal/operators/ThrottleLatestSuite.scala
Wosin: Sure, gonna take a look at them.

monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala
```diff
@@ -3351,6 +3351,28 @@ abstract class Observable[+A] extends Serializable { self =>
...
+    * }}}
+    *
+    * @param period duration of windows within which the last item
+    *        emitted by the source Observable will be emitted
+    * @param emitLast if true last element will be emitted when source completes
+    *                 no matter if interval has passed or not
+    */
+  final def throttleLatest(period: FiniteDuration, emitLast: Boolean): Observable[A] =
+    new ThrottleLatestObservable[A](self, period, emitLast)
   /** Emit the most recent items emitted by the source within
```
paualarco, line 3376: @Wosin shouldn't be a line break between the two methods? :P

Wosin: Which methods do You mean  ?:P

paualarco: I mean line 3375 and 3376, the end of `throttleLatest` and the _scaladoc_ for `sample`.

Wosin: Oh, yeah You are right ofc. I thought scalafmt should get that.

Avasil (approved):
Thanks, LGTM 👍

oleg-py (approved):
Yeah, looks good 👍

PR #1416, opened 2021-05-06
[TEST] Series/4.x (do not merge!)

alexandru:
Creating a PR, just so we can see that nothing was screwed in my merge.


Avasil:
@alexandru fyi there's a PR with quite a lot of changes regarding Scala 3 upgrade: https://github.com/monix/monix/pull/1323
It's probably good to merge that one (there's only one unsolved issue) before diverging too much in `4.x` branch

PR #1440, opened 2021-05-13
Update for Scala 2.13.5, 3.0.0, other dependencies

alexandru:


monix-execution/shared/src/main/scala/monix/execution/internal/exceptions.scala
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package monix.execution.internal
+
+private[monix] object exceptions {
+  /** 
+    * To use in `case _ =>` branches that are absurd, but needed
+    * due to the compiler complaining.
+    */  
+  def matchError[A](received: A): Nothing =
```
alexandru, line 25: This was needed because Scala 2.13.5 now complains on missing cases when using the partial-function syntax, when a `Function1` is requested as parameter. Pretty cool, not sure why it wasn't working before.

But we now have to throw MatchErrors manually in some cases.

Issue #1438, opened 2021-05-12
(Maybe) monix-execution should depend on org.slf4j.Logger


alexandru:
`org.slf4j` is a pervasive dependency in Scala's ecosystem.

We may need the ability to log warnings. Also, it would be nice if the Scheduler's `reportError` would do an error log via `org.slf4j.Logger`, because that's what we usually want.

What do you think?

alexandru:
Another possibility is to configure logging to be done via `java.util.logging` (JUL).

JUL is supported out of the box by [GraalVM's native image](https://www.graalvm.org/reference-manual/native-image/Logging/).

Slf4j / Logback users can then simply include this in their project if they want to use their logback.xml config:

```scala
"org.slf4j" % "jcl-over-slf4j" % "1.7.30"
```

PR #1444, opened 2021-05-14
Cats Effect 3 work stealing thread pool

vasilmkd:
Opening as a draft to get a feel for the CI. Will update the description in the future.

### DIFF
--- a/monix-execution/jvm/src/main/java/monix/execution/workstealing/LocalQueueConstants.java
+++ b/monix-execution/jvm/src/main/java/monix/execution/workstealing/LocalQueueConstants.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2014-2021 by The Monix Project Developers.
+ * See the project homepage at: https://monix.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * This code is a direct copy of the Cats Effect 3 source code, donated by the
+ * code authors and maintainers of the original code.
+ *
+ * The original Cats Effect 3 source code can be found at
+ * `https://github.com/typelevel/cats-effect/blob/27bb3fdffe86d299b8966eb352fa1b2fd8125395/core/jvm/src/main/java/cats/effect/unsafe/LocalQueueConstants.java`.
+ */
+
+package monix.execution.workstealing;
+
+/**
+ * Contains static constants used throughout the implementation of the
+ * [[monix.execution.workstealing.LocalQueue]].
+ */
+class LocalQueueConstants {
+
+  /**
+   * Fixed capacity of the [[monix.execution.workstealing.unsafe.LocalQueue]] implementation,
+   * empirically determined to provide a balance between memory footprint and
+   * enough headroom in the face of bursty workloads which spawn a lot of fibers
+   * in a short period of time.
+   * 
+   * Must be a power of 2.
+   */
+  static final int LocalQueueCapacity = 256;
+
+  /**
+   * Bitmask used for indexing into the circular buffer.
+   */
+  static final int LocalQueueCapacityMask = LocalQueueCapacity - 1;
+
+  /**
+   * Half of the local queue capacity.
+   */
+  static final int HalfLocalQueueCapacity = LocalQueueCapacity / 2;
+
+  /**
+   * Overflow fiber batch size.
+   */
+  static final int OverflowBatchSize = HalfLocalQueueCapacity + 1;
+
+  /**
+   * Bitmask used to extract the 16 least significant bits of a 32 bit integer
+   * value.
+   */
+  static final int UnsignedShortMask = (1 << 16) - 1;
+}
--- a/monix-execution/jvm/src/main/scala/monix/execution/schedulers/SchedulerCompanionImpl.scala
+++ b/monix-execution/jvm/src/main/scala/monix/execution/schedulers/SchedulerCompanionImpl.scala
@@ -22,6 +22,7 @@ import java.util.concurrent._
 import monix.execution.{Features, Scheduler, SchedulerCompanion, UncaughtExceptionReporter}
 // Prevents conflict with the deprecated symbol
 import monix.execution.{ExecutionModel => ExecModel}
+import monix.execution.workstealing.WorkStealingThreadPool
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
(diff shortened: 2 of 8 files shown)


monix-execution/jvm/src/main/scala/monix/execution/schedulers/SchedulerCompanionImpl.scala
```diff
@@ -473,7 +480,7 @@ private[execution] class SchedulerCompanionImpl extends SchedulerCompanion {
     implicit lazy val global: Scheduler =
       AsyncScheduler(
         DefaultScheduledExecutor,
-        ExecutionContext.Implicits.global,
+        workStealingThreadPool,
```
alexandru, line 483: Don't change the default for now, because `Scheduler.global` is meant for `Future` too, we need to think about the implications.

This is also why I opened #1438 — if we changed the default to a limited fixed thread-pool, maybe we should warn people that still execute `blocking` tasks on top.

monix-execution/jvm/src/main/java/monix/execution/workstealing/LocalQueueConstants.java
```diff
@@ -0,0 +1,64 @@
...
+ * `https://github.com/typelevel/cats-effect/blob/27bb3fdffe86d299b8966eb352fa1b2fd8125395/core/jvm/src/main/java/cats/effect/unsafe/LocalQueueConstants.java`.
+ */
+
+package monix.execution.workstealing;
+
+/**
+ * Contains static constants used throughout the implementation of the
+ * [[monix.execution.workstealing.LocalQueue]].
+ */
+class LocalQueueConstants {
```
alexandru, line 32: Out of curiosity, why is this a Java file? Just curious, the bulk of the implementation seems to be in Scala.

monix-execution/jvm/src/main/scala/monix/execution/workstealing/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,458 @@
...
+    }
+  }
+
+  /** Reschedules the given fiber directly on the local work stealing queue on the same thread,
+    * but with the possibility to skip notifying other fibers of a potential steal target, which
+    * reduces contention in workloads running on fewer worker threads. This method executes an
+    * unchecked cast to a `WorkerThread` and should only ever be called directly from a
+    * `WorkerThread`.
+    */
+  private[this] def rescheduleFiber(fiber: Runnable): Unit = {
```
alexandru, line 356: Why are these `private[this]`? Aren't `scheduleFiber` and `rescheduleFiber` supposed to be used?

Or maybe we'll expose them after modifying `Scheduler`.

alexandru (commented):
Awesome ❤️ 

On a first pass seems like a well-thought-out implementation, well documented, which I like.

monix-execution/jvm/src/main/scala/monix/execution/schedulers/SchedulerCompanionImpl.scala
vasilmkd: Oh, I only changed this to try running everything in the CI.

monix-execution/jvm/src/main/java/monix/execution/workstealing/LocalQueueConstants.java
vasilmkd: Optimization really. Scala `val`s are fields of object instances, and these are static constants. Loading static constants is faster than loading object fields. 😄

monix-execution/jvm/src/main/scala/monix/execution/workstealing/WorkStealingThreadPool.scala
vasilmkd: I really only did a single pass over the code,  this is far from ready. I guess it can be commented out for now. They definitely cannot be used at the moment. :/

monix-execution/jvm/src/main/scala/monix/execution/schedulers/SchedulerCompanionImpl.scala
vasilmkd: I think you need to explicitly allow this CI run since I'm a first time contributor to the project. New GitHub policy.

alexandru: Done — the tests are now running.

monix-execution/jvm/src/main/java/monix/execution/workstealing/LocalQueueConstants.java
alexandru: True. Although if you add them as `final val` in Scala, they'll be inlined AFAIK.

vasilmkd: I will definitely test that. Thanks.

monix-execution/jvm/src/main/scala/monix/execution/schedulers/SchedulerCompanionImpl.scala
alexandru: Do you have some tests for it BTW? Would be nice.

vasilmkd: I've been meaning to write jcstress tests for the queues. I'm not sure how to test the pool itself. I'm open to suggestions.

alexandru: IDK, some tests involving concurrency, happy path would be better than nothing. You might trigger the execution of some tasks via `sequence` and `parSequence` and see if the result is correct.

alexandru: Note that usage of `Future` in those tests is fine.

monix-execution/jvm/src/main/java/monix/execution/workstealing/LocalQueueConstants.java
vasilmkd: @alexandru 

Scala code:
```scala
object Main {
  final val n: Byte = 1
}
```

Compiles to the following:
```
public final class Main$ {
  public static final Main$ MODULE$;

  private static final byte n;

  public static {};
    Code:
       0: new           #2                  // class Main$
       3: dup
       4: invokespecial #14                 // Method "<init>":()V
       7: putstatic     #16                 // Field MODULE$:LMain$;
      10: iconst_1
      11: putstatic     #18                 // Field n:B
      14: return

  public final byte n();
    Code:
       0: getstatic     #18                 // Field n:B
       3: ireturn

  private Main$();
    Code:
       0: aload_0
       1: invokespecial #21                 // Method java/lang/Object."<init>":()V
       4: return
}
```

You're right, the field does indeed get inlined to a constant, and the virtual call `n()` can be inlined too. Let me know how you want to proceed.

alexandru: Well, if that satisfies your needs, I'd rather have the Scala file. We have some Java files too, I believe, for the atomic stuff, but this case seems simple enough.

vasilmkd:
I haven't forgotten about this. We're currently making some foundational changes to the CE3 thread pool and we're discussing splitting it out in a separate library.

alexandru:
@vasilmkd any word on that? 🙂

PR #1467, opened 2021-06-14
Add Scaladex badge to ReadMe to show Scala version support

rtyley:
This badge summarises which versions of Scala are supported by Monix (and what the latest Monix version is for each of those Scala versions):

[monix Scala version support](https://index.scala-lang.org/monix/monix/monix)

More details on the badge format: https://github.com/scalacenter/scaladex/pull/660


Avasil (approved):
That's pretty cool, thanks!

PR #1511, opened 2021-11-03
Adds Netflix to the list of monix adopters

jvican:
When I joined, the Notebooks team started to use Monix and all the backend
systems for our next-generation tooling uses Monix.


paualarco (approved):
👌🏻

PR #1512, opened 2021-11-09
Only include JCTools in executionShadedJCTools

gabro:
Currently JCTools accidentally brings in scala-collection-compat, this change makes sure that only JCTools is included and nothing else)

Here's the content of the currently published jar

```
~/Downloads/monix-internal-jctools_2.13-3.3.0
❯ tree -L 4
.
├── META-INF
│   └── MANIFEST.MF
├── monix
│   └── execution
│       └── internal
│           └── jctools
├── scala
│   ├── collection
│   │   └── compat
│   │       ├── immutable
│   │       ├── package$.class
│   │       └── package.class
│   └── util
│       └── control
│           └── compat
└── scala-collection-compat.properties
```

All the `scala/collection/compat` stuff shouldn't probably be there and it can cause conflicts when running assembly in an app which brings in a different version of `scala-collection-compat` (e.g. `scala-collection-compat.properties` will conflict).

This bug was originally reported on the Scala Discord (see https://discordapp.com/channels/632150470000902164/635669047588945930/907577696270295070)


gabro:
A similar, and maybe more general fix could be to "reset" the `libraryDependencies` to `Nil` in `assemblyShadeSettings` (although that's currently only used by the jctools module)

gabro:
Closing in favor of https://github.com/monix/monix/pull/1514

Issue #1519, opened 2021-12-10
Use macrotask executor for scala.js


Jasper-M:
See https://github.com/scala-js/scala-js-macrotask-executor

armanbilge:
This can be closed as of https://github.com/monix/monix/pull/1522.

PR #1522, opened 2021-12-15
Switch to `MacrotaskExecutor` on JS

armanbilge:
Closes https://github.com/monix/monix/issues/1519. Targeting series/3.x.

See https://github.com/scala-js/scala-js-macrotask-executor for more details.

Some notes:
1. This required bumping to Scala.js 1.7.1.
2. The monix-internal `StandardContext` this replaces actually did not suffer from the fairness problem that the default global `ExecutionContext` in Scala.js does. However, it seems like it would suffer from serious performance issues when used in browsers. Is anyone aware of this?


monix-execution/js/src/main/scala/monix/execution/schedulers/StandardContext.scala
```diff
@@ -1,50 +0,0 @@
...
-
-  override def reportFailure(cause: Throwable): Unit =
-    reporter.reportFailure(cause)
-
-  private[this] val executeRef: js.Dynamic = {
-    if (js.typeOf(js.Dynamic.global.setImmediate) == "function")
-      js.Dynamic.global.setImmediate
-    else
-      js.Dynamic.global.setTimeout
-  }
```
armanbilge, line 47: The `MacrotaskExecutor` is basically an advanced version of this conditional. First it looks for `setImmediate` (best performance). However, if that's not available, instead of giving up and falling back to `setTimeout` (very bad performance), `MacrotaskExecutor` attempts to implement a number of higher-performance polyfills instead.

armanbilge:
> 2\. The monix-internal `StandardContext` this replaces actually did not suffer from the fairness problem that the default global `ExecutionContext` in Scala.js does. However, it seems like it would suffer from serious performance issues when used in browsers. Is anyone aware of this?

Yes, according to https://github.com/scala-js/scala-js/issues/4129#issuecomment-900081602:

> Monix has been using `setImmediate` in its `Scheduler.global` whenever available, ever since I ported it to Scala.js. Nobody complained.

alexandru:
I was thinking of doing window.postMessage, not sure if feasible

https://github.com/monix/monix/issues/1556

alexandru:
I took the liberty of merging with the base branch, and doing some fixes in `build.sbt` and in that test.

Note that in Monix I'd prefer to drop 3rd party dependencies in the core modules, however seeing this is an "official" Scala.js project, I guess it's all good.

Thanks for the PR. Will merge as soon as the build passes.

armanbilge:
> Note that in Monix I'd prefer to drop 3rd party dependencies in the core modules, however seeing this is an "official" Scala.js project, I guess it's all good.

Yes, I read your blog post about the CE2/CE3 thing and 👍

The MacrotaskExecutor project should be very stable, ideally as stable as Scala.js itself. It's not part of Scala.js core largely due to their own policy/preference to avoid polyfills and non-standard APIs in core.

Issue #1529, opened 2022-01-12
operator request: groupByOrdered


borissmidt:
The operator ` def groupByOrdered[K](key: T => K): Observable[(K, Observable[T])] ` would be handy to check a stream until it changes. 
You could see it as a repeated takeWhile while equivalence of K is used to check if the condition changed.

The same behavior can be made with a scan, but i quite often ends up as complex code because you have to filter out the intermediate values and have to wrap everything in an option to know the end of the stream.

The groupByOrdered would also make it easier to reduce by key more efficiently if you know your data comes in a known order.
For example when you query a database.

```
def reduceByKeyOrdered[T, K](key : T => K)(op: (T, T)) ={
    o.groupByOrdered(key).flatMap { case (_, group) =>
      group.reduce(select)
    }
}
```
Of course a more generic implementation could accept `(previous: T, current: T) => Boolean` to know when to break the current group. 

I wouldn't mind to implement this operator if you think it is a good idea.

PR #1536, opened 2022-02-09
#1529 Add consecutiveTakeWhile and ConsecutiveGroupBy

borissmidt:
### consecutiveTakeWhile
Works like a takeWhile. Where you can compare two elements and take the observable as long as they are the same.
But starts a new group when the condition isn't met anymore.

### consecutiveGroupBy
Reuses the consecutiveTakeWhile to implement a groupBy that creates a new group when the key changes in the order of the observable.

Both are useful when you know your data comes in a certain order or you want to get statics depending on the order of the elements.

For example when i get data out of a database which is ordered by date the consecutiveGroupBy can be used to efficiently aggregate the database lines while responding downstream instead of waiting until the whole stream is consumed before answering:

```scala
case class Row(second: Int, value: Int)

//this will stream out of the database without caching the group results which might will create too many keys otherwise
query[Row].consecutiveGroupBy(_.second)
    .flatMap{case (second, rowsOfTheDay) => second -> rows.map(_.value).sum}

//this will group the elements as long as there value is the same
query[Row].consecutiveTakeWhile((prev, current) => prev.value == current.value).mapEval{ rows =>
  rows.firstL.flatMap{ firstRow => 
     rows.lastL
    }.map{ 
       case (firstRow, LastRow) => (  firstRow.second, LastRow.second) -> lastRow.value
    }
}
```

The same operation cannot be implemented with a scan or map Accumulate because then you have to buffer all elements of the group and drop all intermediate results. 

solves issue request: #1529


alexandru:
@borissmidt I'll have to digest this first, but at a first glance, it looks very appealing.

Sorry for the late response.

borissmidt:
No problem, i'm happy you are active again!

Issue #1363, opened 2021-01-21
Add `.memoizeFor` method to the Task and Observable API


jvican:
I think it would be useful to have a `(t: Task[T]).def memoizeFor(duration: FiniteDuration)` method that would memoize the value of a task only for a given period of time.
  
I've found myself in need of this method several times. A common scenario that requires these semantics is when you want to cache a value that you get from an external service but after a period of time you want to fetch it again. I feel this would be a powerful addition to the Task and Observable APIs. Any thoughts?

Avasil:
I would probably go for [ScalaCache](https://cb372.github.io/scalacache/docs/memoization.html) for these scenarios but I think we could add it for `Task` if it's not too hard to implement 👍 

What behavior do you have in mind for `Observable`? 
There's no `memoize` operator atm and it is not clear to me if it should be a variant of `ReplaySubject` that resets the cache each `duration`, or something entirely else

jvican:
Great, I'll take a look at implementing it. I haven't thought yet if/how implementing this in the Observable API makes sense.
  
It's interesting that you mention ScalaCache. I would use ScalaCache only for replacements of `ConcurrentHashMap`-based caching that needs to support rich eviction policies and/or synchronous waiting modes.

`memoizeFor` can be expressed with the ScalaCache API, but it really fells like a natural extension of `memoize` or some of the patterns that can be expressed with observables. For example, I've been mostly using `ReplaySubject` with multicast observables to implement `memoizeFor`. I think it's such a common idiom that it merits its inclusion in the Task API.

PR #1550, opened 2022-04-12
# add task cache

borissmidt:
#1363 add a cache method which works like memoize but only caches the value for a limited amount of time. 

I think it might be better to actually add it to the MemoizeTask and have the default memoize duration at Duration.inf but this is a quick POC to see how hard it was to implement and to get feedback on the feature.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -5,4 +5,4 @@
 -XX:+TieredCompilation
 -XX:-UseGCOverheadLimit
 -XX:+CMSClassUnloadingEnabled
-
+-XX:+IgnoreUnrecognizedVMOptions
--- a/monix-eval/shared/src/main/scala/monix/eval/Task.scala
+++ b/monix-eval/shared/src/main/scala/monix/eval/Task.scala
@@ -1176,29 +1176,39 @@ sealed abstract class Task[+A] extends Serializable with TaskDeprecated.BinCompa
   final def memoizeOnSuccess: Task[A] =
     TaskMemoize(this, cacheErrors = false, Duration.Inf)
 
-  /** cache the successful result of the source task for a given duration
-    * and reuses it on subsequent invocations of `runAsync`.
-    * Thrown exceptions are not cached.
+  /** Cache the result of the source task and reuses it on
+    * subsequent invocations of `runAsync` untill the duration has expired.
     *
-    * The resulting task will be idempotent, but only if the
-    * result is successful.
+    * The resulting task will be idempotent, meaning that
+    * evaluating the resulting task multiple times will have the
+    * same effect as evaluating it once.
     *
     * $memoizeCancel
     *
     * Example:
     * {{{
-    * ???
+    *   import scala.concurrent.CancellationException
+    *   import scala.concurrent.duration._
+    *
+    *   val source = Task(1).delayExecution(5.seconds)
+    *
+    *   // Option 1: trigger error on cancellation
+    *   val err = new CancellationException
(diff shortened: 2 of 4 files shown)


monix-eval/shared/src/main/scala/monix/eval/internal/TaskCache.scala
```diff
@@ -0,0 +1,191 @@
...
+            Task.unsafeStartTrampolined(self.thunk, ctx2, self.complete)
+          }
+
+        case ref: Promise[A] @unchecked =>
+          self.registerListener(ref, context, cb)
+
+        case (cacheTime: Long, ref: Try[A]) =>
+          // Race condition happened
+          // $COVERAGE-OFF$
+          if (sc.clockRealTime(MILLISECONDS) - cacheTime > duration.toMillis) {
```
alexandru, line 175: It's easier and possibly more efficient to store the time at which the value is supposed to expire, instead of the current timestamp. It saves you from doing this math operation on every read. Also, reuse `duration.toMillis` if possible.

alexandru (commented):
I would like to have this operator in Task's API.

There is concern about it being an "unsafe" operator, in the sense that it isn't referentially transparent. In Monix v4 I'd like for such operators to be marked by unsafe.

But for v3, its naming and signature is consistent with memoize 👍

Please ensure that it's properly tested. I'm assuming you copied the Memoize operator, but looking at it, I can't figure out whether it's correct or not — the downside of such low-level implementations.

Also, note that in v4 the implementation of Task will change drastically, so this code won't survive. But the API will, maybe with changes, this temporary caching possibly being a good idea 👍

monix-eval/shared/src/main/scala/monix/eval/internal/TaskCache.scala
borissmidt: Good idea, now i can recall this from my time with embedded systems. 
I think the added advantage of calculating the expiration time that it is immune to a counter that rolls over.

.jvmopts
```diff
@@ -5,4 +5,4 @@
 -XX:+TieredCompilation
 -XX:-UseGCOverheadLimit
 -XX:+CMSClassUnloadingEnabled
-
+-XX:+IgnoreUnrecognizedVMOptions
```
borissmidt, line 8: needed to run sbt with jdk17 (which boots up much faster)

borissmidt:
I found that i can debug the tests, but is there any way to run 1 specific test since this code is called by lots of tests. 
https://github.com/monix/minitest/issues/3
Maybe it would be more usefull to help working on version/4.0 instead of trying to get to make this low level api work. Since i haven't been able to fix it.

alexandru:
> I found that i can debug the tests, but is there any way to run 1 specific test since this code is called by lots of tests.
> https://github.com/monix/minitest/issues/3

Not sure what you mean.

You can run a specific test suite via sbt's `testOnly *Name`. A specific test in a test suite is harder to run, but you can just comment everything you don't need while debugging.

I personally don't run tests from inside the IDE. I understand that others do, which is why I'm contemplating switching to ScalaTest, since it has better IDE integration. The reason for why Monix doesn't use ScalaTest is because it's a big library that's hard to port to new platforms (back in the day, the issue was with Scala.js).

borissmidt:
I think [Munit](https://scalameta.org/munit/) would be very close to minitest and has IDE support.
But will try the `sbt testOnly *Name`

PR #1589, opened 2022-06-15
DynamicWorkerThreadFactory no longer manages max threads

### LINKED ISSUE #1588: Scheduler.forkJoin stops executing tasks after reaching its maximum thread count
Try running this code:

```scala
implicit val scheduler: Scheduler = Scheduler.forkJoin(4, 8)

// make the pool reach its maximum thread count
Task.parTraverseUnordered(0 until 1000) { _ =>
   Task(scala.concurrent.blocking(Thread.sleep(10)))
}.runSyncUnsafe()

println("slept")

// this task never executes
Task(println("foo")).executeAsync.runSyncUnsafe()

println("done")
```

After the pool reaches its maximum size caused by blocking operations, it does not execute any tasks **even despite the fact that blocking operations have already finished and threads are idle**.

The culprit seems to be `monix.execution.internal.forkJoin.DynamicWorkerThreadFactory` - the code above works fine if the Scheduler is defined like this:

```scala
implicit val scheduler: Scheduler = 
  Scheduler(new ForkJoinPool(4, ForkJoinPool.defaultForkJoinWorkerThreadFactory, null, true, 0, 8, 1, null, 60, TimeUnit.SECONDS))
```

Environment: Monix 3.4.1, Scala 2.13.8, JDK 11

ghik:
Fixes #1588

`DynamicWorkerThreadFactory` was previously managing `maxThreads` parameter of `Scheduler.forkJoin` manually, instead of passing it to the `ForkJoinPool` itself. Now this is changed so that only the pool itself is responsible for that.

I don't exactly understand why this fixes the problem or why it wasn't done like that from the start. Therefore I'm not sure if I've done the right thing since I'm not fully familiar with nitty-gritty details of `ForkJoinPool`.


halotukozak:
@alexandru this PR is still valid

Issue #1559, opened 2022-05-09
[4.x] Migrate project to usage of ScalaTest


alexandru:

Avasil:
👍 for using a different testing library and potentially retiring Minitest to decrease the project's scope

PR #1625, opened 2022-08-15
#1559 Topic/switch to munit test 2

borissmidt:
Hey alexandru,

I'm trying to move to the munit library and most of it seems to work kind of fine.

I'm already opening the PR to avoid someone else starting from 0 on this big change.


monix-eval/shared/src/test/scala/monix/eval/BaseTestSuite.scala
```diff
@@ -16,15 +16,21 @@
...
 import minitest.laws.Checkers
+import monix.execution.internal.Platform
 import monix.execution.schedulers.TestScheduler
 
-abstract class BaseTestSuite extends TestSuite[TestScheduler] with Checkers with ArbitraryInstances {
+abstract class BaseTestSuite extends FunSuite with Checkers with ArbitraryInstances {
+  override lazy val isCI = Platform.getEnv("CI").map(_.toLowerCase).contains("true")
 
+  def testAsync(name: String)(testBody: TestScheduler => Unit)(implicit loc: Location): Unit = {
+    super.test(name)(testBody(setup()))
```
alexandru, line 28: I don't see `tearDown` being called.

alexandru:
Looking good, but you need to solve those conflicts.

borissmidt:
I still need to fix a lot of compiler errors before this is ready for merging.

borissmidt:
I've been able to get the tests running, 

but the execution tests hang on the tests using the latches:
monix.execution.schedulers.ExecutorSchedulerSuite
```
  testAsync("reports errors on execute") { scheduler =>
    val latch = new CountDownLatch(1)
    self.synchronized {
      lastReportedFailure = null
      lastReportedFailureLatch = latch
    }

    try {
      val ex = DummyException("dummy")
      scheduler.execute(() => throw ex)

      assert(latch.await(15, TimeUnit.MINUTES), "lastReportedFailureLatch.await")
      self.synchronized(assertEquals(lastReportedFailure, ex))
    } finally {
      self.synchronized {
        lastReportedFailure = null
        lastReportedFailureLatch = null
      }
    }
  }
```

It seems that it is now possible to run tests from intellij which is nice, but sadly intellij doesn't understand the shaded library and complains on imports. Do you have any tricks to fix this?

alexandru:
On 23 Aug 2022, at 23:05, Boris Smidt wrote:
> but the execution tests hang on the tests using the latches:
> monix.execution.schedulers.ExecutorSchedulerSuite

In such cases it’s good to pay attention to the used thread-pool. What thread-pool is MUnit configured to use by default?

> It seems that it is now possible to run tests from intellij which is nice, but sadly intellij doesn't understand the shaded library and complains on imports. Do you have any tricks to fix this?

IntelliJ should be fine with the shaded libraries, I don’t remember issues, but Monix is the kind of project that is challenging for IntelliJ, so such issues happen — especially when code shared between targeted platforms is involved.

I’d try reimporting the project in IntelliJ. Sometimes it helps.

alexandru:
> For the rest everything seems to run fine and is green. I'm not sure how we will merge such a huge pr.

It's important that you merge with `series/4.x` and solve the conflicts, which are probably superficial — formatting related? I can also take a look afterwards at those flaky tests.

borissmidt:
I rebased on upstream/series4.x
It seems that the evalJS project fails on the linking stage. But i really have no experience in scalaJS to resolve these issues.

build.sbt
```diff
@@ -100,9 +100,10 @@ lazy val macrotaskExecutorLib =
 lazy val kindProjectorCompilerPlugin =
   "org.typelevel" % "kind-projector" % kindProjector_Version cross CrossVersion.full
 
-/** [[https://github.com/monix/minitest/]] */
-lazy val minitestLib =
-  Def.setting { "io.monix" %%% "minitest-laws" % minitest_Version }
+/** [[https://github.com/scalameta/munit]] */
+lazy val munitLib =
+  Def.setting { "org.scalameta" %%% "munit" % "1.0.0-M6" }
```
alexandru, line 105: Depending on `-M6` might be unwise, although I'm not familiar with MUnit's release cycle. Anything wrong with the latest stable version `0.7.29`?

Are there major API changes?

alexandru:
>  It seems that the evalJS project fails on the linking stage. But i really have no experience in scalaJS to resolve these issues.

I'll also take a look. Sounds like a dependency issue.

build.sbt
borissmidt: I'll swap it to the 0.7.29 version and see if it breaks the current code.

borissmidt:
I've downgraded munit and it seems to compile fine, so no big compatiblity issues are to be expected in the future when 1.0.0 is released.

borissmidt:
Closing the branch to remove loose ends in my open PR list.

PR #1786, opened 2023-10-16
Update scala-js and Scala 3

xuwei-k:


project/plugins.sbt
```diff
@@ -1,4 +1,4 @@
-addSbtPlugin("org.scala-js"              % "sbt-scalajs"             % "1.10.0")
+addSbtPlugin("org.scala-js"              % "sbt-scalajs"             % "1.14.0")
```
xuwei-k, line 1: old scala-js does not support Scala 3.3

monix-tail/shared/src/test/scala/monix/tail/IterantToReactivePublisherSuite.scala
```diff
@@ -227,9 +227,13 @@ object IterantToReactivePublisherSuite extends BaseTestSuite {
...
-    intercept[NullPointerException] {
-      Iterant[Task].of(1).toReactivePublisher.subscribe(null)
-      ()
+    if (Platform.isJVM) {
+      intercept[NullPointerException] {
+        Iterant[Task].of(1).toReactivePublisher.subscribe(null)
+        ()
+      }
+    } else {
+      // TODO
```
xuwei-k, line 236: does not work with latest scala-js 🤔 😢

Issue #1850, opened 2024-07-22
by-name parameter in `as` method


halotukozak:
`as` method should have by-name parameter, because `Task(...).map{ _ => ...}` cannot be always shorted to `Task(...).as{...}` due to the potential side effects

halotukozak:
PR: https://github.com/monix/monix/pull/1851

Jasper-M:
Whatever code you pass to `as` or `map` isn't supposed to have side effects though. The intended API in that case is `Task(...) >> Task(...)` (or `Task(...).flatMap(x => Task(...))` instead of `map`).

PR #1851, opened 2024-07-22
Use by-name parameter in `as` method

halotukozak:


halotukozak:
Closes: https://github.com/monix/monix/issues/1850

PR #2029, opened 2026-04-10
Synchronize ZipObservable onComplete implementations

halotukozak:
Original contribution from the AVSystem's fork made by @lpetka
https://github.com/AVSystem/monix/pull/1

> ZipObservable implementations (all flavors Zip{2,3,4,5,6}Observable) uses shared state in multithreaded environment. > ZipObservable implementations (all flavors `Zip{2,3,4,5,6}Observable`) uses shared state in multithreaded environment. There was a problem with underterministically non completing observables which was induced by the read of a non-volatile shared boolean without prior memory barrier.
> 
> I wasn't able to reproduce the problem on a current monix fork and the reason behind it is still unknown to me, although the major contributors might be:
> 
> * different java / scala version
> * custom test executor with sophisticated ticking implementation
> 
> The problem is reproducing kinda well for a current DHCP implementation both in testing suites and production code (the executor provided for the observable is irrelevant there).
> 
> Playground: https://gitlab.avsystem.com/csp/dslam/dhcp/-/commit/ca08ccb46078c6f8e8cde3f1261d51527e2f39d1

### DIFF
--- a/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Zip2Observable.scala
+++ b/monix-reactive/shared/src/main/scala/monix/reactive/internal/builders/Zip2Observable.scala
@@ -156,7 +156,7 @@ private[reactive] final class Zip2Observable[A1, A2, +R](obsA1: Observable[A1], 
       def onError(ex: Throwable): Unit =
         signalOnError(ex)
       def onComplete(): Unit =
-        signalOnComplete(hasElemA1)
+        lock.synchronized(signalOnComplete(hasElemA1))
     })
 
     composite += obsA2.unsafeSubscribeFn(new Subscriber[A2] {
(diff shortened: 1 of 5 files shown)


alexandru:
@halotukozak the Copilot review is right. The synchronization is correct, but it's slightly inefficient, because there's also a `lock.synchronized` in `signalOnComplete`. So the fix would be for `signalOnComplete` to no longer synchronize (with warning as comment) and thus rely on the added synchronization at the call-site.

Also, you're right, this is indeed a bug.

alexandru:
Can't modify this branch directly, I've started a new PR here: https://github.com/monix/monix/pull/2040

(I was under the impression that I can just modify PRs, still have issues with tech stuff :))

alexandru:
Change was easy enough that it was safe to just merge it. I hope you don't mind.

- https://github.com/monix/monix/pull/2040
- https://github.com/monix/monix/commit/936ac338e39e6eaccf2ff3734ac54abafd5a84d8

Will close this PR now, since it's fixed.

halotukozak:
> Change was easy enough that it was safe to just merge it. I hope you don't mind.
> 
> * [Fix zip observable to upstream (v2) #2040](https://github.com/monix/monix/pull/2040)
> * [936ac33](https://github.com/monix/monix/commit/936ac338e39e6eaccf2ff3734ac54abafd5a84d8)
> 
> Will close this PR now, since it's fixed.

Yes, thank you so much! I've only forwarded our PRs

PR #2030, opened 2026-04-10
Add logging of the mapping function in MapTaskObservable

halotukozak:
Original contribution from the AVSystem's fork made by @mikkolaj
https://github.com/AVSystem/monix/pull/2

> `MapTaskObservable` reports invalid state on some environments.
> 
> We weren't able to reproduce it synthetically, so in an attempt to gain better understanding of the error (and it's extent) I suggest logging the function used to process each element of the Observable when such failure occurs.


mikkolaj:
We stopped observing this behavior after introducing changes from #2031, so adding this logging might be unnecessary now.

